PR #16, opened 2023-01-17
refactor: Deduplicate optimized build step

KirilMihaylov:


.github/workflows/smart-contracts.yaml
```diff
@@ -94,24 +94,41 @@ jobs:
...
+          }
+
+          copy_deploy_script() {
+            sudo cp scripts/deploy-contracts-genesis.sh artifacts_${1}_net/
+          }
       - name: "Run workspace-optimizer docker image for test-net"
         shell: "sh"
-        run: "sudo docker run --rm -v \"$(pwd)\":/code -v \"$(pwd)/artifacts_test_net\":/code/artifacts --mount type=volume,source=\"$(basename \"$(pwd)\")_cache\",target=/code/target --mount type=volume,source=rustup_dir_cache,target=/usr/local/rustup --mount type=volume,source=cargo_dir_cache,target=/usr/local/cargo --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry cosmwasm/workspace-optimizer:0.12.11 --env ALT_NET_SYMBOLS=1"
-      - name: "Check for WASM binaries larger than allowed limit"
+        run: "build test 1"
```
gmanev7, line 116: Why do we pass them both here? Would suggest passing only "test/main" and have the script ingest the ALT_NET_SYMBOLS value

gmanev7:
good job! now the code is more expressive and less error-prone!

PR #19, opened 2023-01-25
[marketprice, oracle] PriceAlarms parameterization

go2bodhi:


gmanev7 (commented):
@go2bodhi please review and cleanup variants of marketprice::alarms::AlarmError that are stale

contracts/oracle/src/contract/alarms.rs
```diff
@@ -1,62 +1,216 @@
...
+            AddAlarms {
+                storage,
+                addr,
+                above,
+                price_alarms: Self::PRICE_ALARMS,
+            },
+        )
+    }
+
+    fn schedule_alarm(
```
gmanev7, line 89: please order item per their visibility, public first

contracts/oracle/src/contract/alarms.rs
```diff
@@ -1,62 +1,216 @@
...
+
+            fn exec<C>(self, price: Price<C, BaseC>) -> Result<Self::Output, Self::Error>
+            where
+                C: Currency,
+            {
+                let initial = *self.next_id;
+
+                self.price_alarms
+                    .alarms(self.storage, price)
+                    .take(self.max_count as usize)
```
gmanev7, line 139: please avoid using `as` since it may mask overflow issues, try_into()? would be a better alternative. I realize overflow may never happen since `usize` at wasm32 should be u32 ... but this builds bad code practices.

contracts/oracle/src/contract/alarms.rs
```diff
@@ -1,62 +1,216 @@
...
+                    .try_for_each(|addr_result| {
+                        addr_result.map(|addr| {
+                            MarketAlarms::schedule_alarm(self.batch, addr, self.next_id)
+                        })?
+                    })?;
+                let processed = *self.next_id - initial;
```
gmanev7, line 145: this will overflow once getting next id wraps

gmanev7: would suggest counting the elements instead

contracts/oracle/src/contract/alarms.rs
```diff
@@ -1,62 +1,216 @@
...
+            }
+        }
+
+        let mut next_id = Self::MSG_ID.may_load(storage)?.unwrap_or_default();
+        let mut processed = 0u32;
+
+        prices
+            .iter()
+            .try_for_each(|price| -> Result<(), ContractError> {
+                processed += with_quote::execute::<_, _, _, BaseC>(
```
gmanev7, line 156: To simplify collecting items to Batch and capping their number would suggest `NotifyCmd` return only an iterator (not sure how that would fit `WithQuote`). Then use Iterator::flatten and Iterator::take.

A not-so-good alternative would be for `NotifyCmd` to return Batch building it with Iterator::fold. Then use Batch::merge to combine into a final Batch. Here you may want to use Iterator::reduce instead of Iterator::for_each. The last piece of the puzzle is missing - how to get the length of a Batch. I suggest adding a method.

contracts/oracle/src/contract/alarms.rs
```diff
@@ -1,62 +1,385 @@
...
+                .next()
+                .transpose()?
+                .is_some(),
+        })
+    }
+
+    fn alarms_iter<'a, BaseC>(
+        storage: &'a dyn Storage,
+        prices: &'a [SpotPrice],
+    ) -> impl Iterator<Item = Result<Addr, StdError>> + 'a
```
gmanev7, line 134: StdError -> ContractError?

gmanev7: more specifically, the PriceAlarms::alarms should return items that are Result with AlarmError ... to make the marketprice API consistent and to have it use its own error type instead of the one of the internal implementation

gmanev7:
LGTM feel free to merge once address the last suggestion

PR #23, opened 2023-01-26
feat: Extend `tree` library to support for getting a subtree, iterator over children and conversion back to human-readable form

KirilMihaylov:


packages/tree/src/node.rs
```diff
@@ -46,13 +51,59 @@ impl<'r, T> NodeRef<'r, T> {
...
+
+pub struct ParentsIter<'r, T> {
+    node: NodeRef<'r, T>,
+}
+
+impl<'r, T> Iterator for ParentsIter<'r, T> {
+    type Item = NodeRef<'r, T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let parent = self.node.parent()?;
```
gmanev7, line 109: this is unstable yet https://doc.rust-lang.org/src/core/option.rs.html#2285

gmanev7: how does it compile?

packages/tree/src/node.rs
```diff
@@ -46,13 +51,59 @@ impl<'r, T> NodeRef<'r, T> {
...
+        Subtree::from_node(*self)
+    }
+
+    #[inline]
+    pub(crate) fn tree(&self) -> &'r Tree<T> {
+        self.tree
+    }
+
+    #[inline]
+    pub(crate) fn this_index(&self) -> u16 {
```
gmanev7, line 83: u16 !!!

packages/tree/src/node.rs
KirilMihaylov: As discussed in DMs, this is just a sugar. It's usable with both Result and Option types.

PR #27, opened 2023-02-02
refactor: Changed versioning mechanism to have separate storage & software version

KirilMihaylov:


packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,88 @@
...
-    schemars::{self, JsonSchema},
 };
 
 pub type Version = u16;
 
-pub const VERSION_ITEM: Item<'static, Version> = Item::new("contract_version");
+pub type SemVer = (Version, Version, Version);
+
+#[macro_export]
+macro_rules! package_version {
```
gmanev7, line 11: why do we define it as a macro instead of as a regular function?

gmanev7: wny not just passing core::env!("CARGO_PKG_VERSION")?

contracts/dispatcher/src/contract.rs
```diff
@@ -23,7 +23,8 @@ use crate::{
 };
 
 // version info for migration info
-const CONTRACT_VERSION: Version = 0;
+// const EXPECTED_MIGRATION_STORAGE_VERSION: Version = 0;
```
gmanev7, line 26: [everywhere] please rename to `CONTRACT_STORAGE_VERSION_FROM`

contracts/lease/src/contract/mod.rs
```diff
@@ -26,7 +26,8 @@ mod cmd;
 pub mod msg;
 mod state;
 
-const CONTRACT_VERSION: Version = 0;
+const EXPECTED_MIGRATION_STORAGE_VERSION: Version = 0;
```
gmanev7, line 29: that should be an edge case because there is no version in the storage at all. The current implementation of `versioning::upgrade_contract` does not handle it and would fail.

gmanev7: so if `CONTRACT_STORAGE_VERSION == ZERO` then the function should only store it

packages/versioning/src/lib.rs
KirilMihaylov: My point was to not have any constants and stuff lying around and at the same time making it more ergonomic to use.

contracts/dispatcher/src/contract.rs
KirilMihaylov: Addressed.

contracts/lease/src/contract/mod.rs
KirilMihaylov: Added a compatibility method to handle those cases. As marked with TODO, should be removed after all contracts have been migrated.

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,88 @@
...
-    cosmwasm_std::{to_binary, Binary, StdError, StdResult, Storage},
+    cosmwasm_std::{StdError, StdResult, Storage},
     cw_storage_plus::Item,
-    schemars::{self, JsonSchema},
 };
 
 pub type Version = u16;
 
-pub const VERSION_ITEM: Item<'static, Version> = Item::new("contract_version");
+pub type SemVer = (Version, Version, Version);
```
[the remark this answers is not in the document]
KirilMihaylov: I agree about the structure that it would be a bit lighter when both are changed, I am not sure whether that would be the case just for a software upgrade.

I strongly disagree with this approach because that should already be a function of the `migrate` function, not the version checking one. This leads me to the thought of renaming it to say something like `update_version_with_check` instead of `upgrade_contract` but again, I'd like to emphasize on the fact that putting a function call will not only break the single responsibility principle but also introduce a separate function call and call frame, when we already have a context, the `migrate` method, which is supposed to do that.

KirilMihaylov:
> once you define a single type representing the contract version then each contract would have only one constant of that type ... and there would be a macro just for determining the storage component

I feel like there is a misunderstanding here. I specifically added this macro, not just for fun, but because otherwise the `env!("CARGO_PGK_VERSION")` will be resolved within the **`versioning`** crate. I can of course always just move out the `env!` macro call into each crate which then would imply we have the same constant which, mind you, can't substitute because the macro expects a literal. Doing that is way more error-prone than a 1:1 macro that has close to little impact on our code.

I can always extract the function from it, if that's the problem, but the macro, in my opinion, has to stay.

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,183 @@
...
-pub enum WithVersion<Q> {
-    Version {
-        version: (),
-    },
-    Query {
-        #[serde(flatten)]
-        query: Q,
-    },
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct VersionPair {
```
gmanev7, line 20: rename to Version pls

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,183 @@
...
+    () => {{
+        $crate::parse_semver(::core::env!(
+            "CARGO_PKG_VERSION",
+            "Cargo package version is not set as an environment variable!",
+        ))
+    }};
+}
 
-    use super::WithVersion;
+const VERSION_PAIR_ITEM: Item<'static, VersionPair> = Item::new("contract_version_pair");
```
gmanev7, line 69: contract_version_pair -> contract_version

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,183 @@
...
+    component_version: SemVer,
+    migrate_storage_functor: Option<MigrateStorageFunctor>,
+) -> Result<(), MigrateStorageError>
+where
+    MigrateStorageFunctor: FnOnce(&'r mut dyn Storage) -> Result<(), MigrateStorageError>,
+    MigrateStorageError: From<StdError> + Error,
+{
+    pub const CW_VERSION_ITEM: Item<'static, u16> = Item::new("contract_info");
+
+    pub const OLD_VERSION_ITEM: Item<'static, u16> = Item::new("contract_version");
```
gmanev7, line 101: do we need them `public`?

KirilMihaylov: No, that's a, while not an issue, a mistake.

packages/versioning/src/lib.rs
gmanev7, line 69: -> VERSION_STORAGE_KEY

gmanev7: or just `VERSION`

KirilMihaylov: Addressed.

contracts/leaser/src/contract.rs
```diff
@@ -17,7 +17,9 @@ use crate::{
     state::{config::Config, leases::Leases},
 };
 
-const CONTRACT_VERSION: Version = 1;
+// version info for migration info
+// const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 0;
+const CONTRACT_STORAGE_VERSION: VersionSegment = 0;
```
gmanev7, line 22: We agreed on keeping the `CONTRACT_VERSION` not only its component

KirilMihaylov: I now understand what you mean and I have tried doing that in the beginning. I wasn't able to get it to compile within a constant context. It thrown an error that the computations exceeded const-eval limits.

KirilMihaylov: The only reliable way I can think of doing that is to instead write a procedural macro, which I am not confident in being able to do.

gmanev7: no, let's not complicate going with that new macro .... then we can convert the existing macro `parse_.....` into `version!` taking the storage constant as a parameter

The end design goal is to have the `version` lib API deal with the `Version` entity, not with its two components separately, as that is now.

KirilMihaylov: Addressed.

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,187 @@
...
-        assert_eq!(
-            serde_json::from_str::<WithVersion<VariantsQuery>>(r#"{"version":{}}"#).unwrap(),
-            QUERY_VERSION
-        );
+    if version.storage != FROM_STORAGE_VERSION.wrapping_add(1) {
+        return Err(StdError::generic_err("Expected and new storage versions are not directly adjacent! This could indicate an error!").into());
     }
+
+    VERSION_STORAGE_KEY.update(storage, |saved_version| {
+        if saved_version.storage != version.storage {
```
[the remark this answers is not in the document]
KirilMihaylov: Good catch! Thank you!

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,187 @@
...
+            ));
+        }
+
+        if saved_version.software < version.software {
+            Ok(version)
+        } else {
+            Err(StdError::generic_err(
+                "Couldn't upgrade contract because software version isn't monotonically increasing!",
+            ))
+        }
```
gmanev7, line 183: I believe we may re-use this code fragment with  `update_software` .... and limit the bugs :-)

KirilMihaylov: Addressed.

packages/versioning/src/lib.rs
```diff
@@ -1,115 +1,180 @@
...
-
-        assert_eq!(
-            serde_json::from_str::<WithVersion<VariantsQuery>>(r#"{"version":{}}"#).unwrap(),
-            QUERY_VERSION
-        );
+    if version.storage != FROM_STORAGE_VERSION.wrapping_add(1) {
+        return Err(StdError::generic_err("Expected and new storage versions are not directly adjacent! This could indicate an error!").into());
     }
+
+    update_version(storage, version.storage, version)?;
```
gmanev7, line 157: should not pass `FROM_STORAGE_VERSION` instead?

KirilMihaylov: Yes, it should be.

Issue #22, opened 2023-01-25
[oracle] MarketAlarms singleton


go2bodhi:
We need to get rid of singleton approach for the `oracle::contract::alarms::MarketAlarms` in the oracle.

We have to use a new instance of `MarketAlarms` that is tailored to the specific needs of the oracle instance, for example, the oracle base currency.

gmanev7:
done

PR #30, opened 2023-02-08
[oracle] prices calculation optimization

go2bodhi:


contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +193,148 @@ pub fn try_query_alarms<OracleBase>(
...
+                .is(Coin::<TheCurrency>::new(1))
+                .into(),
+            price::total_of(Coin::<Juno>::new(2))
+                .is(Coin::<TheCurrency>::new(1))
+                .into(),
+            price::total_of(Coin::<Cro>::new(6))
+                .is(Coin::<TheCurrency>::new(1))
+                .into(),
+        ];
+
```
gmanev7, line 337: pls remove the code duplications and make it more readable by defining a local fn

contracts/oracle/src/contract/alarms.rs
```diff
@@ -89,33 +99,31 @@ impl MarketAlarms {
...
-        let mut next_id = Self::MSG_ID.may_load(storage)?.unwrap_or_default();
-        let initial_id = next_id;
+        let initial_id = self.0;
 
         Self::alarms_iter::<BaseC>(storage, prices)
             .take(max_count.try_into()?)
-            .try_for_each(|addr| Self::schedule_alarm(&mut batch, addr?, &mut next_id))?;
-
-        Self::MSG_ID.save(storage, &next_id)?;
+            .try_for_each(|addr| Self::schedule_alarm(&mut batch, addr?, &mut self.0))?;
```
gmanev7, line 116: [everywhere] to improve readability pls do not use so generic argument names like `addr`. `Addr` is a type. In this particular case, we may name the reference `receiver` or `listener`.

contracts/oracle/src/contract/alarms.rs
```diff
@@ -25,14 +26,23 @@ use crate::{
 
 pub type AlarmReplyId = u64;
 
-pub struct MarketAlarms;
+pub struct MarketAlarms(AlarmReplyId);
 
 impl MarketAlarms {
     const PRICE_ALARMS: PriceAlarms<'static> =
         PriceAlarms::new("alarms_below", "index_below", "alarms_above", "index_above");
```
gmanev7, line 33: Create a new instance always and store it in MarketAlarms. Pass a storage instance. Then a lot of code will become obsolete and be removed, e.g. passing a storage down the stack

contracts/oracle/src/contract/alarms.rs
```diff
@@ -25,14 +26,23 @@ use crate::{
...
 
+    pub fn load(storage: &dyn Storage) -> Result<Self, ContractError> {
+        let next_id = Self::MSG_ID.may_load(storage)?.unwrap_or_default();
+        Ok(MarketAlarms(next_id))
+    }
+
+    pub fn save(&self, storage: &mut dyn Storage) -> Result<(), ContractError> {
+        Ok(Self::MSG_ID.save(storage, &self.0)?)
+    }
+
```
gmanev7, line 45: why do we store the reply id? After all, it should be unique in the scope of a single transaction.

gmanev7: after consulting [this](https://docs.cosmwasm.com/docs/smart-contracts/message/submessage) and [this](https://github.com/CosmWasm/cosmwasm/blob/32f308a1a56ae5b8278947891306f7a374c3df94/packages/std/src/results/submessages.rs#L32) it seems that they even not need to be unique at all. 

So please remove the code that stores it. I believe a single constant would suffice.

contracts/oracle/src/contract/alarms.rs
```diff
@@ -25,14 +26,23 @@ use crate::{
 
 pub type AlarmReplyId = u64;
 
-pub struct MarketAlarms;
+pub struct MarketAlarms(AlarmReplyId);
```
gmanev7, line 29: [probably out of the scope of this PR] parameterizing by BaseC?

contracts/oracle/src/contract/alarms.rs
gmanev7, line 116: please refrain from using output function arguments

gmanev7: what about using `Iterator::try_fold` with Init=Batch and R=Batch, and adding a fn `Batch::len(&self) -> u32` for calculating `processed`

contracts/oracle/src/contract/feed.rs
```diff
@@ -77,20 +83,88 @@ where
         Ok(prices)
     }
 
-    fn calc_all_prices(
-        &self,
-        storage: &dyn Storage,
+    pub fn all_prices_iter<'a>(
```
gmanev7, line 86: this function is more than 80 lines long, contains many sub-scopes and the logic inside is not tested enough

please simplify and *unit* test the code that goes out of the refactoring

gmanev7, line 86: think about unifying the implementation of this use-case and the previous one for specific currencies done with `fn calc_prices()`

packages/finance/src/price/dto/mod.rs
```diff
@@ -194,6 +194,53 @@ where
         Base: Currency;
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct BasePrice<QuoteC, BaseG>
```
gmanev7, line 198: please:
- extract it into its own module `finance::price::base`. Rationale to be outside of `dto` is that it is not used for data transfer, no ser-de implemented, it is just another form of `finance::price::Price`
- change the places of the two type parameters to reflect the order in `finance::price::Price` and of their usage in the data members

contracts/oracle/src/contract/feed.rs
```diff
@@ -77,20 +83,88 @@ where
...
+    ) -> Result<impl Iterator<Item = BasePrice<OracleBase, SwapGroup>> + 'a, ContractError> {
+        struct LegCmd<'a, 'b, OracleBase>
+        where
+            OracleBase: Currency,
+        {
+            feeds: &'a PriceFeeds<'static>,
+            storage: &'a dyn Storage,
+            at: Timestamp,
+            total_feeders: usize,
+            stack: &'b mut Vec<BasePrice<OracleBase, SwapGroup>>,
```
gmanev7, line 101: Not sure I understand the need of storing them all. Besides the extra memory that requires probably a not-so-cheap conversion back into Price<C1, C2>.

contracts/oracle/src/contract/alarms.rs
gmanev7: fixed with #22

contracts/oracle/src/contract/alarms.rs
gmanev7: Discussed offline that the untyped `::remove` on alarms reply is an obstacle. @go2bodhi please file an issue and link it here.

contracts/oracle/src/contract/feed.rs
gmanev7: @go2bodhi please add a ticket for that and link it here

contracts/oracle/src/contract/feed.rs
gmanev7: clarified offline, we do need it. Just agreed to hide it as part of the state.

packages/platform/src/batch.rs
```diff
@@ -142,6 +142,14 @@ impl Batch {
...
     }
 
+    pub fn len(&self) -> usize {
+        self.msgs.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.msgs.is_empty()
+    }
+
```
gmanev7, line 152: please add unit tests to the `test` rust sub-module or at least add asserts to the existing unit tests

contracts/oracle/src/contract/alarms.rs
```diff
@@ -82,40 +76,44 @@ impl MarketAlarms {
...
-        storage: &mut dyn Storage,
-        mut batch: Batch,
-        prices: &[SpotPrice],
+    pub fn try_notify_alarms<'a, BaseC>(
+        &mut self,
+        storage: &dyn Storage,
+        prices: impl Iterator<Item = BasePrice<SwapGroup, BaseC>> + 'a,
         max_count: u32, // TODO: type alias
-    ) -> Result<Response, ContractError>
+    ) -> Result<Batch, ContractError>
```
gmanev7, line 91: From a design perspective and for better testability we shall return here `impl Iterator<Item = Addr>`.

contracts/oracle/src/contract/alarms.rs
```diff
@@ -82,40 +76,44 @@ impl MarketAlarms {
...
+                |mut batch, receiver| -> Result<Batch, ContractError> {
+                    // TODO: get rid of the Nls dummy type argument
+                    batch.schedule_execute_wasm_reply_always::<_, Nls>(
+                        &receiver?,
+                        ExecuteAlarmMsg::PriceAlarm(),
+                        None,
+                        batch.len().try_into()?,
+                    )?;
+                    Ok(batch)
+                },
```
gmanev7, line 108: collecting the alarms' receivers into a Batch should be taken up by the caller

contracts/oracle/src/contract/alarms.rs
```diff
@@ -130,68 +128,56 @@ impl MarketAlarms {
...
+            OracleBase: Currency,
         {
             type Error = ContractError;
             type Output = AlarmsIterator<'a>;
 
-            fn exec<C>(self, price: Price<C, BaseC>) -> Result<Self::Output, Self::Error>
+            fn on<C>(self) -> finance::currency::AnyVisitorResult<Self>
             where
-                C: Currency,
+                C: Currency + serde::Serialize + serde::de::DeserializeOwned,
```
gmanev7, line 154: [nit] add `use` statements for the serde types?

contracts/oracle/src/contract/alarms.rs
```diff
@@ -311,23 +280,23 @@ mod test {
...
 
             MarketAlarms::try_add_price_alarm::<Base>(
                 &mut storage,
-                addr,
+                receiver,
                 AlarmDTO::new(
                     price::total_of(Coin::<Atom>::new(1)).is(Coin::<Base>::new(10 + delta)),
                     Some(price::total_of(Coin::<Atom>::new(1)).is(Coin::<Base>::new(30 + delta))),
                 ),
             )
```
gmanev7, line 292: with the changed result type, please craft more tests and tailor them to cover more specific cases, e.g. various combinations of above/below alarms and receivers. The asserts should check the addresses, not just their length.

gmanev7: that is **very important to be tested thoroughly** because a bug here would have huge implication to the system as a whole

gmanev7: it's much better to have more shorter tests than a few longer

contracts/oracle/src/contract/alarms.rs
```diff
@@ -82,40 +76,44 @@ impl MarketAlarms {
...
-            .set_data(to_binary(&DispatchAlarmsResponse(processed.try_into()?))?))
+        Ok(batch)
     }
 
-    pub fn try_query_alarms<BaseC>(
+    pub fn try_query_alarms<'a, BaseC>(
         storage: &dyn Storage,
-        prices: &[SpotPrice],
+        prices: impl Iterator<Item = BasePrice<SwapGroup, BaseC>> + 'a,
     ) -> Result<AlarmsStatusResponse, ContractError>
```
gmanev7, line 117: from a design perspective, similar to the above issue, we shall return only `bool`

contracts/oracle/src/contract/feed.rs
gmanev7: #38

contracts/oracle/src/contract/feed.rs
```diff
@@ -19,7 +21,10 @@ use crate::{
 
 use super::{alarms::MarketAlarms, feeder::Feeders};
 
-pub struct Feeds<OracleBase> {
+pub struct Feeds<OracleBase>
+where
+    OracleBase: Currency + DeserializeOwned,
+{
```
gmanev7, line 27: unnecessary

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
+
+        if price.is_none() {
+            self.err = true;
+        }
+
+        Ok(price)
+    }
+}
+
+// TODO: rename to something meaningfull
```
gmanev7, line 267: suggest `FedPrices`

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
+            .feeds
+            .price_of_feed(self.storage, self.at, self.total_feeders)
+            .map(Some)
+            .or_else(|err| match err {
+                PriceFeedsError::NoPrice() => Ok(None),
+                _ => Err(err),
+            })?;
+        Ok(price)
+    }
+}
```
gmanev7, line 298: No tests for this code. Suggest extracting the logic that converts `Result<Price<>, E> -> Result<Option<Price<>>, E>` into a function to allow testing it.

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
 {
     let total_registered = Feeders::total_registered(storage)?;
     use crate::state::config::Config as OracleConfig;
     let config = OracleConfig::load(storage)?;
     let oracle = Feeds::<OracleBase>::with(config.price_config);
-    oracle.calc_all_prices(storage, block_time, total_registered)
+    oracle.all_prices_iter(storage, tree.swap_pairs_df(), block_time, total_registered)
+}
+
+struct LegCmd<OracleBase, Querier>
```
gmanev7, line 172: too much code for a module. Please extract this and the next entities in sub-modules.

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
+    stack: Vec<BasePrice<SwapGroup, OracleBase>>,
+    err: bool,
+}
+
+impl<Querier, OracleBase> LegCmd<OracleBase, Querier>
+where
+    OracleBase: Currency,
+    Querier: PriceQuerier,
+{
+    // TODO: improve implementation
```
gmanev7, line 187: let's not forget to address it before merging

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
+}
+
+struct LegCmd<OracleBase, Querier>
+where
+    OracleBase: Currency,
+    Querier: PriceQuerier,
+{
+    price_querier: Querier,
+    stack: Vec<BasePrice<SwapGroup, OracleBase>>,
+    err: bool,
```
gmanev7, line 179: please find more meaningful name

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
+            let parent_idx = self
+                .stack
+                .iter()
+                .position(|parent| parent.base_ticker() == Q::TICKER);
+
+            if let Some(n) = parent_idx {
+                self.stack.truncate(n + 1);
+                self.err = false;
+            } else {
+                return false;
```
gmanev7, line 206: try to refactor this function to avoid this mid-function return. In general, it's a source of bugs and high-cost maintenance

contracts/oracle/src/contract/feed.rs
```diff
@@ -127,20 +149,442 @@ pub fn try_query_alarms<OracleBase>(
...
+                _ => {
+                    self.stack.truncate(self.stack.len() - 1);
+                }
+            }
+        }
+        .map(|price| {
+            let bprice: BasePrice<SwapGroup, OracleBase> = price.into();
+            self.stack.push(bprice.clone());
+            bprice
+        });
```
gmanev7, line 257: Too cryptic logic and hard to maintain further, therefore:
- please replace `loop-break` with `Iterator::find_map` that returns Option<(position, price)>
- None means no price
- Some means the next price is that and it is an ancestor of the stack element at that position
- eliminate `LegCmd::err`
- eliminate `LegCmd::recover`
- remove any logic related to setting/reading `err`

contracts/oracle/src/contract/feed.rs
gmanev7: that logic becomes redundant when address the issue below

packages/platform/src/batch.rs
```diff
@@ -142,6 +142,14 @@ impl Batch {
         res
     }
 
+    pub fn len(&self) -> usize {
+        self.msgs.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.msgs.is_empty()
+    }
```
gmanev7, line 151: it seems this is unused, please remove

contracts/oracle/src/contract/alarms.rs
go2bodhi: New related issues were added: #47 , #48

contracts/oracle/src/contract/feed/price_querier.rs
```diff
@@ -0,0 +1,68 @@
...
+        Q: Currency + DeserializeOwned,
+    {
+        let price = self
+            .feeds
+            .price_of_feed(self.storage, self.at, self.total_feeders);
+        maybe_price(price)
+    }
+}
+
+pub fn maybe_price<B, Q>(
```
gmanev7, line 34: not necessary to be `pub`

contracts/oracle/src/contract/feed/leg_cmd.rs
```diff
@@ -0,0 +1,174 @@
...
+use serde::de::DeserializeOwned;
+use swap::SwapGroup;
+
+pub struct LegCmd<OracleBase, Querier>
+where
+    OracleBase: Currency,
+    Querier: PriceQuerier,
+{
+    pub price_querier: Querier,
+    pub stack: Vec<BasePrice<SwapGroup, OracleBase>>,
```
gmanev7, line 16: please fix this encapsulation issue, the members are `pub` in a non-DTO struct

contracts/oracle/src/contract/feed/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+        price::{dto::PriceDTO, Price},
+    };
+    use price_querier::PriceQuerier;
+    use sdk::cosmwasm_std::testing::{self, MockStorage};
+    use tree::HumanReadableTree;
+
+    #[derive(Clone)]
+    pub struct TestFeeds(pub HashMap<(SymbolStatic, SymbolStatic), PriceDTO<SwapGroup, SwapGroup>>);
+    impl TestFeeds {
+        pub fn add<B, Q>(&mut self, total_of: u128, is: u128)
```
gmanev7, line 208: please use the right type instead of `u128`

contracts/oracle/src/contract/feed/leg_cmd.rs
```diff
@@ -0,0 +1,174 @@
...
+            .map(Option::flatten)?;
+
+        let idx_price = branch_price.map_or_else(
+            || {
+                self.price_querier
+                    .price::<B, OracleBase>()
+                    .map(|res_price| res_price.map(|price| (0, price)))
+            },
+            |idx_price| Ok(Some(idx_price)),
+        )?;
```
gmanev7, line 56: I am afraid here the logic is a source of bugs. There are three cases for `branch_price` to be None:
- Q is found in the stack but there is no B-to-Q price at the moment
- Q is not found because Q is a descendant to the top of the stack T, for example, on the previous round price(Q, T) was None and Q was not pushed to the stack
- Q is not found because Q == OracleBase and this is the first time it is invoked
The implementation here assumes it is the last option.

gmanev7: IMHO, one way to fix it is to consider two main cases - stack empty or not. If empty, add B-to-Q price only if Q == OracleBase, otherwise skip. If not empty, search for Q. If found, calculate the price, and add it if Some. In both other cases, Q is not found or no price, None.

contracts/oracle/src/contract/feed/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+            .scan(cmd, |cmd, leg| {
+                let res = currency::visit_any_on_tickers::<SwapGroup, SwapGroup, _>(
+                    &leg.from,
+                    &leg.to.target,
+                    cmd,
+                )
+                .expect("price calculation error");
+                Some(res)
+            })
+            // TODO: process errors
```
gmanev7, line 112: This is part of the alarms delivery call path which is a subject of this PR. Any idea if/how this should be processed?

packages/finance/src/price/base.rs
```diff
@@ -0,0 +1,54 @@
...
+    error::Error,
+};
+
+use super::Price;
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct BasePrice<BaseG, QuoteC>
+where
+    QuoteC: Currency,
+    BaseG: Group,
```
gmanev7, line 13: [everywhere] please align the order with their appearance in the type list

contracts/oracle/src/contract/alarms/iter.rs
```diff
@@ -0,0 +1,133 @@
...
+    I: Iterator<Item = PriceResult<BaseC>> + 'a,
+    BaseC: Currency,
+{
+    type Item = Result<Addr, ContractError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.alarms
+            .as_mut()
+            .and_then(|alarms| alarms.next())
+            .map(|res| res.map_err(ContractError::from))
```
gmanev7, line 92: Define a function `&mut AlarmsIterator -> Option<<Self as Iterator>::Item>` to reuse obtaining the next element and mapping the error. Then re-use that function in `fn next_price` lines 71-73.

contracts/oracle/src/contract/alarms/iter.rs
```diff
@@ -0,0 +1,133 @@
...
+        let res = alarms_iter.next();
+        assert_eq!(Some(Ok(Addr::unchecked("recv2"))), res);
+
+        // passing empty Cro iterator
+        let res = alarms_iter.next();
+        assert_eq!(Some(Ok(Addr::unchecked("recv4"))), res);
+
+        // price error propagation
+        let res = alarms_iter.next();
+        assert_eq!(Some(Err(StdError::generic_err("error").into())), res);
```
gmanev7, line 131: could you extend the test, or add another, to check if the alarms iteration continues with the next prices

contracts/oracle/src/contract/feed/price_querier.rs
```diff
@@ -0,0 +1,68 @@
...
+use finance::{currency::Currency, price::Price};
+use marketprice::{error::PriceFeedsError, market_price::PriceFeeds};
+use sdk::cosmwasm_std::{Storage, Timestamp};
+use serde::de::DeserializeOwned;
+
+pub struct FedPrices<'a> {
+    pub feeds: PriceFeeds<'static>,
+    pub at: Timestamp,
+    pub total_feeders: usize,
+    pub storage: &'a dyn Storage,
```
gmanev7, line 11: please fix the encapsulation bug

contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -405,7 +402,7 @@ mod tests {
         let paths = test_case();
         let tree = SupportedPairs::<Usdc>::new(paths.into_tree()).unwrap();
 
-        let mut response = tree.query_supported_pairs();
+        let mut response: Vec<_> = tree.swap_pairs_df().collect();
         response.sort_by(|a, b| a.from.cmp(&b.from));
```
gmanev7, line 406: [nit] extract the lambda fn as a nested function and reuse at line 452 as well

contracts/oracle/src/contract/alarms/iter.rs
go2bodhi: Note, in case of an error here the iteration continues. The early return is in the `try_fold` call in the `try_notify_alarms()`.

gmanev7:
LGTM, Well done!

PR #37, opened 2023-02-18
feat: Added Admin contract handling migration

### LINKED ISSUE #33: Atomically update a network to the next release
Currently, we use gov proposals directly to update contracts one by one. That labor process is error-prone and slow. In addition, it may bring services down for the time of applying all migrations.

This ticket is about migrating all new contract releases with a single proposal, i.e. atomically, as per the [policy](https://github.com/Nolus-Protocol/nolus-money-market/blob/04e4c5210c36150aec47cfed6bc3f743086721a3/release.md#networks).

Assumption: The code of the new contracts is already stored on the network.
 
A solution based on multiple migrations is impossible since cosmwasm does not support it. The key is to create a dedicated contract `Admin` that is going to orchestrate the update process. A command to the contract is executed with a gov proposal. Input parameters are the release version, and code id-s of the updated contracts with their migrate messages. Just so you know, some may not be upgraded.

First, each contract should implement a release version query. It is fed by a constant set by the cargo build.

At its core, the `Admin` begins the update process by sending migration messages to the updated contracts. Note that the administrator of the contracts should be set to the `Admin` address on the initial network initialization or post it with gov proposals.

Finally, the `Admin` validates the migrated contracts pertaining to the release by querying their versions.

KirilMihaylov:
Closes #33


contracts/admin/build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    if option_env!("LOCAL_NET_ADMIN_CONTRACT")
+        .map_or(false, |value| ["1", "y", "Y"].contains(&value))
+    {
+        println!("cargo:rustc-cfg=feature=\"admin_contract_exec\"");
```
gmanev7, line 5: two questions:
- why do we go through an env variable to enable a feature?
- I doubt we need such a shortcut of having an `execute` to `sudo` bridge. More code and potential human error are among the cons of it. We have mastered in gov proposals so would ask of removing this feature.

contracts/admin/src/state.rs
```diff
@@ -0,0 +1,82 @@
...
+
+pub(crate) fn store_migration_release(
+    storage: &mut dyn Storage,
+    migration_release: String,
+) -> StdResult<()> {
+    MIGRATION_RELEASE.save(storage, &migration_release)
+}
+
+pub(crate) fn load_and_remove_migration_release(storage: &mut dyn Storage) -> StdResult<String> {
+    let release: String = MIGRATION_RELEASE.load(storage)?;
```
gmanev7, line 26: `String` is not necessary

contracts/admin/src/state.rs
```diff
@@ -0,0 +1,82 @@
+use finance::currency::SymbolOwned;
```
gmanev7, line 1: we do not want:
-  to mix the code of different entities, and
-  separate entity definition from its persistence

That is the `cw`-style we **do not follow**. Hence please:
- define each entity in its own module, and
- provide load/store functions in OOP style, as we have been doing for most of the other code

packages/sdk/src/lib.rs
```diff
@@ -13,6 +13,11 @@ pub use cw_storage_plus;
 #[cfg(feature = "neutron")]
 pub use neutron_sdk;
 
+pub const RELEASE_VERSION: &str = env!(
+    "RELEASE_VERSION",
+    r#"No release version provided as an environment variable! Please set "RELEASE_VERSION" environment variable!"#,
+);
+
```
gmanev7, line 20: There is no point in keeping this constant in the `sdk` library. Is not `versioning` a better alternative?

contracts/treasury/src/contract.rs
```diff
@@ -43,7 +43,7 @@ pub struct MigrateMsg {}
 pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
     versioning::update_software(deps.storage, version!(CONTRACT_STORAGE_VERSION))?;
 
-    Ok(Response::default())
+    Ok(Response::default().set_data(to_binary(sdk::RELEASE_VERSION)?))
```
gmanev7, line 46: @KirilMihaylov there've been long time since you've done copy/paste in such a scale :-).

Let's replace it with a common function.

scripts/deploy-contracts-genesis.sh
```diff
@@ -122,6 +131,6 @@ _add_wasm_message() {
     local -r init_msg="$8"
 
     run_cmd "$genesis_home_dir" add-wasm-genesis-message instantiate-contract "$code_id" "$init_msg" --label "$contract_name" \
-      --run-as "$admin_addr" --no-admin $amount_flag
+      --run-as "$admin_addr" --admin "$(admin_contract_instance_addr)" $amount_flag
```
gmanev7, line 134: We do not need yet another admin, so let's unify them into being only the Admin contract. Please review the `internal/genesis.sh::generate_genesis()` from `nolus-core` for more info on how we call this code.

contracts/admin/src/msg.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub enum ExecuteMsg {
+    #[cfg(any(debug_assertions, test, feature = "admin_contract_exec"))]
+    LocalNetSudo { sudo: SudoMsg },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct MigrateContractsData {
```
gmanev7, line 50: please try to avoid such common and parasite suffixes, like `Data`.  `MigrateRelease` is all we need here

contracts/admin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    match msg {
+        #[cfg(any(debug_assertions, test, feature = "admin_contract_exec"))]
+        ExecuteMsg::LocalNetSudo { sudo: msg } => sudo(deps, env, msg),
+    }
+}
+
+#[cfg_attr(feature = "contract-with-bindings", entry_point)]
+pub fn sudo(deps: DepsMut<'_>, env: Env, msg: SudoMsg) -> Result<Response, ContractError> {
+    match msg {
+        SudoMsg::RegisterSymbolGroup {
```
gmanev7, line 91: do not get the aim of this message, could you elaborate? discuss it offline is also an option

schema/instantiate_msg.json
```diff
@@ -0,0 +1,69 @@
+{
```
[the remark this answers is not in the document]
KirilMihaylov: That was a mistake during checking in other files. Removed them.

contracts/admin/src/common.rs
```diff
@@ -0,0 +1,20 @@
...
+#[serde(rename_all = "snake_case")]
+pub struct GeneralContractsGroup<T> {
+    pub profit: T,
+    pub timealarms: T,
+    pub treasury: T,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct SpecializedContractsGroup<T> {
```
gmanev7, line 15: probably I do not get the idea, but why the dispatcher and leaser are part of it?

KirilMihaylov: I have stored all contracts directly based off a specific currency and the ones which refer to such instances into this group. Dispatcher holds reference to LPP and Oracle, which both as far as I understand have one currency and there would be an instance for each base. Same story with Leaser.

I might have misunderstood you when discussed offline.

contracts/admin/build.rs
KirilMihaylov: This was meant to simplify the process for local testing, removing the need to run migrations through government proposals.

Removed as you have asked.

contracts/admin/src/lib.rs
gmanev7: RegisterSymbolGroup -> RegisterLpnContracts

contracts/admin/src/common.rs
gmanev7: SpecializedContractsGroup -> LpnContracts

contracts/admin/src/msg.rs
```diff
@@ -0,0 +1,42 @@
...
+pub struct MigrateMsg {}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub enum SudoMsg {
+    RegisterLpnContracts {
+        symbol: SymbolOwned,
+        contracts: LpnContracts<Addr>,
+    },
+    MigrateContracts(Box<MigrateContracts>),
```
gmanev7, line 33: what's the purpose of `Box` here?

gmanev7 (commented):
In general, there are a couple of points I'd like to discuss:
- complicated implementation due to the expensive usage of Iterator-s .... in use-cases where we do not take any runtime benefit
- the ability to migrate more than one LPN set of contracts is nice but not needed, yet ... so here I'd be rather pushing back to provide it until really needed
- the reuse of parameterized types for storing the contract addresses and for passing the migration data seems artificial. The functionality is implemented somewhere far away from the data.

contracts/admin/src/msg.rs
KirilMihaylov: Discussed in DMs. Will suppress linter warning.

contracts/admin/src/state/contracts.rs
```diff
@@ -0,0 +1,58 @@
...
+            },
+        )
+        .map(|_| ())
+}
+
+pub(crate) fn load_general(storage: &dyn Storage) -> StdResult<GeneralContracts<Addr>> {
+    GENERAL_CONTRACTS.load(storage)
+}
+
+pub(crate) fn load_lpn_contracts(storage: &dyn Storage) -> LpnContractsAddrsIter<'_> {
```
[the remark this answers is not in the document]
KirilMihaylov: Point here is that it in fact returns a `Box<dyn Iterator<Item = ...>>` and the type becomes too big to spell out, thus added this typedef. Will check whether `Iterator` is implemented for `Box<dyn Iterator...>`.

contracts/admin/src/common/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::{
+    common::type_defs::MaybeMigrateContract,
+    error::{ContractError, Never},
+};
+
+pub(crate) mod type_defs;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct CodeIdWithMigrateMsg<M> {
```
gmanev7, line 20: shorten the name to `MigrationSpec`

contracts/admin/src/common/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub code_id: u64,
+    pub migrate_msg: M,
+}
+
+pub(crate) trait ValidateAddresses {
+    fn validate(&self, querier: &QuerierWrapper<'_>) -> Result<(), ContractError>;
+}
+
+pub(crate) trait MigrateContracts {
+    type GatSelf<T>;
```
gmanev7, line 30: what is `Gat` like?

contracts/admin/src/common/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    type Error: Into<ContractError> + Error;
+
+    fn migrate(
+        self,
+        migration_msgs: Self::GatSelf<MaybeMigrateContract>,
+    ) -> Result<Batch, Self::Error>;
```
gmanev7, line 37: returning Batch is enough ... which makes the `type Error` useless

contracts/admin/src/common/mod.rs
KirilMihaylov: Genetic Associated Type, in short GAT, which then converted to PascalCase gives us this.

contracts/admin/src/common/mod.rs
KirilMihaylov: Intention was to not have to change the interface when, inevitably, we go back to having the variant of missing symbols. I'll remove it if you think it's not necessary.

contracts/admin/src/common/mod.rs
gmanev7: you meant Generic not Genetic

KirilMihaylov: Checked on stabilization report and it was stabilized in Rust 1.65. Going to have to resort back to using a traditional associated type.

contracts/admin/src/common/mod.rs
gmanev7: yes, please

contracts/admin/src/common/type_defs.rs
```diff
@@ -0,0 +1,12 @@
+use super::{CodeIdWithMigrateMsg, GeneralContracts, LpnContracts};
+
+pub type MigrateContract = CodeIdWithMigrateMsg<String>;
+pub type MaybeMigrateContract = Option<MigrateContract>;
```
gmanev7, line 4: [overingeneered] while admitting that defining separate types is a good practice, and I personally am a strong believer in that, that makes sense when having different entities. In this case, we have got three copies of the same types `MigrationSpec` and `MaybeMigrationSpec`. Please get rid of these 6 extra types.

contracts/admin/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+pub fn instantiate(
+    deps: DepsMut<'_>,
+    _env: Env,
+    _info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    versioning::initialize(deps.storage, version!(CONTRACT_STORAGE_VERSION))?;
+
+    msg.general_contracts.validate(&deps.querier)?;
+    msg.lpn_contracts.validate(&deps.querier)?;
```
gmanev7, line 36: validate the message itself

contracts/admin/src/common/type_defs.rs
```diff
@@ -0,0 +1,12 @@
...
+
+pub type MigrateGeneralContract = CodeIdWithMigrateMsg<String>;
+pub type MaybeMigrateGeneralContract = Option<MigrateGeneralContract>;
+pub type MigrateGeneralContracts = GeneralContracts<MaybeMigrateGeneralContract>;
```
gmanev7, line 8: on the other hand side, the types `GeneralContracts<T>` and `LpnContracts<T>` with `T=Addr` are left not defined.

PR #45, opened 2023-02-24
ci: zip contracts dir before adding to release

ivivanov:


.github/workflows/smart_contracts.yaml
```diff
@@ -126,12 +126,22 @@ jobs:
           uses: actions/download-artifact@v3
         - name: Display structure of downloaded files
           run: ls -R
+        - name: Zip test
+          uses: vimtor/action-zip@v1.1
```
KirilMihaylov, line 130: I'd much rather personally prefer just using the `zip` command line utility instead of pulling another action into the workflow.

ivivanov: Makes sense, will do the change

PR #54, opened 2023-03-10
feat: Added preferred max loan parameter to leaser execute message

### LINKED ISSUE #32: [lease] opening a new lease should let the customer limit the borrowed amount
Currently, the process of opening a new lease borrows the maximum amount as configured with the initial liability percent. There is a requirement for providing the customer the flexibility to enter the lease with less liability than the pre-configured.

For example, if the init liability is set to 60% then for a downpayment of 100 USDC we provide a loan of 150 USDC, which is 60% of the total value of 100 + 150.

On opening a new lease, the customer should be able to provide an amount to borrow called `max_loan`. It is an optional argument. When provided, the actual amount to borrow should be equal to the min of `max_loan` and calculated per the initial liability configuration. If not, the existing algorithm applies.

For example, the customer may want to take only 100 USDC turning its lease into one with a 50% liability, or LTV ratio.

KirilMihaylov:
Closes #32

### DIFF
--- a/contracts/lease/src/api/open/mod.rs
+++ b/contracts/lease/src/api/open/mod.rs
@@ -1,11 +1,6 @@
 use serde::{Deserialize, Serialize};
 
-use finance::{
-    currency::SymbolOwned,
-    duration::Duration,
-    liability::Liability,
-    percent::{NonZeroPercent, Percent},
-};
+use finance::{currency::SymbolOwned, duration::Duration, liability::Liability, percent::Percent};
 use sdk::{
     cosmwasm_std::Addr,
     schemars::{self, JsonSchema},
@@ -34,7 +29,7 @@ pub struct NewLeaseForm {
     /// Ticker of the currency this lease will be about.
     pub currency: SymbolOwned,
     /// Maximum Loan-to-Value percentage of the new lease, optional.
-    pub max_ltv: Option<NonZeroPercent>,
+    pub max_ltv: Option<Percent>,
     /// Liability parameters
     pub liability: Liability,
     /// Loan parameters
--- a/contracts/lease/src/api/open/mod.rs
+++ b/contracts/lease/src/api/open/mod.rs
@@ -1,8 +1,10 @@
 use serde::{Deserialize, Serialize};
 
 use finance::{
-    coin::Amount as CoinAmount, currency::SymbolOwned, duration::Duration, liability::Liability,
-    percent::Percent,
+    currency::SymbolOwned,
+    duration::Duration,
+    liability::Liability,
+    percent::{NonZeroPercent, Percent},
 };
 use sdk::{
     cosmwasm_std::Addr,
@@ -31,9 +33,8 @@ pub struct NewLeaseForm {
     pub customer: Addr,
     /// Ticker of the currency this lease will be about.
     pub currency: SymbolOwned,
-    /// Optional parameter indicating preferred maximum amount of the currency
-    /// the loan should be opened with.
-    pub max_loan: Option<CoinAmount>,
+    /// Maximum Loan-to-Value percentage of the new lease, optional.
+    pub max_ltv: Option<NonZeroPercent>,
     /// Liability parameters
     pub liability: Liability,
     /// Loan parameters
--- a/contracts/lease/src/api/open/mod.rs
+++ b/contracts/lease/src/api/open/mod.rs
@@ -1,6 +1,9 @@
 use serde::{Deserialize, Serialize};
 
-use finance::{currency::SymbolOwned, duration::Duration, liability::Liability, percent::Percent};
+use finance::{
+    coin::Amount as CoinAmount, currency::SymbolOwned, duration::Duration, liability::Liability,
+    percent::Percent,
+};
 use sdk::{
     cosmwasm_std::Addr,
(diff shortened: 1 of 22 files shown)


contracts/lease/src/api/open/mod.rs
```diff
@@ -28,6 +31,9 @@ pub struct NewLeaseForm {
     pub customer: Addr,
     /// Ticker of the currency this lease will be about.
     pub currency: SymbolOwned,
+    /// Optional parameter indicating preferred maximum amount of the currency
+    /// the loan should be opened with.
+    pub max_loan: Option<CoinAmount>,
```
gmanev7, line 36: After a thorough analysis of the requirements and the "ease-of-use" and "intuitive" interface principles, let's change the input to `max_ltv: Option<Percent>` with doc string:
`Maximum Loan-to-Value percentage of the new lease, optional`

contracts/lease/src/contract/cmd/open.rs
```diff
@@ -61,18 +64,19 @@ impl<'a> WithLppLender for OpenLoanReq<'a> {
...
 
-            lpp.open_loan_req(borrow_lpn)?;
+        let borrow_lpn = self.liability.init_borrow_amount(downpayment_lpn);
 
-            Ok(Self::Output {
-                batch: lpp.into().batch,
-                downpayment,
-            })
-        }
+        lpp.open_loan_req(self.max_loan.map_or(borrow_lpn, Coin::new).min(borrow_lpn))?;
```
gmanev7, line 74: perform the calculation as part of the `.init_borrow_amount(downpayment_lpn, max_ltv)`. 

Hint:
Apply `self.initial.max(max_ltv)` as a first step in the `fn Liability::init_borrow_amount`.

Update the current and add sufficient unit tests there.

contracts/leaser/src/contract.rs
```diff
@@ -80,12 +80,13 @@ pub fn execute(
         ExecuteMsg::MigrateLeases { new_code_id } => owner_allowed_only(deps.storage, info, |s| {
             leaser::try_migrate_leases(s, new_code_id.u64())
         }),
-        ExecuteMsg::OpenLease { currency } => Borrow::with(
+        ExecuteMsg::OpenLease { currency, max_loan } => Borrow::with(
```
gmanev7, line 83: do propagate the changes in the Lease API to the Leaser API

packages/finance/src/liability/mod.rs
```diff
@@ -91,18 +91,28 @@ impl Liability {
...
 
-        // borrow = init%.of(borrow + downpayment)
-        // (100% - init%).of(borrow) = init%.of(downpayment)
-        // borrow = init% / (100% - init%) * downpayment
-        let ratio = Rational::new(self.initial, Percent::HUNDRED - self.initial);
-        ratio.of(downpayment)
+        #[cfg(debug_assertions)]
+        if let Some(max_ltv) = max_ltv {
+            debug_assert!(max_ltv.percent() < Percent::HUNDRED);
+        }
```
gmanev7, line 106: such debug assert makes sense on the value = min of the two ... it's more important to that... though self.initial is already check that is less than 100%

packages/finance/src/liability/mod.rs
```diff
@@ -91,18 +91,28 @@ impl Liability {
...
+
+        let percent: Percent = max_ltv.map_or(self.initial, |percent: NonZeroPercent| {
```
gmanev7, line 108: [repeating this comment again and again] names given after their type carry little info on what they are like. Please rename `percent` to `initial_ltv`

packages/finance/src/percent.rs
gmanev7, line 1: strongly against this new type. The benefits are negligible whereas the added noise to the code is significant.

contracts/leaser/src/cmd/quote.rs
```diff
@@ -218,7 +218,7 @@ where
             return Err(ContractError::ZeroDownpayment {});
         }
 
-        let borrow = self.liability.init_borrow_amount(downpayment_lpn);
+        let borrow = self.liability.init_borrow_amount(downpayment_lpn, None);
```
gmanev7, line 221: this is not correct. 

we have to take `max_ltv` as an input quote parameter, otherwise the calculation would not be accurate

gmanev7: we have agreed to fix it

packages/finance/src/liability/mod.rs
```diff
@@ -91,18 +91,27 @@ impl Liability {
...
 
-        // borrow = init%.of(borrow + downpayment)
-        // (100% - init%).of(borrow) = init%.of(downpayment)
-        // borrow = init% / (100% - init%) * downpayment
-        let ratio = Rational::new(self.initial, Percent::HUNDRED - self.initial);
-        ratio.of(downpayment)
+        #[cfg(debug_assertions)]
+        if let Some(max_ltv) = max_ltv {
+            debug_assert!(max_ltv > Percent::ZERO);
+        }
```
gmanev7, line 106: agreed not to add artificial limits, remove it and add a test with `max_ltv = Zero`

gmanev7:
LGTM

PR #57, opened 2023-03-14
feat(oracle)!: simplify the api for prices query

go2bodhi:


contracts/oracle/src/contract/oracle/mod.rs
gmanev7, line 1: why don't we keep  `&dyn Storage` as a member variable to simplify the member functions' API

go2bodhi: We agreed offline to postpone this change

PR #58, opened 2023-03-15
refactor!: Moved privileged execute commands to sudo handlers

### LINKED ISSUE #36: Move admin-priviledged actions from `execute` to `sudo`
Most of the privileged actions are permitted only to the contracts admin. While there is protection from unauthorized access, moving them under `sudo` operations only would bring these benefits:
- an extra level of defense provided by `wasmd`
- state explicitly that they are allowed only for super users

KirilMihaylov:
Closes #36

### DIFF
--- a/contracts/dispatcher/src/contract.rs
+++ b/contracts/dispatcher/src/contract.rs
@@ -158,15 +158,18 @@ fn try_dispatch(deps: DepsMut<'_>, env: Env, info: MessageInfo) -> Result<Respon
 #[cfg(test)]
 mod tests {
     use finance::percent::Percent;
-    use sdk::cosmwasm_std::{
-        coins, from_binary,
-        testing::{mock_dependencies_with_balance, mock_env, mock_info},
-        Addr, DepsMut,
+    use sdk::{
+        cosmwasm_ext::Response,
+        cosmwasm_std::{
+            coins, from_binary,
+            testing::{mock_dependencies_with_balance, mock_env, mock_info},
+            Addr, Attribute, DepsMut,
+        },
+        testing::customized_mock_deps_with_contracts,
     };
-    use sdk::testing::customized_mock_deps_with_contracts;
 
-    use crate::contract::sudo;
     use crate::{
+        contract::sudo,
         msg::{ConfigResponse, InstantiateMsg, QueryMsg, SudoMsg},
         state::reward_scale::{Bar, RewardScale, TotalValueLocked},
     };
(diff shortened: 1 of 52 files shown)


contracts/dispatcher/src/contract.rs
```diff
@@ -90,18 +90,21 @@ pub fn execute(
...
 }
 
+#[cfg_attr(feature = "contract-with-bindings", entry_point)]
+pub fn sudo(deps: DepsMut<'_>, _env: Env, msg: SudoMsg) -> Result<Response, ContractError> {
+    match msg {
+        SudoMsg::Config { cadence_hours } => try_config(deps.storage, cadence_hours),
+    }
+}
+
 pub fn try_config(
```
gmanev7, line 104: `pub` not needed

contracts/dispatcher/src/contract.rs
```diff
@@ -229,30 +232,34 @@ mod tests {
...
-        let res = execute(deps.as_mut(), mock_env(), unauth_info, msg);
-        match res {
-            Err(ContractError::Unauthorized(..)) => {}
-            _ => panic!("Must return unauthorized error"),
-        }
-
-        let auth_info = mock_info("creator", &coins(2, "token"));
-        let msg = ExecuteMsg::Config { cadence_hours: 12 };
-        let _res = execute(deps.as_mut(), mock_env(), auth_info, msg).unwrap();
+        let _auth_info = mock_info("creator", &coins(2, "token"));
```
gmanev7, line 235: [all occurences] why do we keep them?

contracts/dispatcher/src/contract.rs
```diff
@@ -229,30 +232,34 @@ mod tests {
...
+        drop(
```
gmanev7, line 236: [everywhere] let's not use `drop` - the message it sends differs from our intention. Instead check the result for everything you might come up with, e.g. no messages, etc...

contracts/leaser/src/msg.rs
```diff
@@ -28,6 +28,19 @@ pub struct MigrateMsg {}
...
 #[serde(rename_all = "snake_case")]
 pub enum ExecuteMsg {
+    OpenLease {
+        currency: SymbolOwned,
+        #[serde(default)]
+        max_ltv: Option<Percent>,
+    },
+    Sudo {
+        msg: SudoMsg,
+    },
```
gmanev7, line 38: as commented above, replace it with MigrateLeases

gmanev7:
last but not least, please bump the package's major version components. After all, we do breaking API changes

gmanev7:
LGTM, merge it once pass all the checks

PR #61, opened 2023-03-23
timealarms: single alarm per address

go2bodhi:


Cargo.lock
```diff
@@ -1250,6 +1250,20 @@ dependencies = [
...
+ "cosmwasm-std",
+ "platform",
+ "schema",
+ "sdk",
+ "serde",
+ "serde-json-wasm 0.5.0",
+ "thiserror",
+ "timealarms",
+]
+
```
gmanev7, line 1266: where is the code of this new package?

contracts/timealarms/src/alarms.rs
```diff
@@ -30,28 +41,42 @@ impl TimeAlarms {
...
-        let dispatcher = Self::TIME_ALARMS.notify(storage, dispatcher, ctime, max_count)?;
+        let dispatcher = self
+            .time_alarms
+            .alarms_selection(storage, ctime)
+            .take(max_count.try_into()?)
+            .enumerate()
+            .try_fold(
+                OracleAlarmDispatcher::new(),
+                |dispatcher, (id, alarm)| -> Result<OracleAlarmDispatcher, ContractError> {
+                    dispatcher.send_to(id.try_into()?, alarm?.0)
```
gmanev7, line 62: I believe we should stop using generated Id-s. Any constant would suffice.

contracts/timealarms/src/dispatcher.rs
```diff
@@ -25,16 +26,17 @@ impl OracleAlarmDispatcher {
             emitter: Emitter::of_type(EVENT_TYPE),
         }
     }
-}
-impl AlarmDispatcher for OracleAlarmDispatcher {
-    fn send_to(mut self, id: Id, addr: Addr) -> Result<Self, AlarmError> {
+
+    pub fn send_to(mut self, id: Id, addr: Addr) -> Result<Self, ContractError> {
```
gmanev7, line 30: why do we need it `pub`?

go2bodhi: earlier it was in the impl of a trait, so it was `pub`. It's used outside in the `timealarms::alarms`, don't see the problem.

contracts/timealarms/src/alarms.rs
gmanev7: @go2bodhi I do not see it resolved

go2bodhi: forgot to commit

packages/time-oracle/src/migrate_v1.rs
```diff
@@ -0,0 +1,158 @@
...
+        alarms.add(storage, addr3, t3).unwrap();
+        alarms.add(storage, addr4, t4).unwrap();
+
+        // multiple alarms per address(5) + index(5) + next_id(1)
+        assert_eq!(11, storage.range(None, None, Order::Ascending).count());
+
+        alarms.migrate(storage).unwrap();
+
+        // single alarm per address(4) + index(4)
+        assert_eq!(8, storage.range(None, None, Order::Ascending).count());
```
gmanev7, line 156: please add more checks on the alarms post-migration, e.g. what alarms for what addresses are there

packages/time-oracle/src/migrate_v1.rs
```diff
@@ -0,0 +1,158 @@
...
+
+use sdk::{
+    cosmwasm_std::{Addr, Order, StdError, Storage, Timestamp},
+    cw_storage_plus::{Index, IndexList, IndexedMap, Item, MultiIndex},
+};
+
+use super::Alarms;
+use crate::AlarmError;
+
+type TimeSeconds = u64;
```
gmanev7, line 11: if it has not been changed then use the one from v1

contracts/timealarms/src/alarms.rs
```diff
@@ -30,28 +42,52 @@ impl TimeAlarms {
...
+    pub fn migrate_v1(storage: &mut dyn Storage) -> Result<(), ContractError> {
+        AlarmsOld::new(
+            Self::ALARMS_NAMESPACE,
+            Self::ALARMS_IDX_NAMESPACE,
+            "alarms_next_id",
+        )
+        .migrate(storage)?;
+
+        Ok(())
     }
```
gmanev7, line 91: let's not mix the migration code with the rest

The aim is to locate it in module(s) that are going to be removed once the migration is over.

contracts/timealarms/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "timealarms"
-version = "0.2.4"
+version = "0.3.0"
```
gmanev7, line 3: there is no API change, so let's stay with the same major version

gmanev7: since it introduces a response so disregard

PR #68, opened 2023-03-31
Alarm replies

### LINKED ISSUE #60: Market alarms recipients do not return their address
The recipient should return that info in the response in order to allow the cosmwasm reply mechanism to remove them.

go2bodhi:
Closes #60

### DIFF
--- a/contracts/admin/src/lib.rs
+++ b/contracts/admin/src/lib.rs
@@ -3,13 +3,14 @@ use platform::response;
 use sdk::cosmwasm_std::entry_point;
 use sdk::{
     cosmwasm_ext::Response,
-    cosmwasm_std::{ensure_eq, DepsMut, Env, MessageInfo, Reply, StdResult},
+    cosmwasm_std::{ensure_eq, DepsMut, Env, MessageInfo, Reply},
 };
 use versioning::{version, VersionSegment};
 
 use self::{
     error::ContractError,
     msg::{InstantiateMsg, MigrateMsg, SudoMsg},
+    result::ContractResult,
     state::{contracts as state_contracts, migration_release},
 };
 
@@ -17,6 +18,7 @@ pub mod common;
 pub mod error;
 pub mod migrate_contracts;
 pub mod msg;
+pub mod result;
 pub mod state;
 
 // version info for migration info
@@ -29,7 +31,7 @@ pub fn instantiate(
     _env: Env,
     _info: MessageInfo,
     msg: InstantiateMsg,
-) -> Result<Response, ContractError> {
+) -> ContractResult<Response> {
     versioning::initialize(deps.storage, version!(CONTRACT_STORAGE_VERSION))?;
 
     msg.validate(&deps.querier)?;
@@ -40,14 +42,16 @@ pub fn instantiate(
 }
 
 #[cfg_attr(feature = "contract-with-bindings", entry_point)]
-pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
+pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> ContractResult<Response> {
     versioning::update_software(deps.storage, version!(CONTRACT_STORAGE_VERSION))?;
 
-    response::response(versioning::release()).map(Into::into)
+    response::response(versioning::release())
+        .map(Into::into)
+        .map_err(Into::into)
 }
 
 #[cfg_attr(feature = "contract-with-bindings", entry_point)]
-pub fn sudo(deps: DepsMut<'_>, env: Env, msg: SudoMsg) -> Result<Response, ContractError> {
+pub fn sudo(deps: DepsMut<'_>, env: Env, msg: SudoMsg) -> ContractResult<Response> {
     match msg {
         SudoMsg::MigrateContracts(migrate_contracts) => {
             migrate_contracts::migrate(deps.storage, env.contract.address, migrate_contracts)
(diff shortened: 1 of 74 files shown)


contracts/lease/src/contract/state/mod.rs
```diff
@@ -131,6 +128,11 @@ impl Response {
             next_state: next_state.into(),
         }
     }
+
+    pub fn attach_alarm_response(mut self, env: &Env) -> ContractResult<Self> {
+        self.cw_response = self.cw_response.set_data(to_binary(&env.contract.address)?);
+        Ok(self)
+    }
```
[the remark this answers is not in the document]
gmanev7: there is an fn `platform::response::response` that is a good candidate

contracts/lease/src/contract/state/ica_connector.rs
```diff
@@ -82,7 +81,7 @@ impl<const PRECONNECTABLE: bool, Connectee> Enterable for IcaConnector<PRECONNEC
 where
     Connectee: IcaConnectee + DexConnectable,
 {
-    fn enter(&self, _deps: Deps<'_>, _env: Env) -> ContractResult<Batch> {
+    fn enter(&self, _deps: Deps<'_>, _env: &Env) -> ContractResult<Batch> {
```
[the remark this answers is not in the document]
KirilMihaylov: Motivation is that we do **not** need an owned copy of the data, thus I think we do follow it.

The other thing is that, if we do pass the ownership, the data needs to be cloned which might in turn prove useless if an error occurs during the execution.

packages/platform/src/response.rs
```diff
@@ -1,12 +1,31 @@
...
-    cosmwasm_ext::Response,
-    cosmwasm_std::{to_binary, StdResult},
+    cosmwasm_ext::Response as CwResponse,
+    cosmwasm_std::{to_binary, Binary, StdResult},
 };
-use serde::Serialize;
 
-pub fn response<T>(resp: &T) -> StdResult<Response>
+#[inline]
+pub fn response<T>(response: &T) -> StdResult<Response>
```
gmanev7, line 9: [here and below] let's keep the `Std` prefix for the Rust Std namespace and pertaining to that space types. For uniformity let's use the `Cw` prefix.

contracts/admin/src/lib.rs
```diff
@@ -40,10 +40,10 @@ pub fn instantiate(
 }
 
 #[cfg_attr(feature = "contract-with-bindings", entry_point)]
-pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
+pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
```
gmanev7, line 43: what is the motivation for switching to the cosmwasm error type?

contracts/dispatcher/src/contract.rs
```diff
@@ -75,12 +75,12 @@ pub fn instantiate(
...
 
 #[cfg_attr(feature = "contract-with-bindings", entry_point)]
-pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
+pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> StdResult<Response> {
     versioning::update_software(deps.storage, version!(CONTRACT_STORAGE_VERSION))?;
 
     SingleUserAccess::remove_contract_owner(deps.storage);
 
-    response::response(versioning::release()).map_err(Into::into)
+    response(versioning::release()).map(Into::into)
```
gmanev7, line 83: let's follow the function calling convention by providing the function module

packages/platform/src/response.rs
```diff
@@ -1,12 +1,31 @@
...
 where
     T: Serialize + ?Sized,
 {
-    Ok(Response::new().set_data(to_binary(resp)?))
+    response_with_messages(CwResponse::new(), response)
+}
+
+pub fn response_with_messages<T, U>(messages: T, response: &U) -> StdResult<Response>
```
gmanev7, line 16: cosmetic: change the order of fn parameters to reflect the fn name, first the response, then the messages

contracts/lease/src/contract/state/controller.rs
```diff
@@ -130,9 +130,9 @@ pub fn execute(
                  cw_response,
                  next_state,
              }| {
-                super::save(deps.storage, &next_state)?;
-
-                Ok(cw_response)
+                super::save(deps.storage, &next_state)
+                    .map(move |()| cw_response)
+                    .map_err(Into::into)
```
gmanev7, line 135: two arguments against applying the functional style here:
- only this occurrence in this module changed, there are at least a few more
- the previous code was shorter and easier to read

contracts/lease/src/contract/state/ica_connector.rs
```diff
@@ -21,7 +20,7 @@ use crate::{
 use super::{ica_post_connector::Postpone, State};
 
 pub(crate) trait Enterable {
-    fn enter(&self, deps: Deps<'_>, _env: Env) -> ContractResult<Batch>;
+    fn enter(&self, deps: Deps<'_>, _env: &Env) -> ContractResult<Batch>;
```
gmanev7, line 23: we've agreed to postpone such a change

contracts/admin/src/lib.rs
KirilMihaylov: Conversion to our error type is not needed as of now. This change was only because of minimization.

gmanev7: we do not want to change the interface later when we have to implement the method. As you may see, all methods at this point return StdResult<Response, ContractError>. If you wanna minimize, then define a `type ContractResult<T> = StdResult<T, ContractError>` as we did at the Lease, for example

gmanev7: that applies to all occurrences, pls

gmanev7: in all contracts

PR #71, opened 2023-04-06
feat(treasury): Configure the treasury contract on instantiation

### LINKED ISSUE #67: [Treasury] Configure the rewards dispatcher at genesis
Currently, rewards cannot be dispatched because the dispatcher is not configured in the treasury. It is possible to do post genesis but would require a gov proposal.

The aim of this task is to have that configured at the genesis time. Add the dispatcher address as a parameter of the Treasury's InitMsg.

KirilMihaylov:
Closes #67

### DIFF
--- a/contracts/treasury/src/contract.rs
+++ b/contracts/treasury/src/contract.rs
@@ -3,8 +3,8 @@ use currency::native::Nls;
 use finance::coin::Coin;
 use platform::{
     bank::{self, BankAccount},
-    batch::Batch,
-    response::{self},
+    message::Response as PlatformResponse,
+    response,
 };
 #[cfg(feature = "contract-with-bindings")]
 use sdk::cosmwasm_std::entry_point;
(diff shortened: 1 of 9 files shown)


contracts/treasury/src/contract.rs
```diff
@@ -28,10 +28,12 @@ pub fn instantiate(
     deps: DepsMut<'_>,
     _env: Env,
     _info: MessageInfo,
-    _msg: InstantiateMsg,
+    msg: InstantiateMsg,
 ) -> ContractResult<CwResponse> {
     versioning::initialize(deps.storage, version!(CONTRACT_STORAGE_VERSION))?;
 
+    try_configure_reward_dispatcher(deps.storage, msg.rewards_dispatcher)?;
```
gmanev7, line 35: is it doable to validate the rewards_dispatcher address as we do on `sudo`, `platform::contract::validate_addr(&deps.querier, &rewards_dispatcher)?;` ? Check pls, the order of contract instantiations.

gmanev7 (commented):
looks good!

contracts/treasury/src/contract.rs
KirilMihaylov: We'd have to sacrifice either of the checks in the contract instantiations because `dispatcher` also checks the validity of the `treasury` address.

gmanev7: ok, let's keep it as it is

PR #80, opened 2023-04-26
feat: Profit contract's buy-back

### LINKED ISSUE #2: [Profit] Buy-back NLS (all currencies)
https://docs.google.com/document/d/1IikdVfwg8q6XpE55mh5WOTdlecE6HuqnYgvpxv49iZU/edit#heading=h.a58x92veo5qm

KirilMihaylov:
Closes #2

### DIFF
--- a/packages/platform/Cargo.toml
+++ b/packages/platform/Cargo.toml
@@ -6,7 +6,7 @@ edition.workspace = true
 license.workspace = true
 
 [features]
-testing = []
+testing = ["sdk/testing"]
 
 [dependencies]
 finance = { workspace = true }
@@ -20,3 +20,4 @@ serde-json-wasm = { workspace = true }
 [dev-dependencies]
 currency = { workspace = true }
 finance = { workspace = true, features = ["testing"] }
+sdk = { workspace = true, features = ["testing"] }


contracts/lease/src/contract/endpoins.rs
```diff
@@ -45,6 +47,7 @@ pub fn instantiate(
         .map(response::response_only_messages)
 }
 
+#[cfg(feature = "migration")]
```
gmanev7, line 50: this feature is about storage migration, so do not diable the migration at all

contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -67,22 +70,25 @@ impl Active {
     }
 
     fn try_repay(self, deps: Deps<'_>, env: Env, info: MessageInfo) -> ContractResult<Response> {
-        let payment = bank::may_received::<PaymentGroup, _>(
+        let payment = never!(bank::may_received::<PaymentGroup, _>(
```
gmanev7, line 73: we do not prefer macro-ses when regular functions would work. look at dex::never

packages/platform/src/bank.rs
```diff
@@ -16,19 +16,19 @@ use crate::{
     error::{Error, Result},
 };
 
+pub type BalancesResult<Cmd> =
+    StdResult<StdResult<Option<<Cmd as WithCoin>::Output>, <Cmd as WithCoin>::Error>, Error>;
+
```
gmanev7, line 21: why this double-nesting?

gmanev7:
this MR should be decomposed to at least a couple of MRs before we proceed, please do

tests/src/common/test_case.rs
```diff
@@ -283,17 +286,70 @@ where
...
+        let profit_addr: &Addr = self
+            .profit_addr
+            .insert(ProfitWrapper::default().instantiate(
+                &mut self.app,
+                cadence_hours,
+                self.treasury_addr.clone().unwrap(),
+                self.oracle.clone().unwrap(),
+                self.timealarms.clone().unwrap(),
+                CONNECTION_ID.into(),
+            ));
```
gmanev7, line 300: rebase needed to take this `insert` out?

tests/src/lpp_tests.rs
```diff
@@ -927,7 +939,14 @@ fn compare_lpp_states() {
...
     message_receiver.assert_empty();
 
-    let profit = ProfitWrapper::default().instantiate(&mut app, 24, treasury, time_alarms.clone());
+    let profit = ProfitWrapper::default().instantiate(
+        &mut app,
+        24,
+        treasury,
+        market_oracle.clone(),
+        time_alarms.clone(),
+        "connection-1".into(),
```
gmanev7, line 948: [multiple instances] define a const, pls

gmanev7 (commented):
do you mind adding a test for amounts in other currencies?

contracts/profit/src/contract.rs
```diff
@@ -61,28 +63,91 @@ pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> ContractResult
...
+fn try_time_alarm(deps: DepsMut<'_>, env: Env) -> ContractResult<MessageResponse> {
+    let state: State = State::load(deps.storage)?;
+
+    let DexResponse::<State> {
+        response,
+        next_state,
+    } = match state.on_time_alarm(deps.as_ref(), env) {
+        DexResult::Continue(result) => result?,
+        DexResult::Finished(response) => response,
+    };
```
gmanev7, line 146: this is already implemented, Look at packages/dex/src/response.rs for `impl<H, Err> From<Result<H>> for StdResult<Response<H>, Err>`. Then use `Into::into` instead.

tests/src/common/test_case.rs
KirilMihaylov: Actually no. That's needed as I later use the address and there's no need to do an `unwrap`.

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,231 @@
...
+            StateEnum::OpenIca(_) => Err(ContractError::unsupported_operation(
+                "Querying configuration is not allowed during ICA opening process.",
+            )),
+            StateEnum::Idle(idle) => Ok(idle.config()),
+            StateEnum::BuyBack(_) => Err(ContractError::unsupported_operation(
+                "Querying configuration is not allowed during buy-back.",
+            )),
+        }
+    }
+}
```
gmanev7, line 120: Avoid using  `match` expressions to implement polymorphic behavior.  Use trait(s) implemented for all states, for example, `UpdateConfig`, and call its methods. The particular state would decide how to handle such requests.

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,231 @@
...
+    fn on_response(self, data: Binary, deps: Deps<'_>, env: Env) -> DexResult<Self> {
+        match self.0 {
+            StateEnum::OpenTransferChannel(transfer) => {
+                transfer.on_response(data, deps, env).map_into()
+            }
+            StateEnum::OpenIca(ica) => ica.on_response(data, deps, env).map_into(),
+            StateEnum::Idle(idle) => idle.on_response(data, deps, env).map_into(),
+            StateEnum::BuyBack(buy_back) => match buy_back.on_response(data, deps, env) {
+                DexResult::Continue(result) => DexResult::Continue(result.map(state_machine::from)),
+                DexResult::Finished(result) => DexResult::Continue(result),
```
gmanev7, line 193: [few occurences] this is already implemented, Look at packages/dex/src/response.rs for `pub fn map_into<RFrom, HTo, RTo, S>(self) -> Result<HTo>`

contracts/profit/src/contract.rs
KirilMihaylov: Actually not implemented for this case as it's specialized for cases where `Handler::SwapResult` is a `std::result::Result`, not `dex::response::Response`.

Implemented a new method for that case, though.

contracts/profit/src/state/mod.rs
KirilMihaylov: Actually, it isn't. There it's identity mapping, while here I am mapping both to `Continue` because the types don't match up.

tests/src/common/test_case.rs
gmanev7: I do not understand why we need to change the `self.profit_addr` in the `init_profit` method. We do want to initialize it.

KirilMihaylov: I use it to initialize the contract from start to end. I'm not changing anything about `self.profit_addr` apart from setting it.

contracts/profit/src/msg.rs
```diff
@@ -19,12 +22,17 @@ pub struct MigrateMsg {}
...
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum SudoMsg {
-    Config { cadence_hours: u16 },
+    SetupDex {
+        dex: ConnectionParams,
+        oracle: Addr,
+        time_alarms: Addr,
```
gmanev7, line 34: obsolete, delete it

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,229 @@
...
+pub(crate) trait ConfigManagement
+where
+    Self: Sized,
+{
+    fn try_update_config(self, cadence_hours: u16) -> ContractResult<Self>;
+
+    fn try_query_config(&self) -> ContractResult<ConfigResponse>;
+}
+
+pub(crate) trait ProfitMessageHandler
```
gmanev7, line 41: rename to  SetubDexHandler

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,229 @@
...
+where
+    Self: Handler,
```
gmanev7, line 43: remove this to detach it from `Handler`

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,229 @@
...
+{
+    fn confirm_open(
+        self,
+        deps: Deps<'_>,
+        env: Env,
+        _channel: Ics20Channel,
+        counterparty_version: String,
```
gmanev7, line 50: change the name to `setup_dex`, parameters to (self, params: dex::connection::ConnectionParams), and result to std::result::Result<platform::state_machine::Response<State>, ContractError>

contracts/profit/src/state/open_ica.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::{error::ContractError, msg::ConfigResponse, result::ContractResult};
+
+use super::{idle::Idle, Config, ConfigManagement, IcaConnector, ProfitMessageHandler, State};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(super) struct OpenIca {
+    config: Config,
+    dex: ConnectionParams,
+    oracle: OracleRef,
+    time_alarms: TimeAlarmsRef,
```
gmanev7, line 16: move them to Config

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,229 @@
...
+    ) -> ContinueResult<Self> {
+        self.on_open_ica(counterparty_version, deps, env)
```
gmanev7, line 52: return error that the operation is not supported at this state

contracts/profit/src/state/mod.rs
```diff
@@ -1 +1,229 @@
...
+        self,
+        counterparty_version: String,
+        deps: Deps<'_>,
+        env: Env,
+    ) -> ContinueResult<Self> {
+        if let StateEnum::OpenIca(ica) = self.0 {
+            ica.on_open_ica(counterparty_version, deps, env)
+        } else {
+            unimplemented!()
+        }
```
gmanev7, line 179: forward the call to all variants, as it's done for the other methods

contracts/profit/src/contract.rs
```diff
@@ -61,179 +63,91 @@ pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> ContractResult
...
+            ..
+        } => state.confirm_open(
+            deps.as_ref(),
+            env,
+            Ics20Channel {
+                local_endpoint: channel_id,
+                remote_endpoint: counterparty_channel_id,
+            },
+            counterparty_version,
+        )?,
```
gmanev7, line 120: here we have to route the invocation to `DexSetupHandler` or `dex::Handler` depending on the input arguments, as follows:
- if `counterparty_version` is provided then call the latter, otherwise
- call the former mapping `port_id` as `ConnectionParams::connection_id`, `channel_id` as `Ics20Channel::local_endpoint`, and `counterparty_channel_id` as `Ics20Channel::remote_endpoint`

contracts/profit/src/msg.rs
```diff
@@ -9,7 +10,9 @@ use sdk::{
 pub struct InstantiateMsg {
     pub cadence_hours: u16,
     pub treasury: Addr,
+    pub oracle: Addr,
     pub timealarms: Addr,
+    pub connection_id: String,
```
gmanev7, line 15: remove it from here,, we'll provide it later with OpenAck

gmanev7 (commented):
this review is focused on the handler traits .... with the result types and their mapping left for the next iteration

contracts/profit/src/contract.rs
KirilMihaylov: Is that handler not called automatically by the `wasmd` modules?

gmanev7: it is, with ICA account passed in `counterparty_version`. We'll call it via proposal when setting up DEX with out passing `counterparty_version`

tests/src/common/test_case.rs
gmanev7: the code would be more readable to keep the value in a local variable, and then assign a copy to the instance variable

gmanev7: it's of your preference so resolving it

packages/dex/src/response.rs
```diff
@@ -84,6 +83,18 @@ where
...
+    H: Handler<SwapResult = Response<H>>,
+{
+    pub fn continue_or_ok(self) -> StdResult<Response<H>, Error> {
+        match self {
+            Result::Continue(cont_res) => cont_res,
+            Result::Finished(finish_res) => Ok(finish_res),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
KirilMihaylov: I did try implementing it, but AFAIR the compiler complained about the two implementations conflicting with each other.

contracts/profit/src/contract.rs
```diff
@@ -61,179 +64,97 @@ pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> ContractResult
...
 ) -> ContractResult<CwResponse> {
     match msg {
         ExecuteMsg::TimeAlarm {} => {
+            SingleUserAccess::load(deps.storage, crate::access_control::TIMEALARMS_NAMESPACE)?
+                .check_access(&info.sender)?;
+
             let alarm_recepient = env.contract.address.clone();
-            try_transfer(deps.as_ref(), env, info).and_then(|resp| {
+
+            try_time_alarm(deps.branch(), env).and_then(|resp| {
```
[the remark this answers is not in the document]
KirilMihaylov: Yes, because down in `try_time_alarm` I save the new state, thus I need `&mut dyn Storage` which in turn requires `DepsMut`.

contracts/profit/src/state/buy_back.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn oracle(&self) -> &OracleRef {
+        self.config.oracle()
+    }
+
+    fn time_alarm(&self) -> &TimeAlarmsRef {
+        self.config.time_alarms()
+    }
+
+    fn out_currency(&self) -> Symbol<'_> {
+        Nls::DEX_SYMBOL
```
gmanev7, line 72: Ticker

contracts/profit/src/state/buy_back.rs
```diff
@@ -0,0 +1,132 @@
...
+            if early_exit {
+                return Ok(IterState::Incomplete);
+            }
+
+            if matches!(visitor.visit(coin)?, IterNext::Stop) {
+                early_exit = true;
+            }
+        }
+
+        Ok(IterState::Complete)
```
gmanev7, line 91: what about using Iterator::find?

contracts/profit/src/state/buy_back.rs
```diff
@@ -0,0 +1,133 @@
...
+        Err(ContractError::UnsupportedOperation(String::from(
+            "Configuration changes are not allowed during ICA opening process.",
+        )))
+    }
+
+    fn try_query_config(&self) -> ContractResult<ConfigResponse> {
+        Err(ContractError::UnsupportedOperation(String::from(
+            "Querying configuration is not allowed during buy-back.",
+        )))
+    }
```
gmanev7, line 128: with the new design of the state, it looks trivial to implement, a?

contracts/profit/src/state/idle.rs
```diff
@@ -0,0 +1,153 @@
...
+        B: BankAccount,
+    {
+        let balance_nls: Coin<Nls> = account.balance()?;
+
+        if balance_nls.is_zero() {
+            Ok(None)
+        } else {
+            Profit::transfer_nls(account, env, self.config.treasury()).map(Some)
+        }
+    }
```
gmanev7, line 49: It seems unnecessary to send the amount in Nls prior to receiving the exchanged amount past the dex finishes. Suggest removing it.

contracts/profit/src/state/open_ica.rs
```diff
@@ -0,0 +1,55 @@
...
+            "Configuration changes are not allowed during ICA opening process.",
+        )))
+    }
+
+    fn try_query_config(&self) -> ContractResult<ConfigResponse> {
+        Err(ContractError::UnsupportedOperation(String::from(
+            "Querying configuration is not allowed during ICA opening process.",
+        )))
+    }
+}
```
gmanev7, line 51: with the new design of the state, it looks trivial to implement, a?

contracts/profit/src/state/buy_back.rs
gmanev7: a, no thanks!

contracts/profit/src/msg.rs
```diff
@@ -9,6 +9,7 @@ use sdk::{
 pub struct InstantiateMsg {
     pub cadence_hours: u16,
     pub treasury: Addr,
+    pub oracle: Addr,
```
gmanev7, line 12: have you updated the initial message in the script?

KirilMihaylov: Good catch! Updated in bb9914ef0bf8a0823c3a5b6bb430ca13838be4c2

PR #86, opened 2023-05-02
refactor(test): Refactored tests to account for all Neutron messages

KirilMihaylov:


tests/src/common/test_case.rs
```diff
@@ -246,32 +268,47 @@ where
...
+        const CONNECTION_ID: &str = "dex-connection";
+
+        let _: &Addr = self
+            .profit_addr
+            .insert(ProfitWrapper::default().instantiate(
+                &mut self.app,
+                cadence_hours,
+                self.treasury_addr.clone().unwrap(),
+                self.timealarms.clone().unwrap(),
+            ));
```
gmanev7, line 296: why do we use `insert`? why not assign a value as it was?

KirilMihaylov: My bad. That's a leftover from the split from buy-back.

gmanev7 (approved):

PR #88, opened 2023-05-02
Added a check that only the owner can execute 'close'

maneva3:


contracts/lease/src/contract/state/paid/mod.rs
```diff
@@ -39,6 +39,9 @@ impl Handler for Active {
         match msg {
             ExecuteMsg::Repay() => handler::err("repay", deps.api),
             ExecuteMsg::Close() => {
+                if self.lease.lease.customer != _info.sender {
```
gmanev7, line 42: [nit] rename the variable back when start using it

gmanev7 (approved):
LGTM

PR #94, opened 2023-05-12
refactor: Time alarms delivery

### LINKED ISSUE #81: [Timealarms] Confirm the correctness of alarm rescheduling during alarm delivery
A time alarm is deleted once its delivery is successful. Since there is only one alarm per recipient, it seems, that any new alarms a recipient schedules during alarm delivery would be deleted.

As per the actor model, the reply is called after all nested messages are executed. Such a message might be adding a new alarm.

Here is a test to start with: timealarms_tests::reschedule_alarm

KirilMihaylov:
Closes #81

### DIFF
--- a/contracts/timealarms/src/alarms.rs
+++ b/contracts/timealarms/src/alarms.rs
@@ -48,7 +48,12 @@ impl<'r> TimeAlarms<'r> {
 impl<'r> TimeAlarmsMut<'r> {
     pub fn new(storage: &'r mut dyn Storage) -> Self {
         Self {
-            time_alarms: AlarmsMut::new(storage, ALARMS_NAMESPACE, ALARMS_IDX_NAMESPACE, IN_DELIVERY_NAMESPACE),
+            time_alarms: AlarmsMut::new(
+                storage,
+                ALARMS_NAMESPACE,
+                ALARMS_IDX_NAMESPACE,
+                IN_DELIVERY_NAMESPACE,
+            ),
         }
     }
 
--- a/contracts/timealarms/src/alarms.rs
+++ b/contracts/timealarms/src/alarms.rs
@@ -51,12 +51,6 @@ impl<'r> TimeAlarmsMut<'r> {
         }
     }
 
-    pub fn remove(&mut self, addr: Addr) -> Result<(), ContractError> {
-        self.time_alarms.remove(addr)?;
-
-        Ok(())
-    }
-
     pub fn try_add(
         &mut self,
         querier: &QuerierWrapper<'_>,
--- a/contracts/timealarms/src/alarms.rs
+++ b/contracts/timealarms/src/alarms.rs
@@ -18,11 +18,11 @@ const REPLY_ID: Id = 0;
 const EVENT_TYPE: &str = "timealarm";
 
 pub(super) struct TimeAlarms<'r> {
-    time_alarms: Alarms<'r, 'static>,
+    time_alarms: Alarms<'r>,
 }
 
 pub(super) struct TimeAlarmsMut<'r> {
-    time_alarms: AlarmsMut<'r, 'static>,
+    time_alarms: AlarmsMut<'r>,
 }
 
 impl<'r> TimeAlarms<'r> {
--- a/contracts/timealarms/src/alarms.rs
+++ b/contracts/timealarms/src/alarms.rs
@@ -4,7 +4,7 @@ use platform::{
     message::Response as MessageResponse,
 };
 use sdk::cosmwasm_std::{Addr, Env, QuerierWrapper, Storage, Timestamp};
-use time_oracle::{Alarms, AlarmsMut};
+use time_oracle::{Alarms, AlarmsMut, AlarmsSelection as _};
 
 use crate::{
     msg::{AlarmsCount, AlarmsStatusResponse, ExecuteAlarmMsg},
@@ -18,7 +18,7 @@ const REPLY_ID: Id = 0;
 const EVENT_TYPE: &str = "timealarm";
 
 pub(super) struct TimeAlarms<'r> {
-    time_alarms: Alarms<'r, 'static, 'static>,
+    time_alarms: Alarms<'r, 'static>,
 }
 
 pub(super) struct TimeAlarmsMut<'r> {
@@ -80,14 +80,15 @@ impl<'r> TimeAlarmsMut<'r> {
         max_count: AlarmsCount,
     ) -> ContractResult<(AlarmsCount, MessageResponse)> {
         self.time_alarms
-            .as_alarms()
             .alarms_selection(ctime)
             .take(max_count.try_into()?)
             .collect::<Result<Vec<_>, _>>()?
             .into_iter()
             .try_fold(
                 AlarmsDispatcher::new(ExecuteAlarmMsg::TimeAlarm {}, EVENT_TYPE),
-                |mut dispatcher, (subscriber, _)| -> ContractResult<_> {
+                |mut dispatcher: AlarmsDispatcher<ExecuteAlarmMsg>,
+                 subscriber: Addr|
+                 -> ContractResult<_> {
                     dispatcher = dispatcher.send_to(&subscriber, REPLY_ID)?;
 
                     self.time_alarms.out_for_delivery(subscriber)?;
--- a/contracts/timealarms/src/alarms.rs
+++ b/contracts/timealarms/src/alarms.rs
@@ -4,7 +4,7 @@ use platform::{
     message::Response as MessageResponse,
 };
 use sdk::cosmwasm_std::{Addr, Env, QuerierWrapper, Storage, Timestamp};
-use time_oracle::{Alarms, InDelivery};
+use time_oracle::Alarms;
 
 use crate::{
     msg::{AlarmsCount, AlarmsStatusResponse, ExecuteAlarmMsg},
@@ -30,6 +30,7 @@ impl TimeAlarms {
 
     pub fn remove(&self, storage: &mut dyn Storage, addr: Addr) -> Result<(), ContractError> {
         self.time_alarms.remove(storage, addr)?;
+
         Ok(())
     }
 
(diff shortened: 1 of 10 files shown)


contracts/timealarms/src/alarms.rs
```diff
@@ -90,6 +99,60 @@ impl TimeAlarms {
...
+        self.time_alarms
+            .add(storage, address, Timestamp::from_seconds(time))
+            .map_err(Into::into)
+    }
+}
+
+pub(super) struct InDelivery;
+
+impl InDelivery {
+    const ALARMS: Deque<'_, TimeAlarm> = Deque::new("alarms_in_delivery");
```
gmanev7, line 118: please use a shorter namespace since it is multiplied in the DB a lot and contributes to high gas and high storage size

contracts/timealarms/src/alarms.rs
```diff
@@ -90,6 +99,60 @@ impl TimeAlarms {
...
+        address: Addr,
+        time: TimeSeconds,
+    ) -> ContractResult<()> {
+        self.time_alarms
+            .add(storage, address, Timestamp::from_seconds(time))
+            .map_err(Into::into)
+    }
+}
+
+pub(super) struct InDelivery;
```
gmanev7, line 115: move it in the 'time-oracle' package

contracts/timealarms/src/alarms.rs
```diff
@@ -90,6 +99,60 @@ impl TimeAlarms {
...
+            .and_then(|TimeAlarm { addr, time }| {
+                TimeAlarms::new().try_add_old(storage, addr, time.u64())
+            })
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct TimeAlarm {
+    addr: Addr,
+    time: Uint64,
```
gmanev7, line 155: refrain from using cosmwasm types so deep in our design

contracts/timealarms/src/alarms.rs
gmanev7, line 115: keep `storage` as a member

gmanev7 (commented):
this is just the first set of comments ... once we fx the design issue related to the location of 'InDelivery' we are going to do a second pass

contracts/timealarms/src/alarms.rs
```diff
@@ -66,18 +69,24 @@ impl TimeAlarms {
         self.time_alarms
             .alarms_selection(storage, ctime)
             .take(max_count.try_into()?)
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
```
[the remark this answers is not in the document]
KirilMihaylov: Needed in order to be able to use the `dyn Storage` after loading them, to remove the alarm.
First borrow occurs when loading and second one, which is mutable, occurs during the folding transformation.

contracts/timealarms/src/alarms.rs
KirilMihaylov: Using it only because of it's guarantees for proper (de)serialization and mentioned limitations of JSON engines.

contracts/timealarms/src/alarms.rs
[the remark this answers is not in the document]
gmanev7: I see ... the cause comes from the lack of 'modify-as-we-go' functionality of cosmwasm iterators.

The following will not fix the issue but will simplify the code:
- move the `storage` as a member variable of the `time_oracle::Alarms`
- merge InDelivery into `time_oracle::Alarms` thus avoiding passing the time back and forth. Instead of calling `IndexedMap::remove` use `IndexedMap::may_load`, `IndexedMap::replace` and `Deque::push_back`

contracts/timealarms/src/alarms.rs
```diff
@@ -66,18 +64,24 @@ impl TimeAlarms {
...
-                |dispatcher, alarm| {
-                    dispatcher
-                        .send_to(&alarm?.0, Self::REPLY_ID)
-                        .map_err::<ContractError, _>(Into::into)
+                |mut dispatcher, (addr, time)| -> ContractResult<_> {
+                    dispatcher = dispatcher.send_to(&addr, Self::REPLY_ID)?;
+
+                    self.time_alarms.remove(storage, addr.clone())?;
+
+                    InDelivery::new(storage).add(addr, time)?;
```
gmanev7, line 76: See my previous comment. Combine into one `Alarms::out_for_delivery` call

packages/time-oracle/src/alarms.rs
```diff
@@ -74,6 +78,49 @@ impl<'a> Alarms<'a> {
...
+        Self { storage }
+    }
+
+    pub fn add(&mut self, subscriber: Addr, time: TimeSeconds) -> Result<(), AlarmError> {
+        Self::ALARMS
+            .push_back(self.storage, &TimeAlarm { subscriber, time })
+            .map_err(Into::into)
+    }
+
+    pub fn delivered(&mut self) -> Result<(), AlarmError> {
```
gmanev7, line 98: move as a method `Alarms::last_delivered`

packages/time-oracle/src/alarms.rs
```diff
@@ -74,6 +78,49 @@ impl<'a> Alarms<'a> {
...
+        Self::ALARMS
+            .pop_front(self.storage)
+            .map(|maybe_alarm: Option<TimeAlarm>| debug_assert!(maybe_alarm.is_some()))
+            .map_err(Into::into)
+    }
+
+    pub fn failed(&mut self, alarms: &mut Alarms<'_>) -> Result<(), AlarmError> {
```
gmanev7, line 105: move as a method `Alarms::next_failed`

packages/time-oracle/src/alarms.rs
```diff
@@ -74,6 +78,49 @@ impl<'a> Alarms<'a> {
...
+        Self::ALARMS
+            .pop_front(self.storage)
+            .map_err(Into::into)
+            .and_then(|maybe_alarm: Option<TimeAlarm>| {
+                maybe_alarm.ok_or(AlarmError::ReplyOnEmptyAlarmQueue)
+            })
+            .and_then(|TimeAlarm { subscriber, time }| {
+                alarms.add(self.storage, subscriber, Timestamp::from_seconds(time))
```
gmanev7, line 113: use a private method of `Alarms::add_impl` to avoid the 'seconds - millis - seconds' conversion

packages/time-oracle/src/alarms.rs
```diff
@@ -38,21 +40,43 @@ impl<'a> Alarms<'a> {
...
+        storage: &mut dyn Storage,
+        subscriber: Addr,
+        time: TimeSeconds,
+    ) -> Result<(), AlarmError> {
+        self.alarms.remove(storage, subscriber.clone())?;
+
+        InDelivery::new(storage).add(subscriber, time)
+    }
+
+    #[inline]
```
gmanev7, line 70: [all occurences] remove such compiler directives since they only add noise

gmanev7 (commented):
Hey, most of my comments so far have not been addressed. They are not so clear, or you have concerns?

contracts/timealarms/src/alarms.rs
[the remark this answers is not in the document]
KirilMihaylov: That's not the problem, the real problem is that the `Alarms::alarms_selection` method borrows the storage. The position of the `storage` doesn't matter in this exact case.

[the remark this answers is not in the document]
gmanev7: right, and my comment does not claim that would fix it

packages/time-oracle/src/alarms.rs
```diff
@@ -23,49 +25,64 @@ impl<'a> IndexList<TimeSeconds> for AlarmIndexes<'a> {
...
-    pub fn add(
-        &self,
-        storage: &mut dyn Storage,
-        addr: Addr,
-        time: Timestamp,
-    ) -> Result<(), AlarmError> {
-        self.alarms.save(storage, addr, &as_seconds(time))?;
-        Ok(())
-    }
+type IndexedMap<'namespace> = CwIndexedMap<'namespace, Addr, TimeSeconds, AlarmIndexes<'namespace>>;
```
gmanev7, line 55: I suspect we may safely assume 'namespace == 'static thus simplifying code

contracts/timealarms/src/alarms.rs
```diff
@@ -3,92 +3,108 @@ use platform::{
...
+        &mut self,
         ctime: Timestamp,
         max_count: AlarmsCount,
     ) -> ContractResult<(AlarmsCount, MessageResponse)> {
         self.time_alarms
-            .alarms_selection(storage, ctime)
+            .alarms_selection(ctime)
             .take(max_count.try_into()?)
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
```
gmanev7, line 86: IMHO we have to bail out with an error if we detect that there are alarms left in-delivery

packages/time-oracle/src/alarms.rs
```diff
@@ -11,151 +13,190 @@ fn as_seconds(from: Timestamp) -> TimeSeconds {
...
+        self.in_delivery
+            .pop_front(self.storage)
+            .map(|maybe_alarm: Option<Addr>| debug_assert!(maybe_alarm.is_some()))
+            .map_err(Into::into)
+    }
+
+    pub fn last_failed(&mut self, now: Timestamp) -> Result<(), AlarmError> {
+        self.in_delivery.pop_front(self.storage).map_err(Into::into).and_then(|maybe_alarm: Option<Addr>| {
+            maybe_alarm.ok_or(AlarmError::ReplyOnEmptyAlarmQueue)
+        }).and_then(|subscriber: Addr| self.add_internal(subscriber, as_seconds(now) - /* Minus one second, to ensure it can be run within the same block */ 1))
```
gmanev7, line 120: let's unify the assertions in both cases - now one results in an error, the other - in a debug statement. Since we deal with an external component (wasmd) that may/may not obey the contract then I suggest failing with errors in both cases. Similarly to the way we do before start delivering a batch with alarms.

packages/time-oracle/src/lib.rs
```diff
@@ -11,4 +11,10 @@ pub mod migrate_v1;
 pub enum AlarmError {
     #[error("{0}")]
     Std(#[from] StdError),
+
+    #[error("[Time Oracle] Alarms delivery queue is empty but received reply status")]
+    ReplyOnEmptyAlarmQueue,
+
+    #[error("[Time Oracle] Alarms delivery queue is not empty! Assertion failed")]
+    NonEmptyAlarmQueue,
```
gmanev7, line 19: Suggest defining just plain two errors, EmptyAlarmsInDeliveryQueue and NonEmptyAlarmsInDeliveryQueue. 
The particular cause we shall provide as String, for example: "no alarms in delivery but received a reply"

packages/time-oracle/src/migrate_v1.rs
gmanev7, line 1: Just to note that this file is obsolete and should be removed in a next PR

gmanev7 (approved):

PR #96, opened 2023-05-17
fix(profit): Added storage migration

KirilMihaylov:


contracts/profit/src/msg.rs
```diff
@@ -14,7 +14,11 @@ pub struct InstantiateMsg {
 }
 
 #[derive(Serialize, Deserialize)]
-pub struct MigrateMsg {}
+pub struct MigrateMsg {
+    pub owner: Addr,
```
gmanev7, line 18: seems not used, why do you have it?

KirilMihaylov: Missed it. Thanks.

gmanev7 (approved):

PR #98, opened 2023-05-18
refactor: Refactored `oracle` contract to allow proper rescheduling

### LINKED ISSUE #95: [Price alarms] Reschedule during alarm delivery
Similarly to #81, we need the same mechanism for price alarms.

KirilMihaylov:
Closes #95

### DIFF
--- a/contracts/oracle/src/contract/alarms/iter.rs
+++ b/contracts/oracle/src/contract/alarms/iter.rs
@@ -32,7 +32,7 @@ where
     I: Iterator<Item = PriceResult<BaseC>>,
     BaseC: Currency,
 {
-    pub fn new(alarms: &'alarms PriceAlarms<'storage, S>, price_iter: I) -> Self {
+    pub const fn new(alarms: &'alarms PriceAlarms<'storage, S>, price_iter: I) -> Self {
         Self {
             alarms,
             price_iter,
--- a/contracts/oracle/src/contract/alarms/iter.rs
+++ b/contracts/oracle/src/contract/alarms/iter.rs
@@ -79,11 +79,9 @@ where
 
                 None
             })
-            .or_else(|| {
-                match self.update_alarm_iterator()? {
-                    Ok(iter) => iter.next(),
-                    Err(err) => Some(Err(err))
-                }
+            .or_else(|| match self.update_alarm_iterator()? {
+                Ok(iter) => iter.next(),
+                Err(err) => Some(Err(err)),
             })
     }
 }
--- a/contracts/oracle/src/contract/alarms/iter.rs
+++ b/contracts/oracle/src/contract/alarms/iter.rs
@@ -40,25 +40,24 @@ where
         }
     }
 
-    fn update_alarm_iterator(&mut self) -> Option<ContractResult<&mut AlarmIter<'alarms>>> {
-        self.price_iter.next()?.map_or_else(
-            |error: ContractError| Some(Err(error)),
-            |ref price| {
-                let iter: AlarmIter<'alarms> =
-                    match visit_any_on_ticker::<SwapGroup, Cmd<'storage, 'alarms, '_, S, BaseC>>(
+    fn update_alarm_iterator(&mut self) -> ContractResult<()> {
+        self.alarm_iter = self
+            .price_iter
+            .next()
+            .map(|price_result: PriceResult<BaseC>| {
+                price_result.and_then(|ref price| {
+                    visit_any_on_ticker::<SwapGroup, Cmd<'storage, 'alarms, '_, S, BaseC>>(
                         price.base_ticker(),
                         Cmd {
                             alarms: self.alarms,
                             price,
                         },
-                    ) {
-                        Ok(iter) => iter,
-                        Err(error) => return Some(Err(error)),
-                    };
+                    )
+                })
+            })
+            .transpose()?;
 
-                Some(Ok(self.alarm_iter.insert(iter)))
-            },
-        )
+        Ok(())
     }
 }
 
@@ -71,18 +70,18 @@ where
     type Item = ContractResult<Addr>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.alarm_iter
-            .as_mut()
-            .map(Iterator::next)
-            .unwrap_or_else(|| {
-                self.alarm_iter = None;
+        let result: Option<ContractResult<Addr>> =
+            self.alarm_iter.as_mut().and_then(Iterator::next);
 
-                None
-            })
-            .or_else(|| match self.update_alarm_iterator()? {
-                Ok(iter) => iter.next(),
-                Err(err) => Some(Err(err)),
-            })
+        if result.is_some() {
+            result
+        } else {
+            if let Err(error) = self.update_alarm_iterator() {
+                return Some(Err(error));
+            }
+
+            self.alarm_iter.as_mut().and_then(Iterator::next)
+        }
     }
 }
 
--- a/contracts/oracle/src/contract/alarms/iter.rs
+++ b/contracts/oracle/src/contract/alarms/iter.rs
@@ -75,11 +75,9 @@ where
 
         if result.is_some() {
             result
+        } else if let Err(error) = self.update_alarm_iterator() {
+                Some(Err(error))
         } else {
-            if let Err(error) = self.update_alarm_iterator() {
-                return Some(Err(error));
-            }
-
             self.alarm_iter.as_mut().and_then(Iterator::next)
         }
     }
--- a/contracts/oracle/src/contract/alarms/iter.rs
+++ b/contracts/oracle/src/contract/alarms/iter.rs
@@ -1,144 +1,117 @@
+use std::{iter, ops::Deref};
+
 use serde::{de::DeserializeOwned, Serialize};
 
 use finance::{
-    currency::{self, AnyVisitor, AnyVisitorResult, Currency},
-    price::base::BasePrice,
+    currency::{visit_any_on_ticker, AnyVisitor, AnyVisitorResult, Currency},
+    price::{base::BasePrice, Price},
 };
-use marketprice::alarms::{AlarmsIterator, PriceAlarms};
+use marketprice::alarms::{errors::AlarmError, AlarmsIterator, PriceAlarms};
 use sdk::cosmwasm_std::{Addr, Storage};
 use swap::SwapGroup;
 
-use crate::ContractError;
+use crate::{contract::alarms::PriceResult, error::ContractError, result::ContractResult};
 
-use super::{MarketAlarms, PriceResult};
+type AlarmIterMapFn = fn(Result<Addr, AlarmError>) -> ContractResult<Addr>;
+type AlarmIter<'alarms> = iter::Map<AlarmsIterator<'alarms>, AlarmIterMapFn>;
 
-struct AlarmsCmd<'a, 'b, OracleBase>
+pub struct Iter<'storage, 'alarms, S, I, BaseC>
 where
-    OracleBase: Currency,
+    S: Deref<Target = (dyn Storage + 'storage)>,
+    I: Iterator<Item = PriceResult<BaseC>>,
+    BaseC: Currency,
 {
-    storage: &'a dyn Storage,
-    price_alarms: &'static PriceAlarms<'static>,
-    price: &'b BasePrice<SwapGroup, OracleBase>,
+    alarms: &'alarms PriceAlarms<'storage, S>,
+    price_iter: I,
+    alarm_iter: Option<AlarmIter<'alarms>>,
 }
 
-impl<'a, 'b, OracleBase> AnyVisitor for AlarmsCmd<'a, 'b, OracleBase>
+impl<'storage, 'alarms, S, I, BaseC> Iter<'storage, 'alarms, S, I, BaseC>
 where
-    OracleBase: Currency,
+    S: Deref<Target = (dyn Storage + 'storage)>,
+    I: Iterator<Item = PriceResult<BaseC>>,
+    BaseC: Currency,
 {
+    pub fn new(alarms: &'alarms PriceAlarms<'storage, S>, price_iter: I) -> Self {
+        Self {
+            alarms,
+            price_iter,
+            alarm_iter: None,
+        }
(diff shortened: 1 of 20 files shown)


packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,241 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
-        Ok(self
-            .alarms_above_or_equal()
-            .save(storage, addr.to_owned(), &AlarmStore::new(&alarm))?)
-    }
+    IndexedMap::new(alarms_namespace, indexes)
+}
+
+pub struct PriceAlarms<'storage, S>
+where
+    S: Deref<Target = dyn Storage + 'storage>,
```
gmanev7, line 100: [everywhere] why have you decided to switch from `&'a dyn Storage` to `Deref<Target = dyn Storage + 'a>`

gmanev7: we do not need to work with other types than `&'a dyn Storage`

KirilMihaylov: We do. I did it so I don't have to separate the the implementations for immutable and mutable versions, depending on the storage.

KirilMihaylov: Reasoning is that, `&_` implements `Deref` but `&mut _` implements both `Deref` and `DerefMut`.

KirilMihaylov: That difference can be seen on line +182

gmanev7: indeed we need to use &mut as well

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,241 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
 
-    fn alarms_below(&self) -> IndexedMap<'_, Addr, AlarmStore, AlarmsIndexes<'_>> {
-        let indexes = AlarmsIndexes(MultiIndex::new(
-            |_, price| price.to_owned(),
-            self.alarms_below_namespace,
-            self.index_below_namespace,
-        ));
-        IndexedMap::new(self.alarms_below_namespace, indexes)
+    pub fn ensure_no_in_delivery(&self) -> Result<(), AlarmError> {
+        if self.in_delivery.is_empty(&*self.storage)? {
```
gmanev7, line 142: [everywhere]
```suggestion
        if self.in_delivery.is_empty(self.storage.deref())? {
```

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,241 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+    }
+
+    pub fn last_failed(&mut self) -> Result<(), AlarmError> {
+        self.in_delivery
+            .pop_front(&mut *self.storage)
+            .map_err(Into::into)
+            .and_then(|maybe_alarm: Option<AlarmWithSubscriber>| {
+                maybe_alarm.ok_or(AlarmError::EmptyAlarmsInDeliveryQueue(String::from(
+                    "Received failure reply status",
+                )))
```
gmanev7, line 274: wonder if there would be any gains in using the lazy `ok_or_else` ... if in doubt check that out on https://godbolt.org/

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,241 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+            })
+            .and_then(|alarm: AlarmWithSubscriber| {
+                Self::add_alarm_internal(
+                    &self.alarms_below,
+                    &mut *self.storage,
```
gmanev7, line 279: [everywhere]
```suggestion
                    self.storage.deref_mut(),
```

gmanev7 (commented):
continue ...

packages/marketprice/src/alarms/mod.rs
KirilMihaylov: Usually there is. Clippy used to complain about it, but I suppose it was when in pedantic mode. Will replace it it with `ok_or_else` because it will skip one unnecessary allocation as far as I am aware.

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,241 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+                    } else {
+                        Ok(())
+                    }
+                })
+            })
     }
+
+    fn add_alarm_internal(
+        alarms: &IndexedMap,
+        storage: &mut dyn Storage,
```
gmanev7, line 311: switch their place to reflect their appearance in the PriceAlarms' members list

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,252 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+                })
+            })
+    }
+
+    pub fn last_failed(&mut self) -> Result<(), AlarmError> {
+        self.in_delivery
+            .pop_front(self.storage.deref_mut())
+            .map_err(Into::into)
+            .and_then(|maybe_alarm: Option<AlarmWithSubscriber>| {
+                maybe_alarm.ok_or(AlarmError::EmptyAlarmsInDeliveryQueue(String::from(
```
gmanev7, line 283: use `ok_or_else` instead

packages/marketprice/src/alarms/mod.rs
```diff
@@ -21,23 +25,17 @@ pub mod errors;
...
 
-pub struct PriceAlarms<'m> {
-    alarms_below_namespace: &'m str,
-    alarms_above_namespace: &'m str,
-    index_below_namespace: &'m str,
-    index_above_namespace: &'m str,
-}
-
 #[derive(Clone, Serialize, Deserialize, Debug)]
 struct AlarmStore(CoinDTO<SwapGroup>);
```
gmanev7, line 29: let's take this opportunity to change the name to a more meaningful one, for example, `NormalizedPrice`

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,252 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+    {
+        Self::add_alarm_internal(
+            self.storage.deref_mut(),
+            &self.alarms_above_or_equal,
+            subscriber,
+            &AlarmStore::new(&alarm),
+        )
+    }
+
+    pub fn remove(&mut self, addr: Addr) -> Result<(), AlarmError> {
```
gmanev7, line 218: rename the arg `addr`

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,252 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
-            .sub_prefix(C::TICKER.into())
-            .range(
-                storage,
-                Some(Bound::exclusive((price.0.amount(), Addr::unchecked("")))),
+        Self::add_alarm_internal(
+            self.storage.deref_mut(),
+            &self.alarms_below,
+            subscriber,
+            &AlarmStore::new(&alarm),
+        )
```
gmanev7:
```suggestion
self.add_alarm_below_internal(subscriber, &AlarmStore::new(&alarm))
```

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,252 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+                            alarm.subscriber.clone(),
+                            &above,
+                        )
+                    } else {
+                        Ok(())
+                    }
+                })
+            })
     }
+
```
gmanev7, line 310:
```suggestion
    fn add_alarm_below_internal(
        &mut self,
        subscriber: Addr,
        alarm: &AlarmStore,
    ) -> Result<(), AlarmError> {
        Self::add_alarm_internal(
            self.storage.deref_mut(),
            &self.alarms_below,
            subscriber,
            alarm,
        )
    }

```

packages/marketprice/src/alarms/mod.rs
```diff
@@ -77,136 +75,252 @@ impl<'a> Prefixer<'a> for AlarmStore {
...
+                    "Received failure reply status",
+                )))
+            })
+            .and_then(|alarm: AlarmWithSubscriber| {
+                Self::add_alarm_internal(
+                    self.storage.deref_mut(),
+                    &self.alarms_below,
+                    alarm.subscriber.clone(),
+                    &alarm.below,
+                )
```
gmanev7, line 295:
```suggestion
               self.add_alarm_below_internal(alarm.subscriber.clone(), &alarm.below)
```

gmanev7 (commented):
this set of changes removes an instance of code duplication. Please apply it for the `add_alarm_above`

tests/src/oracle_tests.rs
```diff
@@ -434,3 +450,356 @@ fn test_zero_price_dto() {
...
+
+static EXECUTE_FAIL_SUCCESS_FLAG: AtomicBool = AtomicBool::new(false);
+
+fn reset_execute_fail_success() {
+    EXECUTE_FAIL_SUCCESS_FLAG.store(false, SeqCst)
+}
+
+fn load_and_update_execute_fail_success() -> bool {
+    EXECUTE_FAIL_SUCCESS_FLAG.fetch_xor(true, SeqCst)
+}
```
gmanev7, line 507: could you use another solution that does not involve global variables?

contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -194,23 +198,25 @@ where
         ))
     }
 
-    pub fn swap_pairs_df(&self) -> impl Iterator<Item = SwapLeg> + '_ {
```
gmanev7, line 197: let's not bind the API to concrete types, why do we switch out from `impl Xxx` ???

contracts/oracle/src/contract/mod.rs
```diff
@@ -118,7 +120,7 @@ pub fn sudo(deps: DepsMut<'_>, _env: Env, msg: SudoMsg) -> ContractResult<CwResp
         SudoMsg::UpdateConfig(price_config) => Config::update(deps.storage, price_config),
         SudoMsg::RegisterFeeder { feeder_address } => Feeders::try_register(deps, feeder_address),
         SudoMsg::RemoveFeeder { feeder_address } => Feeders::try_remove(deps, feeder_address),
-        SudoMsg::RemovePriceAlarm { receiver } => MarketAlarms::remove(deps.storage, receiver),
+        SudoMsg::RemovePriceAlarm { receiver } => MarketAlarms::new(deps.storage).remove(receiver),
```
gmanev7, line 123: this API is not needed anymore please remove

packages/marketprice/src/alarms/mod.rs
```diff
@@ -335,74 +450,185 @@ pub mod tests {
...
+            alarms.ensure_no_in_delivery().unwrap_err(),
+            AlarmError::NonEmptyAlarmsInDeliveryQueue(_)
+        ));
+
+        alarms.last_failed().unwrap();
+
+        assert!(matches!(
+            alarms.ensure_no_in_delivery().unwrap_err(),
+            AlarmError::NonEmptyAlarmsInDeliveryQueue(_)
+        ));
```
gmanev7, line 615: am I mistaken or do we have a bug at this point? As far as I understand we should have nothing in delivery past `last_failed` - because it would pop the last item in delivery

KirilMihaylov: Good catch! Thanks.

KirilMihaylov: Good news is that CI caught it too.

contracts/oracle/src/contract/alarms/mod.rs
```diff
@@ -10,101 +12,142 @@ use marketprice::{alarms::PriceAlarms, SpotPrice};
...
+                NAMESPACE_ALARMS_BELOW,
+                NAMESPACE_INDEX_BELOW,
+                NAMESPACE_ALARMS_ABOVE,
+                NAMESPACE_INDEX_ABOVE,
+                NAMESPACE_IN_DELIVERY,
+            ),
+        }
+    }
+
+    pub fn notify_alarms_iter<I, BaseC>(&self, prices: I) -> AlarmsIter<'storage, '_, S, I, BaseC>
```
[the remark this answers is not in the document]
KirilMihaylov: I am not able to do so unless I add a bound which states that the iterator type `I` should outlive the lifetime `'storage`.
That is because under the hood, I capture both lifetimes which are on their own independent.
Another thing is, I don't think that it's necessarily bad to state the type as long as it doesn't imply what the implementation is, unlike the `iter::Map<I, ...>` types that I used in previous iterations.

contracts/oracle/src/contract/mod.rs
```diff
@@ -131,16 +132,19 @@ pub fn reply(deps: DepsMut<'_>, _env: Env, msg: Reply) -> ContractResult<CwRespo
...
+
+    match msg.result {
+        SubMsgResult::Ok(_) => alarms
+            .last_delivered()
+            .map(|()| emitter.emit(KEY_DELIVERED, "success")),
+        SubMsgResult::Err(error) => alarms.last_failed().map(|()| {
+            emitter
+                .emit(KEY_DELIVERED, "error")
+                .emit(KEY_DETAILS, error)
+        }),
```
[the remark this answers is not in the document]
KirilMihaylov: Yes, we can! That's part of the good news of this PR.

contracts/oracle/src/contract/alarms/iter.rs
```diff
@@ -71,18 +70,18 @@ where
...
-            })
-            .or_else(|| match self.update_alarm_iterator()? {
-                Ok(iter) => iter.next(),
-                Err(err) => Some(Err(err)),
-            })
+        if result.is_some() {
+            result
+        } else {
+            if let Err(error) = self.update_alarm_iterator() {
+                return Some(Err(error));
```
gmanev7, line 80: no need of a `return`

gmanev7 (approved):
good job!

PR #99, opened 2023-05-18
feat(leaser): migrate leases in batches

gmanev7:


contracts/leaser/src/migrate.rs
```diff
@@ -2,66 +2,97 @@ use lease::api::MigrateMsg;
...
+                        if self.leases_left == NbInstances::MIN {
+                            res.last_instance = Some(lease);
+                        }
+                        res
+                    })
             })
         };
 
-        self.may_batch = self.may_batch.and_then(op);
+        self.may_result = self.may_result.and_then(op);
```
KirilMihaylov: What is the point in this? If one iteration fails, the next ones will continue to charge gas while the whole transaction won't even be committed. Either the whole part above should be wrapped (which will just lower the cost, but won't solve the problem) of return `ContractResult<()>` and stop iteration of first encountered error.

contracts/leaser/src/migrate.rs
```diff
@@ -2,66 +2,97 @@ use lease::api::MigrateMsg;
...
+        F: FnOnce(&mut Batch) -> ContractResult<()>,
+    {
+        add_fn(&mut self.msgs).map(|()| self)
+    }
+}
 struct MigrateBatch {
     new_code_id: u64,
-    may_batch: ContractResult<Batch>,
+    leases_left: NbInstances,
+    may_result: ContractResult<MigrationResult>,
```
KirilMihaylov: I think this shouldn't be a result type in the first place. Those are better suited for function/method result types, but are rather meaningless to be stored.

contracts/leaser/src/migrate.rs
```diff
@@ -2,66 +2,97 @@ use lease::api::MigrateMsg;
...
+) -> ContractResult<MigrationResult>
 where
     I: Iterator<Item = ContractResult<Addr>>,
 {
-    let no_msgs = MigrateBatch::new(lease_code_id);
-    let migrated_msgs = leases.fold(no_msgs, MigrateBatch::migrate_lease);
+    let no_msgs = MigrateBatch::new(lease_code_id, max_leases);
+    let migrated_msgs = leases
+        .take(max_leases.try_into()?)
+        .fold(no_msgs, MigrateBatch::migrate_lease);
```
KirilMihaylov: `Iterator::try_fold` is stable, `Iterator::try_reduce` isn't, in case that was the reason to avoid it.

contracts/leaser/src/migrate.rs
```diff
@@ -2,66 +2,97 @@ use lease::api::MigrateMsg;
...
+impl TryFrom<MigrateBatch> for MigrationResult {
     type Error = ContractError;
     fn try_from(this: MigrateBatch) -> Result<Self, Self::Error> {
-        this.may_batch
+        this.may_result
     }
 }
 
 #[cfg(test)]
 mod test {
```
KirilMihaylov, line 109: No tests for `max_leases = 0`. Either such should be added or `NonZero*` variant should be used.

contracts/leaser/src/migrate.rs
```diff
@@ -88,15 +119,66 @@ mod test {
...
+        }
+
+        {
+            let exp = add_expected(MigrationResult::default(), &addr6, new_code);
+            let mut exp = add_expected(exp, &addr7, new_code);
+            exp.last_instance = None;
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(leases.into_iter().skip(2 + 3).map(Ok), new_code, 15)
+            );
```
KirilMihaylov: Expected expressions are second arguments to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(leases.into_iter().skip(2 + 3).map(Ok), new_code, 15),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -88,15 +119,66 @@ mod test {
...
+
+        {
+            let exp = add_expected(MigrationResult::default(), &addr3, new_code);
+            let exp = add_expected(exp, &addr4, new_code);
+            let mut exp = add_expected(exp, &addr5, new_code);
+            exp.last_instance = Some(addr5);
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(leases.clone().into_iter().skip(2).map(Ok), new_code, 3)
+            );
```
KirilMihaylov: Expected expressions are second arguments to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(leases.clone().into_iter().skip(2).map(Ok), new_code, 3),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -88,15 +119,66 @@ mod test {
...
+            .into();
+
+        {
+            let exp = add_expected(MigrationResult::default(), &addr1, new_code);
+            let mut exp = add_expected(exp, &addr2, new_code);
+            exp.last_instance = Some(addr2);
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(leases.clone().into_iter().map(Ok), new_code, 2)
+            );
```
KirilMihaylov: Expected expressions are second arguments to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(leases.clone().into_iter().map(Ok), new_code, 2),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -88,15 +119,66 @@ mod test {
...
-            .unwrap();
+        let exp = add_expected(MigrationResult::default(), &addr1, new_code);
+        let mut exp = add_expected(exp, &addr2, new_code);
+        exp.last_instance = Some(addr2);
 
         assert_eq!(
             Ok(exp),
-            super::migrate_leases(leases.into_iter().map(Ok), new_code)
+            super::migrate_leases(leases.into_iter().map(Ok), new_code, 2)
         );
```
KirilMihaylov: Expected expressions are second arguments to the macro.
```suggestion
        assert_eq!(
            super::migrate_leases(leases.into_iter().map(Ok), new_code, 2)
            Ok(exp)
        );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,66 +2,97 @@ use lease::api::MigrateMsg;
...
     #[test]
     fn no_leases() {
         let new_code = 242;
         let no_leases = vec![];
         assert_eq!(
-            Ok(Batch::default()),
-            super::migrate_leases(no_leases.into_iter().map(Ok), new_code)
+            Ok(MigrationResult::default()),
+            super::migrate_leases(no_leases.into_iter().map(Ok), 2, new_code)
         );
```
KirilMihaylov: Expected expressions are second arguments to the macro.
```suggestion
        assert_eq!(
            super::migrate_leases(no_leases.into_iter().map(Ok), 2, new_code),
            Ok(MigrationResult::default())
        );
```

contracts/leaser/src/migrate.rs
```diff
@@ -77,7 +108,7 @@ mod test {
         ];
         assert_eq!(
             Err(ContractError::ParseError { err: err.into() }),
-            super::migrate_leases(no_leases.into_iter(), new_code)
+            super::migrate_leases(no_leases.into_iter(), 5, new_code)
         );
```
KirilMihaylov: Expected expressions are second arguments to the macro.
```suggestion
        assert_eq!(
            super::migrate_leases(no_leases.into_iter(), 5, new_code),
            Err(ContractError::ParseError { err: err.into() })
        );
```

packages/platform/src/batch.rs
```diff
@@ -16,7 +16,7 @@ pub type ReplyId = u64;
 #[derive(Default)]
 #[cfg_attr(
     any(debug_assertions, test, feature = "testing"),
-    derive(Debug, PartialEq)
+    derive(Debug, PartialEq, Eq)
 )]
```
KirilMihaylov, line 20: I'd suggest dropping the conditional compilation for those. Either way they will be removed either by the compiler or linker.
```suggestion
#[derive(Debug, PartialEq, Eq, Default)]
```

contracts/leaser/src/state/leases.rs
```diff
@@ -74,40 +74,18 @@ impl Leases {
...
-                Self::Err(e) => e.take().map(Result::Err),
-            }
-        }
-    }
-
-    match res {
-        Ok(r) => ResultIter::Ok(r.into_iter()),
-        Err(e) => ResultIter::Err(Some(e)),
+            .range(storage, start_bound, None, cosmwasm_std::Order::Ascending)
+            .map(|e| {
```
KirilMihaylov, line 85: Please, rename the parameter to something more meaningful.

contracts/leaser/src/state/leases.rs
```diff
@@ -74,40 +74,18 @@ impl Leases {
         })
     }
 
-    pub fn iter(storage: &dyn Storage) -> impl Iterator<Item = ContractResult<Addr>> + '_ {
+    pub fn iter(
+        storage: &dyn Storage,
+        next_customer: Option<Addr>,
+    ) -> impl Iterator<Item = ContractResult<Customer<IntoIter<Addr>>>> + '_ {
```
KirilMihaylov, line 80: (Note) Please, refrain from using `impl Trait` in return position unless absolutely necessary until it is stabilized in trait methods too. It becomes a pain to refactor later.

KirilMihaylov, line 80: Maybe consider defining a type alias for the `Iterator::Item`.

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        let lease42 = Addr::unchecked(LEASE42);
+        let lease43 = Addr::unchecked(LEASE43);
+        let customer_addr1 = Addr::unchecked(CUSTOMER_ADDR1);
+        let customer_addr2 = Addr::unchecked(CUSTOMER_ADDR2);
+        let customer_addr3 = Addr::unchecked(CUSTOMER_ADDR3);
+        let customer_addr4 = Addr::unchecked(CUSTOMER_ADDR4);
+        let cust1 = Customer::from(customer_addr1, vec![lease1].into_iter());
+        let cust2 = Customer::from(customer_addr2, vec![lease21, lease22].into_iter());
+        let cust3 = Customer::from(customer_addr3, vec![lease3].into_iter());
+        let cust4 = Customer::from(customer_addr4, vec![lease41, lease42, lease43].into_iter());
```
KirilMihaylov, line 302: I'd suggest at least grouping those for better readability.
```suggestion
        let lease1 = Addr::unchecked(LEASE1);
        let customer_addr1 = Addr::unchecked(CUSTOMER_ADDR1);
        let cust1 = Customer::from(customer_addr1, vec![lease1].into_iter());

        let lease21 = Addr::unchecked(LEASE21);
        let lease22 = Addr::unchecked(LEASE22);
        let customer_addr2 = Addr::unchecked(CUSTOMER_ADDR2);
        let cust2 = Customer::from(customer_addr2, vec![lease21, lease22].into_iter());

        let lease3 = Addr::unchecked(LEASE3);
        let customer_addr3 = Addr::unchecked(CUSTOMER_ADDR3);
        let cust3 = Customer::from(customer_addr3, vec![lease3].into_iter());

        let lease41 = Addr::unchecked(LEASE41);
        let lease42 = Addr::unchecked(LEASE42);
        let lease43 = Addr::unchecked(LEASE43);
        let customer_addr4 = Addr::unchecked(CUSTOMER_ADDR4);
        let cust4 = Customer::from(customer_addr4, vec![lease41, lease42, lease43].into_iter());
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+use crate::{msg::NbInstances, result::ContractResult};
 
-pub fn migrate_leases<I>(leases: I, lease_code_id: u64) -> ContractResult<Batch>
+pub struct Customer<LeaseIter> {
+    customer: Addr,
+    leases: LeaseIter,
+}
+
+#[derive(Default)]
+#[cfg_attr(test, derive(Debug, Eq, PartialEq))]
```
KirilMihaylov, line 13: I'd suggest dropping the conditional compilation for those. Either way they will be removed either by the compiler or linker.
```suggestion
#[derive(Debug, Eq, PartialEq, Default)]
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+/// Consumes the customers iterator to the next customer or error.
+pub fn migrate_leases<I, LI>(
+    mut customers: I,
+    lease_code_id: u64,
+    max_leases: NbInstances,
+) -> ContractResult<MigrationResult>
 where
-    I: Iterator<Item = ContractResult<Addr>>,
+    I: Iterator<Item = ContractResult<Customer<LI>>>,
+    LI: ExactSizeIterator<Item = Addr>,
```
KirilMihaylov, line 29: What exactly is the point with this nested iteration?

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
 {
-    let no_msgs = MigrateBatch::new(lease_code_id);
-    let migrated_msgs = leases.fold(no_msgs, MigrateBatch::migrate_lease);
-    migrated_msgs.try_into()
+    let mut msgs = MigrateBatch::new(lease_code_id, max_leases);
+    customers
+        .find_map(|may_customer| match may_customer {
```
KirilMihaylov, line 33: (Typo) Sadly, May the 4th has passed us already. ;)
```suggestion
        .find_map(|maybe_customer| match maybe_customer {
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+    /// None if there is enough room for all customer's leases, otherwise return the customer
+    fn migrate_or_be_next<LI>(&mut self, mut customer: Customer<LI>) -> Option<ContractResult<Addr>>
+    where
+        LI: ExactSizeIterator<Item = Addr>,
+    {
+        let may_leases_nb: Result<NbInstances, _> = customer.leases.len().try_into();
+        match may_leases_nb {
+            Err(err) => Some(Err(err.into())),
+            Ok(leases_nb) => {
+                if leases_nb <= self.leases_left {
```
KirilMihaylov, line 85: What happens if that is never the case and the gas would otherwise not be enough/hits a limit? In my opinion previous commit's way was better in that it did deal with this case too.

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+                    self.leases_left -= leases_nb;
```
KirilMihaylov, line 86: You could easily use `checked_sub`. It will do both the checking and subtraction, thus saving us (even small) some amount of gas. ;)

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        let customers = vec![Ok(cust1)];
+        {
+            let exp = MigrationResult {
+                next_customer: Some(customer_addr1),
+                ..Default::default()
+            };
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(customers.into_iter(), new_code, 2)
+            );
```
KirilMihaylov, line 165: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(customers.into_iter(), new_code, 2),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+
+        {
+            let exp = MigrationResult {
+                next_customer: Some(customer_addr1),
+                ..Default::default()
+            };
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 0)
+            );
```
KirilMihaylov, line 192: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 0),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        }
+        {
+            let mut exp = add_expected(MigrationResult::default(), &lease1, new_code);
+            exp.next_customer = Some(customer_addr2.clone());
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 1)
+            );
```
KirilMihaylov, line 200: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 1),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        }
+        {
+            let mut exp = add_expected(MigrationResult::default(), &lease1, new_code);
+            exp.next_customer = Some(customer_addr2);
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 2)
+            );
```
KirilMihaylov, line 208: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 2),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        }
+        {
+            let exp = add_expected(MigrationResult::default(), &lease1, new_code);
+            let exp = add_expected(exp, &lease21, new_code);
+            let mut exp = add_expected(exp, &lease22, new_code);
+            exp.next_customer = Some(customer_addr3);
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 3)
+            );
```
KirilMihaylov, line 218: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 3),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        {
+            let exp = add_expected(MigrationResult::default(), &lease1, new_code);
+            let exp = add_expected(exp, &lease21, new_code);
+            let exp = add_expected(exp, &lease22, new_code);
+            let mut exp = add_expected(exp, &lease3, new_code);
+            exp.next_customer = Some(customer_addr4.clone());
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 4)
+            );
```
KirilMihaylov, line 229: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 4),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        {
+            let exp = add_expected(MigrationResult::default(), &lease1, new_code);
+            let exp = add_expected(exp, &lease21, new_code);
+            let exp = add_expected(exp, &lease22, new_code);
+            let mut exp = add_expected(exp, &lease3, new_code);
+            exp.next_customer = Some(customer_addr4);
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 5)
+            );
```
KirilMihaylov, line 240: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 5),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+            let exp = add_expected(exp, &lease22, new_code);
+            let exp = add_expected(exp, &lease3, new_code);
+            let exp = add_expected(exp, &lease41, new_code);
+            let exp = add_expected(exp, &lease42, new_code);
+            let mut exp = add_expected(exp, &lease43, new_code);
+            exp.next_customer = None;
+            assert_eq!(
+                Ok(exp),
+                super::migrate_leases(test_customers(), new_code, 7)
+            );
```
KirilMihaylov, line 254: Expected expression is supposed to be the second argument to the macro.
```suggestion
            assert_eq!(
                super::migrate_leases(test_customers(), new_code, 7),
                Ok(exp)
            );
```

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,305 @@ use lease::api::MigrateMsg;
...
+        let customers = vec![
+            Ok(cust1),
             Err(ContractError::ParseError { err: err.into() }),
-            Ok(Addr::unchecked("2424")),
         ];
         assert_eq!(
             Err(ContractError::ParseError { err: err.into() }),
-            super::migrate_leases(no_leases.into_iter(), new_code)
+            super::migrate_leases(customers.into_iter(), new_code, 3)
         );
```
KirilMihaylov, line 277: Expected expression is supposed to be the second argument to the macro.
```suggestion
        assert_eq!(
            super::migrate_leases(customers.into_iter(), new_code, 3),
            Err(ContractError::ParseError { err: err.into() })
        );
```

contracts/leaser/src/msg.rs
```diff
@@ -25,6 +25,8 @@ pub struct InstantiateMsg {
 #[derive(Serialize, Deserialize)]
 pub struct MigrateMsg {}
 
+pub type NbInstances = u32;
```
KirilMihaylov, line 28: Perhaps a better describing name should be considered.

KirilMihaylov (changes requested):
May the 4th was not with us today. 😞

contracts/leaser/src/migrate.rs
```diff
@@ -2,66 +2,97 @@ use lease::api::MigrateMsg;
 use platform::batch::Batch;
 use sdk::cosmwasm_std::Addr;
 
-use crate::{error::ContractError, result::ContractResult};
+use crate::{error::ContractError, msg::NbInstances, result::ContractResult};
```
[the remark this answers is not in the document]
gmanev7: any suggestions?

contracts/leaser/src/migrate.rs
gmanev7: the macro definition does not stipulate that

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/state/leases.rs
gmanev7: what's the relation to the unstable impl in Trait methods? Using concrete types has always been a bigger pain.

contracts/leaser/src/migrate.rs
gmanev7: same

gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: added a test scenario at `paging()`

contracts/leaser/src/migrate.rs
gmanev7: the point here is that they slow down the compilation ... and better describe what is allowed for the types and from there

gmanev7: and that is the policy throughout the whole project ... let's not change it just for a type. If we decide later to do it we are going to prepare a dedicated commit/PR and compare the compilation time and binary size

contracts/leaser/src/migrate.rs
gmanev7: supposing  you call "nested" the list of leases per customer, the idea is to obtain the next customer addr

contracts/leaser/src/migrate.rs
gmanev7: the previous commit in this PR was buggy - it mixed customer with lease addresses

contracts/leaser/src/migrate.rs
gmanev7: disagree - it will cost us more since will check again what we have checked already. If you mean replacing the two lines then the cost of transforming Err to Option<Result<>> would be higher...not counting the worsen code readability

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/migrate.rs
gmanev7: same

contracts/leaser/src/state/leases.rs
KirilMihaylov: The relation is the refactoring of the `timealarms` contract that I did a few days ago, because I needed to extract one method into a trait and thus opened Pandora's box. I had to go through several layers of return types to be able to deal with it.

contracts/leaser/src/migrate.rs
KirilMihaylov: The mistake is mine. It appears that I was left to believe in a lie due to my IDE.

contracts/leaser/src/migrate.rs
```diff
@@ -2,101 +2,310 @@ use lease::api::MigrateMsg;
...
+}
+
+#[derive(Default)]
+#[cfg_attr(test, derive(Debug, Eq, PartialEq))]
+pub struct MigrationResult {
+    pub msgs: Batch,
+    pub next_customer: Option<Addr>,
+}
+
+pub type MayCustomer<LI> = ContractResult<Customer<LI>>;
```
KirilMihaylov, line 19: I'd suggest naming it with `Maybe` as `May` _may_ imply something else.
```suggestion
pub type MaybeCustomer<LI> = ContractResult<Customer<LI>>;
```

contracts/leaser/src/migrate.rs
KirilMihaylov: My point was that because of how we compile it, it will check again for underflow.
My general idea was to use `if let Some(result) = self.leases_left.checked_sub(leases_nb) { ... } else { Some(Ok(customer.customer)) }`. Just to simplify the binary that will be generated.

KirilMihaylov (approved):

PR #102, opened 2023-05-23
refactor(timealarms): Refactored `timealarms` contract to avoid dual definition depending on mutability

KirilMihaylov:


packages/time-oracle/src/migrate_v1.rs
gmanev7, line 1: it is an obsolete file, please remove

gmanev7 (approved):
LGTM

PR #108, opened 2023-05-26
refactor(oracle): Changed supported pairs query

KirilMihaylov:


gmanev7:
I would not recommend doing such a change since we plan to move to a radically new way of checking price alarms.

PR #109, opened 2023-06-01
refactor: Moved liquidation items into finance

maneva3:


contracts/lease/src/lease/liquidation.rs
```diff
@@ -1,10 +1,9 @@
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 
 use finance::{
     coin::Coin,
     currency::Currency,
-    liability::{Liability, Zone},
-    percent::Percent,
+    liability::{check_liability, Status},
```
gmanev7, line 6: the calling convention for free functions is to refer to them through the module they are defined in

packages/finance/src/liability/mod.rs
```diff
@@ -15,9 +15,14 @@ use crate::{
 };
 
 pub use self::level::Level;
+pub use self::liquidation::check_liability;
```
gmanev7, line 18: suggest introducing this function as `check` only to have `liquidation::check(...)` at the calling sites

packages/finance/src/liability/liquidation.rs
```diff
@@ -0,0 +1,615 @@
...
+        lease: &Lease<Lpn, Asset, Lpp, Profit, TimeAlarms, Oracle>,
+    ) -> Coin<Asset> {
+        match self {
+            Self::Partial { amount, cause: _ } => *amount,
+            Self::Full(_) => lease.amount,
+        }
+    }
+}
+
+pub fn inspect_liability<Asset>(
```
gmanev7, line 75: let's keep its previous name `check_liability`

gmanev7 (approved):

PR #111, opened 2023-06-05
Min liquidation amount

maneva3:


packages/finance/src/liability/liquidation.rs
```diff
@@ -85,85 +85,93 @@ where
...
-        },
-        spec.amount_to_liquidate(asset, total_due),
-        min_asset,
-    )
+    let ltv = ltv(overdue, asset);
+    if overdue < min_liquidation {
+        no_liquidation(spec, overdue, ltv)
+    } else {
+        ask_liquidation(asset, Cause::Overdue(), overdue, min_asset)
+    }
```
gmanev7, line 139: this logic is similar to the one used on liquidation by liability ... think on reusing it

packages/finance/src/liability/liquidation.rs
```diff
@@ -117,20 +114,26 @@ fn may_ask_liquidation_liability<Asset>(
...
-            ltv: spec.max(),
-            healthy_ltv: spec.healthy_percent(),
-        },
-        spec.amount_to_liquidate(asset, total_due),
-        min_asset,
-    )
+    let liquidation_amount = spec.amount_to_liquidate(asset, total_due);
+    if liquidation_amount < min_liquidation {
+        None
+    } else {
```
gmanev7, line 126: take this into `may_ask_liquidation` to avoid copy/paste

packages/finance/src/liability/liquidation.rs
```diff
@@ -559,17 +585,68 @@ mod tests {
...
+        let max_ltv = Percent::from_permille(751);
+        let spec = liability_with_max(max_ltv);
+
+        assert_eq!(
+            check_liability::<Nls>(
+                &spec,
+                878.into(),
+                752.into(),
+                0.into(),
+                800.into(),
```
gmanev7, line 599: [all occurrences within this test] select a corner value, not like here `800` which is far from the threshold. The idea is to prove that the result No is returned at that corner value while a value Liquidate is returned at the value + 1

packages/finance/src/liability/liquidation.rs
```diff
@@ -83,30 +83,27 @@ pub fn check_liability<Asset>(
...
-) -> Status<Asset>
+fn no_liquidation<Asset>(spec: &Liability, total_due: Coin<Asset>, ltv: Percent) -> Status<Asset>
 where
     Asset: Currency,
 {
     if total_due.is_zero() {
         Status::NoDebt
     } else {
-        let ltv = Percent::from_ratio(total_due, asset);
         debug_assert!(ltv < spec.max());
```
gmanev7, line 107: please move it at the beginning of the method because it checks a precondition

PR #114, opened 2023-06-09
refactor(integration_tests): Refactored to put a layer of abstraction

KirilMihaylov:


tests/src/common/oracle_wrapper.rs
```diff
@@ -106,10 +106,12 @@ pub fn add_feeder<Lpn>(test_case: &mut TestCase<Lpn>, addr: impl Into<String>)
 where
     Lpn: Currency,
 {
+    let oracle = test_case.oracle().clone();
+
     let response: AppResponse = test_case
         .app
         .wasm_sudo(
-            test_case.oracle.clone().unwrap(),
+            oracle,
```
gmanev7, line 114: [all occurences] please check if it's not more appropriate to take the oracle from the test_case by value

tests/src/common/test_case.rs
```diff
@@ -414,12 +470,26 @@ where
...
-    pub fn leaser(&self) -> Addr {
-        self.leaser_addr.clone().unwrap()
+impl<Lpn> DerefMut for TestCase<Lpn>
+where
+    Lpn: Currency,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.test_case
     }
 }
```
gmanev7, line 495: please refrain from misusing these two traits ... we have already spoken about that. It's preferable to use explicit indirections when doing calls.

tests/src/common/test_case.rs
```diff
@@ -98,33 +91,87 @@ where
...
+    }
+
+    pub fn with_reserve(reserve: &[CwCoin]) -> Self {
+        Self {
+            test_case: GenericTestCase::with_reserve(reserve),
+            _lpn: PhantomData,
+        }
+    }
+
+    pub fn init(&mut self, user: Addr, init_funds: &mut [CwCoin]) -> &mut Self {
```
gmanev7, line 163: please merge this fn into the `new` and even `with_reserve`. The goal is to have the fn that creates the instance initialized with the minimum set of data.

gmanev7 (commented):
support the idea behind this PR!

gmanev7 (approved):
I liked the idea behind "compile-time optional types" implemented by AddrBook

PR #115, opened 2023-06-09
test: Remove `#[ignore]` attribute on tests

KirilMihaylov:


tests/src/common/lease.rs
```diff
@@ -245,6 +252,24 @@ fn do_swap<DownpaymentC, LeaseC>(
...
+    app.send_tokens(
+        Addr::unchecked(ADMIN),
+        Addr::unchecked(ica_addr),
+        &[cwcoin(exp_swap_out)],
+    )
+    .unwrap();
+
     send_swap_response::<DownpaymentC, LeaseC>(app, lease_addr, exp_swap_out, remote_tx_count);
 
     check_state_opened(app, lease_addr);
```
gmanev7, line 275: rename the too-generic naming of `fn send_response_and_expect` to `fn respond_and_expect_ibc_transfer`

tests/src/common/lease.rs
```diff
@@ -215,17 +215,24 @@ pub(crate) fn complete_initialization<Lpn, DownpaymentC, LeaseC>(
...
+        app,
+        lease_addr,
+        ica_addr,
+        connection_id,
+        (downpayment, exp_borrow, exp_lease),
+    );
 }
 
-fn do_swap<DownpaymentC, LeaseC>(
+fn do_swap<DownpaymentC, Lpn, LeaseC>(
```
gmanev7, line 227: why do we organize the requests and responses from the perspective of the contract? That turns into mixing responses for a previous step with requests for the next step. For example, the blank response from DEX should be at the end of the previous step, the transfer-out, not the first step of the swap.

tests/src/common/lease.rs
```diff
@@ -245,6 +252,24 @@ fn do_swap<DownpaymentC, LeaseC>(
         exp_lease
     };
 
+    app.send_tokens(
+        Addr::unchecked(ica_addr),
+        Addr::unchecked(ADMIN),
+        &if currency::equal::<DownpaymentC, LeaseC>() {
```
gmanev7, line 258: there are three times you call `!currency::equal::<DownpaymentC, LeaseC>` in this function. What about doing that once before calling it, computing the values at one place beforehand?

tests/src/common/lease.rs
gmanev7: if you address a previous comment about requests and responses then the naming would be just `fn expect_ibc_transfer`. It would expect an IBC transfer request and send a response.

tests/src/lease_tests.rs
```diff
@@ -189,41 +196,226 @@ fn get_lease_address<Dispatcher, Treasury, Profit, Lpp, Oracle, TimeAlarms>(
...
+            destination_channel: None,
+            data: None,
+            timeout_height: None,
+            timeout_timestamp: None,
+        },
+        data,
+    }
+}
+
+mod repay_mod {
```
gmanev7, line 215: consider splitting this file into `repay.rs`, `close.rs`, `liquidate.rs`, etc .... files of `lease` directory

Cargo.toml
```diff
@@ -54,6 +54,7 @@ cosmos-sdk-proto = { version = "0.17", default-features = false }
 
 # Other
 neutron-sdk = { git = "https://github.com/neutron-org/neutron-sdk", rev = "d0f73c0a377759d0d3fed0bd6d30a17b6b3b110e" }
+osmosis-std = { version = "0.12", default-features = false }
```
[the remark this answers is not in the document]
KirilMihaylov: Will check it out, possibly would open a new PR.

tests/src/common/lease.rs
```diff
@@ -341,79 +341,103 @@ fn send_open_ica_response<'r>(
...
-        .map(swap_trx::build_exact_amount_in_resp);
-    trx::encode_msg_responses(msgs).into()
+    let downpayment_equals_lease_as_amount: Amount = (!downpayment_equals_lease).into();
+
+    let amounts_out = [
+        Amount::from(swap_out) - downpayment_equals_lease_as_amount,
+        downpayment_equals_lease_as_amount,
+    ]
+    .into_iter()
+    .filter(|&amount: &Amount| amount != 0);
```
gmanev7, line 416: too cryptic ... why do we mix bool with Amount?

tests/src/lease_tests.rs
gmanev7: no comment, no action, and resolved state???

KirilMihaylov: My bad! I was planning to but it seems I missed it.

tests/src/common/lease.rs
KirilMihaylov: My intent was to avoid adding more logic depending on whether down-payment and lease are equal. I followed previous code's approach and wanted to skip any steps that would complicate it.

PR #117, opened 2023-06-26
refactor: Remove generic errors

KirilMihaylov:


contracts/lease/src/contract/endpoins.rs
```diff
@@ -69,7 +69,7 @@ pub fn migrate(deps: DepsMut<'_>, _env: Env, _msg: MigrateMsg) -> ContractResult
     #[cfg(not(feature = "migration"))]
     let resp = versioning::update_software(deps.storage, version!(CONTRACT_STORAGE_VERSION))
         .and_then(response::response);
-    resp.or_else(|err| log_error(err, deps.api))
+    resp
```
gmanev7, line 72: what's the motivation ?

contracts/oracle/src/error.rs
```diff
@@ -8,9 +8,52 @@ use sdk::cosmwasm_std::{Addr, StdError};
 
 #[derive(Error, Debug, PartialEq)]
 pub enum ContractError {
+    #[cfg(feature = "testing")]
     #[error("[Oracle] [Std] {0}")]
     Std(#[from] StdError),
 
+    #[error("[Oracle; Stub] Failed to query configuration! Context: {0}")]
```
gmanev7, line 15: [everywhere] Please replace `Context` with `Cause`

contracts/oracle/src/error.rs
```diff
@@ -8,9 +8,52 @@ use sdk::cosmwasm_std::{Addr, StdError};
 
 #[derive(Error, Debug, PartialEq)]
 pub enum ContractError {
+    #[cfg(feature = "testing")]
```
gmanev7, line 11: let's minimize conditional compilation and avoid test-custom errors

contracts/lease/src/contract/endpoins.rs
KirilMihaylov: That must have happened during the rebase as it was really old at this point.

contracts/oracle/src/error.rs
gmanev7: remove the Std variant as you did elsewhere and `.unwrap()`

contracts/oracle/src/error.rs
gmanev7: by "everywhere" I meant any occurrence in this PR including marketprice errors

packages/versioning/src/lib.rs
```diff
@@ -94,18 +94,11 @@ pub fn initialize(storage: &mut dyn Storage, version: Version) -> StdResult<()>
...
 }
 
-pub fn update_software<ContractError>(
-    storage: &mut dyn Storage,
-    new: Version,
-) -> Result<ReleaseLabel, ContractError>
-where
-    StdError: Into<ContractError>,
-{
+pub fn update_software(storage: &mut dyn Storage, new: Version) -> Result<ReleaseLabel, StdError> {
```
gmanev7, line 97: just noticed that the return error is not consistent with `update_software_and_storage`. Update it as well

packages/versioning/src/lib.rs
```diff
@@ -94,33 +94,45 @@ pub fn initialize(storage: &mut dyn Storage, version: Version) -> StdResult<()>
...
+where
+    MapErrorFunctor: Fn(StdError) -> ContractError,
+{
     load_version(storage)
-        .and_then(|current| release::allow_software_update(&current, &new))
-        .and_then(|()| save_version(storage, &new))
+        .map_err(&map_error)
+        .and_then(|current| release::allow_software_update(&current, &new).map_err(&map_error))
+        .and_then(|()| save_version(storage, &new).map_err(map_error))
         .map(|()| release::label())
```
gmanev7, line 109: is not possible to `.map_err` only once at the end?

KirilMihaylov: That's a very good point. Thank you!

gmanev7 (approved):

PR #125, opened 2023-07-05
refactor: Refactor integration tests to unify response with Neutron message passing

KirilMihaylov:


tests/src/common/test_case.rs
```diff
@@ -288,9 +291,197 @@ impl<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
...
+impl WrappedApp {
+    pub const fn new(app: MockApp, message_receiver: WrappedCustomMessageReceiver) -> Self {
+        Self {
+            app,
+            message_receiver,
+        }
+    }
+
+    #[must_use]
+    #[inline]
```
gmanev7, line 309: let's not clutter the code with useless compiler directives

tests/src/common/test_case.rs
```diff
@@ -288,9 +291,197 @@ impl<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
     }
 }
 
-pub struct TestCase<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms> {
-    pub app: MockApp,
-    pub message_receiver: WrappedCustomMessageReceiver,
+#[must_use]
+pub(crate) struct WrappedApp {
```
gmanev7, line 295: May we find better names than "WrappedXyz"?

packages/sdk/src/testing.rs
```diff
@@ -37,6 +37,7 @@ pub type Contract = dyn cw_multi_test::Contract<CustomMsg>;
 pub type CustomMessageSender = std::sync::mpsc::Sender<CustomMsg>;
 type CustomMessageReceiver = std::sync::mpsc::Receiver<CustomMsg>;
 
+#[derive(Debug)]
 pub struct WrappedCustomMessageReceiver(CustomMessageReceiver);
```
gmanev7, line 41: better naming pls

tests/src/common/dispatcher_wrapper.rs
```diff
@@ -4,27 +4,26 @@ use rewards_dispatcher::{
...
+use sdk::cosmwasm_std::Addr;
 
-use crate::common::{ContractWrapper, MockApp};
+use super::{test_case::WrappedApp, ContractWrapper, ADMIN};
 
-use super::ADMIN;
-
-pub struct DispatcherWrapper {
+pub(crate) struct DispatcherWrapper {
     contract_wrapper: Box<DispatcherContractWrapper>,
```
gmanev7, line 12: why we use Box here?

tests/src/common/test_case.rs
KirilMihaylov: Those are the tests, so I am not quite sure why would the `#[inline]` matter, but I'd like to keep the `#[must_use]` as there are way too many places where the results were just implicitly ignored. That would at least make it clear, that we specifically say "Yes, I don't care about this value".

tests/src/common/dispatcher_wrapper.rs
KirilMihaylov: `App::store_code` requires `Box<dyn Contract>` so that is consumed on instantiation. To be honest, it doesn't seem to matter where we put it, because it will be there no matter what.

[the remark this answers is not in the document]
KirilMihaylov: That's not actually a mock, but rather the full-fledged contract.

tests/src/common/test_case.rs
KirilMihaylov: I'll be happy to hear any suggestions, because I can't think of any right now.

tests/src/common/test_case.rs
gmanev7: don't or do care?

tests/src/common/test_case.rs
gmanev7: it's part of the developer's responsabilities

tests/src/common/test_case.rs
KirilMihaylov: The secondary idea behind this PR is to make the "I don't care about this value" explicit, thus the `#[must_use]`.

tests/src/common/test_case.rs
```diff
@@ -288,9 +291,192 @@ impl<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
...
+    pub fn query(&self) -> QuerierWrapper<'_, Empty> {
+        self.message_receiver.assert_empty();
+
+        self.app.wrap()
+    }
+}
+
+#[must_use]
+#[derive(Debug)]
+pub struct WrappedResponse<'r, T> {
```
gmanev7, line 429: PLEASE do not name new types with prefixes or suffixes that do not matter anything from a business/domain perspective

tests/src/common/test_case.rs
gmanev7, line 1: this file is getting close to 1000 lines of code! Try to keep the files in meaningful length. Any types that are not super tightly connected please move to their own modules and files

tests/src/common/test_case.rs
```diff
@@ -288,9 +291,192 @@ impl<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
...
+    receiver: &'r mut CustomMessageReceiverExt,
```
gmanev7, line 430: maybe I am missing any detail ... but cannot find the motivation to bounce it forward and backward?

tests/src/common/test_case.rs
```diff
@@ -288,9 +291,192 @@ impl<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
...
+    }
+}
+
+impl<'r> WrappedResponse<'r, ()> {
+    pub fn set_result<T>(self, result: T) -> WrappedResponse<'r, T> {
+        WrappedResponse {
+            receiver: self.receiver,
+            result,
+        }
+    }
```
gmanev7, line 474: this smells ... so let's rethink the need for "WrappedResponse" ... leaving the naming aside

gmanev7 (commented):
most of the questions relate to test_case.rs code

packages/sdk/src/testing.rs
```diff
@@ -32,14 +30,15 @@ pub type AppBuilder<Exec = CustomMsg, Query = Empty> = cw_multi_test::AppBuilder
...
 
-pub type Contract = dyn cw_multi_test::Contract<CustomMsg>;
+pub type CwContract = dyn cw_multi_test::Contract<CustomMsg>;
 
 pub type CustomMessageSender = std::sync::mpsc::Sender<CustomMsg>;
 type CustomMessageReceiver = std::sync::mpsc::Receiver<CustomMsg>;
 
-pub struct WrappedCustomMessageReceiver(CustomMessageReceiver);
+#[derive(Debug)]
+pub struct CustomMessageReceiverExt(CustomMessageReceiver);
```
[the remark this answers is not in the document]
KirilMihaylov: I respectfully disagree because this is both outside our business logic and not necessarily a DEX message. I am willing to move the methods within a trait that extends the receiver, but not calling it "Dex" related.

packages/sdk/src/testing.rs
```diff
@@ -32,14 +30,15 @@ pub type AppBuilder<Exec = CustomMsg, Query = Empty> = cw_multi_test::AppBuilder
     FailingModule<GovMsg, Empty, Empty>,
 >;
 
-pub type Contract = dyn cw_multi_test::Contract<CustomMsg>;
+pub type CwContract = dyn cw_multi_test::Contract<CustomMsg>;
 
 pub type CustomMessageSender = std::sync::mpsc::Sender<CustomMsg>;
```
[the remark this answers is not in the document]
KirilMihaylov: As mentioned, doesn't have to be a DEX message, thus a different name would be more fitting. I have chosen `Custom` from the beginning as that is what the CW library calls it, thus staying coherent between libraries.

tests/src/common/dispatcher_wrapper.rs
```diff
@@ -4,27 +4,26 @@ use rewards_dispatcher::{
...
-use sdk::{cosmwasm_std::Addr, cw_multi_test::Executor};
+use sdk::cosmwasm_std::Addr;
 
-use crate::common::{ContractWrapper, MockApp};
+use super::{test_case::App, CwContractWrapper, ADMIN};
 
-use super::ADMIN;
-
-pub struct DispatcherWrapper {
+pub(crate) struct DispatcherWrapper {
```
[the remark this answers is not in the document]
KirilMihaylov: I think I didn't understand you correctly. Those are testing structures, if I move them out to their respective crates, that would just make it more spaghetti than it already is.

tests/src/common/test_case.rs
KirilMihaylov: I, again, respectfully disagree because this type has nothing to do with business logic nor the domain itself, it is a type that is supposed to help with testing.

tests/src/common/test_case.rs
KirilMihaylov: Idea behind decision is that this will ensure that there won't be another mutable call before all messages are consumed after the last one.

This makes it hold the lifetime mutably, thus denying access to the `App` before the response is actually consumed.
That way there is a significantly lower chance messages can be "forgotten" by one caller and be spuriously consumed by the next one.

tests/src/common/test_case.rs
KirilMihaylov: Why do you think it does? It is only implemented when the response data is from the unit type, `()`.

tests/src/common/dispatcher.rs
```diff
@@ -1,30 +1,35 @@
...
-        app: &mut MockApp,
+        app: &mut App,
         lpp: Addr,
         oracle: Addr,
         timealarms: Addr,
         treasury: Addr,
     ) -> Addr {
-        let code_id = app.store_code(self.contract_wrapper);
+        // TODO [Rust 1.70] Convert to static item with OnceCell
+        let endpoints: CwContractWrapper<_, _, _, _, _, _, _, _, _, _, _> = CwContractWrapper::new(
```
gmanev7, line 24: [anywhere] sometimes providing types to references does not look perfect, think about removing it ... after all the type could be easily inferred by looking at the code ;-)

tests/src/common/lease.rs
```diff
@@ -0,0 +1,382 @@
...
+    } else {
+        exp_lease
+    };
+
+    swap_response::<DownpaymentC, LeaseC>(app, lease_addr, exp_swap_out, remote_tx_count);
+
+    check_state_opened(app, lease_addr);
+}
+
+fn open_ica<'r>(
```
gmanev7, line 266: rename to 'send_open_ica_response'

tests/src/common/lease.rs
```diff
@@ -0,0 +1,382 @@
...
+        lease_addr.as_str(),
+        ica_addr,
+    );
+
+    () = response.unwrap_response();
+
+    check_state_opening(app, lease_addr);
+
+    let mut response: ResponseWithInterChainMsgs<'_, ()> =
+        send_blank_response(app, lease_addr).ignore_result();
```
gmanev7, line 227: in an attempt to reduce this function length, extract this into a new fn, and reuse twice

tests/src/common/lease.rs
```diff
@@ -0,0 +1,382 @@
...
+        lease_addr.as_str(),
+        ica_addr,
+    );
+
+    () = response.unwrap_response();
+
+    let mut response: ResponseWithInterChainMsgs<'_, ()> =
+        send_blank_response(app, lease_addr).ignore_result();
+
+    let remote_tx_count: usize = usize::from(!currency::equal::<Lpn, Native>())
```
gmanev7, line 241: do not understand why you compare Lpn to Native? Native should always be != Lpn ... but even  if it is in the tests then we should care here if Lpn == LeaseC instead

tests/src/common/lease.rs
```diff
@@ -0,0 +1,382 @@
...
+        + usize::from(
+            !(currency::equal::<DownpaymentC, LeaseC>()
```
gmanev7, line 243: similarly to the above, we should care only if DownpaymentC == LeaseC

tests/src/common/lease.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    check_state_opening(app, lease_addr);
+
+    let exp_swap_out = if currency::equal::<DownpaymentC, LeaseC>() {
+        exp_lease - price::total(downpayment, Price::identity())
+    } else {
+        exp_lease
+    };
+
+    swap_response::<DownpaymentC, LeaseC>(app, lease_addr, exp_swap_out, remote_tx_count);
```
gmanev7, line 261: rename to 'send_swap_response'

tests/src/common/lease.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    check_state_opened(app, lease_addr);
```
gmanev7, line 263: extract into 'do_swap' fn

gmanev7 (commented):
this PR is taking good shape ... just a few more updates needed

tests/src/common/lease.rs
KirilMihaylov: I can't agree with that because on native it wouldn't(/shouldn't) produce an interchain message.

tests/src/common/test_case/response.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    #[must_use]
+    #[track_caller]
+    pub fn unwrap_response(mut self) -> T {
+        self.expect_empty();
+
+        self.result
+    }
+}
```
gmanev7, line 31: Here we use 'result' and 'response' interchangeably. Let's stick to 'response' as being closer to what we are trying to represent.

tests/src/common/test_case/address_book.rs
```diff
@@ -0,0 +1,235 @@
...
+        &self.time_alarms_addr
+    }
+}
+
+impl<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
+    AddressBook<Dispatcher, Treasury, Profit, Leaser, Lpp, Oracle, TimeAlarms>
+{
+    pub(super) fn set_lease_code_id(&mut self, lease_code_id: u64) {
+        self.lease_code_id = lease_code_id;
+    }
```
gmanev7, line 230: please return a new instance than mutating self

KirilMihaylov: I did but then the problem occurred when calling `TestCase::store_new_lease_code` because it is behind a mutable reference and doesn't allow me to move it out.

tests/src/common/test_case/builder.rs
```diff
@@ -0,0 +1,408 @@
...
+            _lpn,
+        }
+    }
+}
+
+impl<Lpn, Dispatcher, Leaser, Lpp> Builder<Lpn, Dispatcher, Addr, (), Leaser, Lpp, Addr, Addr>
+where
+    Lpn: Currency,
+{
+    pub fn init_profit(
```
gmanev7, line 151: please split it in a similar way we did for 'complete_lease_instantiation'

tests/src/common/test_case/app.rs
```diff
@@ -0,0 +1,141 @@
...
+
+                Err(error)
+            }
+        }
+    }
+
+    #[must_use]
+    pub fn query(&self) -> QuerierWrapper<'_, Empty> {
+        self.app.wrap()
+    }
```
gmanev7, line 140: realizing the following is a bit out-of-scope, we can simplify the query-like code in 10x test code locations. We'll avoid the dependency on cosmwasm_std::QuerierWrapper by mimicking its API here and removing this method.

tests/src/oracle_tests.rs
```diff
@@ -94,16 +99,16 @@ fn register_feeder() {
...
-    oracle_wrapper::add_feeder(&mut test_case, ADMIN);
+    add_feeder(&mut test_case, ADMIN);
 }
 
 #[test]
 fn internal_test_integration_setup_test() {
     let mut test_case = create_test_case();
 
-    oracle_wrapper::add_feeder(&mut test_case, ADMIN);
+    add_feeder(&mut test_case, ADMIN);
```
gmanev7, line 109: [many occurences] What's the motivation for using the functions by name only? Rust idioms recommend other

tests/src/common/test_case/response.rs
```diff
@@ -6,18 +6,18 @@ use sdk::{
...
 pub(crate) struct ResponseWithInterChainMsgs<'r, T> {
     receiver: &'r mut InterChainMsgReceiver,
-    result: T,
+    response: T,
 }
 
 impl<'r, T> ResponseWithInterChainMsgs<'r, T> {
     pub(super) fn new(receiver: &'r mut InterChainMsgReceiver, result: T) -> Self {
-        Self { receiver, result }
+        Self { receiver, response: result }
```
gmanev7: [nit] rename the fn arg as well

tests/src/oracle_tests.rs
KirilMihaylov: Interference between the crate and the module, because now that both are with the same name it causes problems when trying to distinguish it.

tests/src/common/test_case/app.rs
KirilMihaylov: While yes, I'd prefer this to be done in a separate PR, so we can merge this and focus on the tests themselves first.

tests/src/common/lease.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+fn check_state_opened(app: &mut App, lease: &Addr) {
+    let StateResponse::Opened { .. } = app.query().query(&QueryRequest::Wasm(WasmQuery::Smart {
+        contract_addr: lease.to_string(),
+        msg: to_binary(&StateQuery {}).unwrap(),
+    })).unwrap() else {
+        panic!("Opening lease failed! Lease is not yet it opened state!");
+    };
+}
```
gmanev7, line 360: please remove code duplication

gmanev7 (approved):

PR #129, opened 2023-07-11
intro UserPermission

gmanev7:


contracts/dispatcher/src/contract.rs
```diff
@@ -83,8 +83,8 @@ pub fn execute(
 ) -> ContractResult<CwResponse> {
     match msg {
         ExecuteMsg::TimeAlarm {} => {
-            SingleUserAccess::load(deps.storage, crate::access_control::TIMEALARMS_NAMESPACE)?
-                .check_access(&info.sender)?;
+            SingleUserAccess::new(&*deps.storage, crate::access_control::TIMEALARMS_NAMESPACE)
```
KirilMihaylov, line 86: Just to clear things up, which one do we prefer because this was something you mentioned in previous PR? `&*` or `.deref()`?

packages/access-control/src/lib.rs
```diff
@@ -1,147 +1,123 @@
...
+            .save(self.storage.deref_mut(), user)
+            .map_err(Into::into)
     }
+}
 
-    pub const fn address(&self) -> &Addr {
-        &self.address
-    }
+#[derive(Debug, Eq, PartialEq)]
+struct UserPermission {
```
KirilMihaylov, line 51: I feel like a function/static method would've done the job just fine. This feels a bit like abusing OOP.

KirilMihaylov (commented):
Looks good apart from two comments.

contracts/dispatcher/src/contract.rs
gmanev7: ooops, as we agreed before, we stick to the more explanatory `.deref()`

packages/access-control/src/lib.rs
gmanev7: true, did not expect initially that the impl would be so simple

PR #131, opened 2023-07-14
chore: Bump versions [Has changes]

KirilMihaylov:


packages/platform/src/ica.rs
```diff
@@ -42,6 +43,12 @@ impl From<HostAccount> for String {
     }
 }
 
+impl AsRef<str> for HostAccount {
+    fn as_ref(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
```
gmanev7, line 51: why do you implement it? To simulate Display? If it's so then derive it

gmanev7 (approved):

PR #137, opened 2023-07-18
[lease] Safe ICA open response delivery

gmanev7:


packages/dex/src/resp_delivery/adapter.rs
```diff
@@ -0,0 +1,20 @@
+use sdk::cosmwasm_std::{Binary, Deps, Env};
+
+use crate::{response::Result, Handler};
+
+pub trait DeliveryAdapter<H, R>
```
KirilMihaylov, line 5: What's the idea behind the trait is it's not exposed and only implemented once?

gmanev7: The aim is to abstract the delivery of response be it via `on_response` or `on_open_ica`. Tha latter follows in the next commits

packages/dex/src/out_local.rs
```diff
@@ -257,6 +302,9 @@ mod impl_handler {
                     // TODO do go through RespDelivery
                     Handler::on_open_ica(inner, counterparty_version, deps, env)
                 }
+                State::SwapExactInRecoverIcaRespDelivery(inner) => {
+                    Handler::on_open_ica(inner, counterparty_version, deps, env)
+                }
```
KirilMihaylov, line 307: I know very well that you are very much against usage of macros, but for such repetitions, I'd strongly suggest it.

PR #140, opened 2023-07-20
fix(profit): BuyBack `on_coins` logic correction

KirilMihaylov:


gmanev7 (commented):
add unit tests ... this bug deserves them

gmanev7 (commented):
why have you added extra dependencies not needed at this moment? That would slow down the compilation and add extra relations to maintain.

KirilMihaylov:
> why have you added extra dependencies not needed at this moment? That would slow down the compilation and add extra relations to maintain.

I have added them because parts of the code already needed them. E.g. derivation of `Debug` was in place behind conditional compilation, but was not used as of now, thus hiding the problem.

Also, when the whole workspace is built, no new features are actually pulled as can be seen by the fact that `Cargo.lock` hasn't changed. That only affects compilation when `--package` is set and not always even.

contracts/profit/src/state/buy_back.rs
```diff
@@ -96,13 +98,15 @@ impl SwapTask for BuyBack {
...
                 .visit(coin)
                 .map(|result: IterNext| matches!(result, IterNext::Stop))
         })
-        .map(|maybe_coin: Option<&CoinDTO<PaymentGroup>>| {
-            if maybe_coin.is_some() {
+        .map(|_| {
+            if coins_iter.as_slice().is_empty() {
                 IterState::Complete
             } else {
                 IterState::Incomplete
```
gmanev7, line 112: the previous implementation seems to work ... just switch the `then` with the `else` blocks

KirilMihaylov: I can't agree with that because it can "find" the last element, in which it would've resolved the whole sequence. Doing that directly contradicts with the debug assertions in the calling component.

KirilMihaylov: Took the liberty of reverting locally and replacing `Option::is_some` with `Option::is_none`. Other tests begin to fail, thus not a wanted behavior.

gmanev7:
> I have added them because parts of the code already needed them. E.g. derivation of `Debug` was in place behind conditional compilation, but was not used as of now, thus hiding the problem.
> 
> Also, when the whole workspace is built, no new features are actually pulled as can be seen by the fact that `Cargo.lock` hasn't changed. That only affects compilation when `--package` is set and not always even.

 more code higher maintenance cost

contracts/leaser/src/migrate.rs
```diff
@@ -2,12 +2,11 @@ use lease::api::MigrateMsg;
 use platform::batch::Batch;
 use sdk::cosmwasm_std::Addr;
 
-use crate::{msg::MaxLeases, result::ContractResult};
-
-pub struct Customer<LeaseIter> {
-    customer: Addr,
-    leases: LeaseIter,
-}
```
gmanev7, line 10: 1. What is the motivation to include this change in this PR?
2. The previous name was descriptive enough. What if tomorrow we add another attribute to the customer?

KirilMihaylov: 1) It needed to be moved because it was used by code outside the migration one, which is guarded by a feature. This was part of making sure everything compiles just fine on it's own.
2) I suppose it was. Perhaps I could revert it.

gmanev7: 1. if I am not mistaken we do not change the features around this code
2. please do

PR #142, opened 2023-07-24
fix: Reset time alarm on configuration update

### LINKED ISSUE #122: [Profit] Reset the time alarm on updating the cadence
Currently, we have to fire an extra time alarm to take the new cadence into account

KirilMihaylov:
Closes #122

### DIFF
--- a/contracts/profit/src/state/buy_back.rs
+++ b/contracts/profit/src/state/buy_back.rs
@@ -17,11 +17,15 @@ use platform::{
     bank::{self, BankAccountView},
     message::Response as PlatformResponse,
     never::Never,
+    state_machine::Response as StateMachineResponse,
 };
-use sdk::cosmwasm_std::{Addr, Env, QuerierWrapper};
+use sdk::cosmwasm_std::{Addr, Env, QuerierWrapper, Timestamp};
 use timealarms::stub::TimeAlarmsRef;
 
-use crate::{error::ContractError, msg::ConfigResponse, profit::Profit, result::ContractResult};
+use crate::{
+    error::ContractError, msg::ConfigResponse, profit::Profit, result::ContractResult,
+    typedefs::CadenceHours,
+};
 
 use super::{
     idle::Idle,
@@ -137,10 +141,11 @@ impl SwapTask for BuyBack {
 }
 
 impl ConfigManagement for StateLocalOut<BuyBack, ForwardToDexEntry, ForwardToDexEntryContinue> {
-    fn with_config<F>(self, _: F) -> ContractResult<super::StateAndResponse<Self>>
-    where
-        F: FnOnce(Config) -> ContractResult<super::ConfigAndResponse>,
-    {
+    fn try_update_config(
+        self,
+        _: Timestamp,
+        _: CadenceHours,
+    ) -> ContractResult<StateMachineResponse<Self>> {
         Err(ContractError::UnsupportedOperation(String::from(
             "Configuration changes are not allowed during ICA opening process.",
         )))
--- a/contracts/profit/src/state/buy_back.rs
+++ b/contracts/profit/src/state/buy_back.rs
@@ -21,10 +21,7 @@ use platform::{
 use sdk::cosmwasm_std::{Addr, Env, QuerierWrapper};
 use timealarms::stub::TimeAlarmsRef;
 
-use crate::{
-    error::ContractError, msg::ConfigResponse, profit::Profit, result::ContractResult,
-    typedefs::CadenceHours,
-};
+use crate::{error::ContractError, msg::ConfigResponse, profit::Profit, result::ContractResult};
 
 use super::{
     idle::Idle,
@@ -140,7 +137,10 @@ impl SwapTask for BuyBack {
 }
 
 impl ConfigManagement for StateLocalOut<BuyBack, ForwardToDexEntry, ForwardToDexEntryContinue> {
-    fn try_update_config(self, _: CadenceHours) -> ContractResult<Self> {
+    fn with_config<F>(self, _: F) -> ContractResult<super::StateAndResponse<Self>>
+    where
+        F: FnOnce(Config) -> ContractResult<super::StateAndResponse<Config>>,
+    {
         Err(ContractError::UnsupportedOperation(String::from(
             "Configuration changes are not allowed during ICA opening process.",
         )))
--- a/contracts/profit/src/state/buy_back.rs
+++ b/contracts/profit/src/state/buy_back.rs
@@ -21,7 +21,7 @@ use platform::{
 use sdk::cosmwasm_std::{Addr, Env, QuerierWrapper};
 use timealarms::stub::TimeAlarmsRef;
 
-use crate::{profit::Profit, result::ContractResult};
+use crate::{error::ContractError, msg::ConfigResponse, profit::Profit, result::ContractResult};
 
 use super::{
     idle::Idle,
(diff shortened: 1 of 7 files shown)


contracts/profit/src/state/idle.rs
```diff
@@ -117,11 +117,19 @@ impl Enterable for Idle {
...
-            config: self.config.update(cadence_hours),
-            ..self
-        })
+    fn with_config<F>(self, f: F) -> ContractResult<StateAndResponse<Self>>
+    where
+        F: FnOnce(Config) -> ContractResult<StateAndResponse<Config>>,
+    {
+        f(self.config).map(
+            |StateAndResponse {
+                 state: config,
```
gmanev7, line 126: obviously, passing the new config via the 'state' is a hack we do not want in our code.
Please do not misuse State Machine State for other purposes.

KirilMihaylov: I used the same structure to reuse code, but your concern is valid. I'll fork it into a separate one for the config.

contracts/profit/src/state/mod.rs
```diff
@@ -36,11 +38,52 @@ const STATE: Item<'static, State> = Item::new("contract_state");
...
+    where
+        T: Into<U>,
+    {
+        StateAndResponse {
+            state: self.state.into(),
+            response: self.response,
+        }
+    }
+}
+
```
gmanev7, line 62: let's not reinvent the wheel, there is `platform::state_machine::Response` which carries both messages and state, and the state includes the new config

in addition, let's not complicate the code with `WithConfig` trait is not needed

KirilMihaylov: As I mentioned in a comment before, I wanted to avoid repetition and this trait was the way to do it. If you insist, I will remote it but that will just reintroduce duplication.

contracts/profit/src/state/buy_back.rs
```diff
@@ -140,7 +141,11 @@ impl SwapTask for BuyBack {
...
 impl ConfigManagement for StateLocalOut<BuyBack, ForwardToDexEntry, ForwardToDexEntryContinue> {
-    fn try_update_config(self, _: CadenceHours) -> ContractResult<Self> {
+    fn try_update_config(
+        self,
+        _: Timestamp,
+        _: CadenceHours,
+    ) -> ContractResult<StateMachineResponse<Self>> {
         Err(ContractError::UnsupportedOperation(String::from(
             "Configuration changes are not allowed during ICA opening process.",
         )))
```
gmanev7, line 151: why don't we provide such an implementation at the trait level? While overwriting an implementation is not acceptable, providing a default rejecting logic saves code duplication.

KirilMihaylov: I admit that Ctrl+C/V is not a good approach, but this was somewhat of a dirty refactor to get the structure look better.

Side question: Why is not acceptable to override a default method?
Side note: If that's the case, then I would suggest having an `*Ext` trait which has blanket implementations for all of our traits that currently do have default implementations which shouldn't be changed.

contracts/profit/src/state/mod.rs
```diff
@@ -36,13 +40,33 @@ const STATE: Item<'static, State> = Item::new("contract_state");
...
 type IcaConnector = dex::IcaConnector<OpenIca, ContractResult<DexResponse<Idle>>>;
 
+fn on_config_update(new_config: &Config, now: Timestamp) -> ContractResult<PlatformResponse> {
+    new_config
+        .time_alarms()
+        .setup_alarm(now + Duration::from_hours(new_config.cadence_hours()))
+        .map(PlatformResponse::messages_only)
+        .map_err(Into::into)
+}
+
```
gmanev7, line 50: this is a copy/paste of `Idle::enter`, please remove

contracts/profit/src/state/mod.rs
```diff
@@ -36,13 +40,33 @@ const STATE: Item<'static, State> = Item::new("contract_state");
...
+            "Configuration changes are not allowed in this state.",
+        )))
+    }
 
-    fn try_query_config(&self) -> ContractResult<ConfigResponse>;
+    fn try_query_config(&self) -> ContractResult<ConfigResponse> {
+        Err(ContractError::UnsupportedOperation(String::from(
+            "Configuration querying is not allowed in this state.",
+        )))
+    }
```
gmanev7, line 69: looking into the code, I realized that it's not correct to refuse config queries. Therefore please remove the default implementation and provide one at each of the four states.

KirilMihaylov: I can't really do that for all four. I can do it only for the two where I have overridden it.

contracts/profit/src/state/mod.rs
KirilMihaylov: Wasn't even an intentionally made. Will clean it up

contracts/profit/src/state/idle.rs
```diff
@@ -117,10 +121,18 @@ impl Enterable for Idle {
...
+        cadence_hours: CadenceHours,
+    ) -> ContractResult<StateMachineResponse<Self>> {
+        let config: Config = self.config.update(cadence_hours);
+
+        super::on_config_update(&config, now).map(|response: PlatformResponse| {
+            StateMachineResponse {
+                response,
+                next_state: Self { config, ..self },
+            }
         })
```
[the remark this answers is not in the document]
KirilMihaylov: I don't feel like calling `enter` is logically sound thing to do. Might be better to instead reuse `on_config_update` in the form of something along the lines of `setup_time_alarm`.

contracts/profit/src/state/buy_back.rs
```diff
@@ -140,15 +137,9 @@ impl SwapTask for BuyBack {
...
-    fn try_update_config(self, _: CadenceHours) -> ContractResult<Self> {
-        Err(ContractError::UnsupportedOperation(String::from(
-            "Configuration changes are not allowed during ICA opening process.",
-        )))
-    }
-
     fn try_query_config(&self) -> ContractResult<ConfigResponse> {
         Err(ContractError::UnsupportedOperation(String::from(
-            "Querying configuration is not allowed during buy-back.",
+            "Configuration querying is not supported while opening interchain account!",
```
gmanev7, line 142: message?

contracts/profit/src/state/mod.rs
```diff
@@ -36,11 +42,31 @@ const STATE: Item<'static, State> = Item::new("contract_state");
...
 {
-    fn try_update_config(self, cadence_hours: CadenceHours) -> ContractResult<Self>;
+    fn try_update_config(
+        self,
+        _: Timestamp,
+        _: CadenceHours,
+    ) -> ContractResult<StateMachineResponse<Self>> {
+        Err(ContractError::UnsupportedOperation(String::from(
+            "Configuration changes are not allowed in this state!",
+        )))
```
gmanev7, line 68: three copies of this code deserve factoring a common fn out

gmanev7: profit/src/error.rs is the perfect location

contracts/profit/src/state/mod.rs
```diff
@@ -40,7 +40,15 @@ pub(crate) trait ConfigManagement
...
```
gmanev7, line 50: take advantage of the new ContractError fn, please

gmanev7 (approved):

PR #143, opened 2023-07-25
refactor: Clean up

KirilMihaylov:


contracts/dispatcher/src/state/mod.rs
```diff
@@ -5,19 +5,16 @@ use sdk::{
...
 pub mod dispatch_log;
-pub mod reward_scale;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 pub struct DispatchLog {
     pub last_dispatch: Timestamp,
 }
 
-pub type CadenceHours = u16;
-
```
gmanev7, line 20: what is the idea of this refactoring? The structure was good enough ... and since we do not add more code I doubt we need it ... moreover this change is in a wrong direction

gmanev7 (commented):
fully support the aspect of this PR which deals with dependencies .... and some of the changes of the conditional compilation
the restructuring is (1) out of scope and (2) in wrong direction

contracts/leaser/src/msg.rs
```diff
@@ -71,6 +72,7 @@ pub enum SudoMsg {
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum QueryMsg {
+    #[cfg(any(test, feature = "contract"))]
```
gmanev7, line 75: do not have time to dig into this change ... could you elaborate more on the motivation od doing it?

KirilMihaylov: It tries to expose code which is feature gated. Variants were to either lock `msg` module under the `contract` feature, which is an overkill, or to just filter out the parts which required it.

contracts/dispatcher/src/state/mod.rs
KirilMihaylov: Again, as I have mentioned before, the logic right now is actually the one in the wrong direction, because we are using types defined in `state`, which is feature gated, in `msg` which isn't.

gmanev7: a change in the right direction is one that does not complicate the stuff and the one that really reflects the motivation and usage of the code.

So, a quick look at the code reveals that the feature "contract" is totally useless because there is no valid usage of the package as a "non-contract". Therefore a simple change that eliminates that feature is a perfect fit.

contracts/leaser/src/msg.rs
gmanev7: in the spirit of the previous my comment this should be reverted as well

gmanev7 (approved):
LGTM

PR #144, opened 2023-07-27
feat(test): Configurable WASM module

KirilMihaylov:
Additionally includes a test for the lease contract, making use of new capabilities, simulating a system error.

### DIFF
--- a/tests/src/common/test_case/app.rs
+++ b/tests/src/common/test_case/app.rs
@@ -1,184 +1,23 @@
-use std::{cell::RefCell, fmt::Debug, rc::Rc};
+use std::fmt::Debug;
 
 use serde::Serialize;
 
 use finance::duration::Duration;
 use sdk::{
     cosmwasm_ext::{CosmosMsg, InterChainMsg},
-    cosmwasm_std::{
-        Addr, Api, Binary, BlockInfo, Coin as CwCoin, Empty, Querier, QuerierWrapper, Storage,
-        WasmMsg, WasmQuery,
-    },
-    cw_multi_test::{
-        AppResponse, Contract as CwContract, CosmosRouter, Executor, Wasm as WasmTrait, WasmKeeper,
-    },
+    cosmwasm_std::{Addr, BlockInfo, Coin as CwCoin, Empty, QuerierWrapper},
+    cw_multi_test::{AppResponse, Contract as CwContract, Executor},
     testing::{CwApp, InterChainMsgReceiver},
 };
 
 use crate::common::{test_case::response::ResponseWithInterChainMsgs, AppExt as _};
 
-pub(crate) trait Wasm: WasmTrait<InterChainMsg, Empty> + Sized {
-    type CounterPart;
-
-    fn store_code(
-        app: &mut CwApp<Self>,
-        counter_part: &mut Self::CounterPart,
-        code: Box<dyn CwContract<InterChainMsg, Empty>>,
-    ) -> u64;
-}
-
-pub(crate) type DefaultWasm = WasmKeeper<InterChainMsg, Empty>;
-
-impl Wasm for DefaultWasm {
-    type CounterPart = ();
-
-    fn store_code(
-        app: &mut CwApp<Self>,
-        &mut (): &mut Self::CounterPart,
-        code: Box<dyn CwContract<InterChainMsg, Empty>>,
-    ) -> u64 {
-        app.store_code(code)
-    }
-}
-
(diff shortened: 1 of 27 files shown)


tests/src/common/test_case/app.rs
```diff
@@ -1,34 +1,209 @@
...
+    cw_multi_test::{
+        AppResponse, Contract as CwContract, CosmosRouter, Executor, Wasm as WasmTrait, WasmKeeper,
+    },
+    testing::{CwApp, InterChainMsgReceiver},
 };
 
-use crate::common::{test_case::response::ResponseWithInterChainMsgs, AppExt as _, MockApp};
+use crate::common::{test_case::response::ResponseWithInterChainMsgs, AppExt as _};
+
+pub(crate) trait Wasm: WasmTrait<InterChainMsg, Empty> + Sized {
```
gmanev7, line 20: could you move the `Wasm`-related code into `wasm.rs`

tests/src/common/test_case/app.rs
```diff
@@ -1,34 +1,209 @@
...
+            inner,
+        )
+    }
+}
+
+pub(crate) struct ConfigurableWasm<ActionSelectionF>
+where
+    ActionSelectionF: Fn(Request<'_>) -> Action,
+{
+    inner: Rc<RefCell<DefaultWasm>>,
```
gmanev7, line 96: interior mutability needed??? could you elaborate on the rationale of defining `inner` of this type?

tests/src/lease/safe_delivery.rs
```diff
@@ -0,0 +1,294 @@
...
+};
+
+use super::{create_test_case_with_wasm, open_lease, Lpn, PaymentCoin, PaymentCurrency};
+
+#[test]
+fn safe_delivery() {
+    const ERROR_MESSAGE: &str = "Simulated error";
+
+    let forward_specific_execute: Cell<bool> = Cell::new(true);
+    let lease_addr_cell: Cell<&str> = Cell::new("DEADCODE");
```
gmanev7, line 35: please refrain from using Cell and RefCell types if their usage is really needed!

tests/src/common/test_case/app.rs
KirilMihaylov: The rationale is that I need to be able to call `store_code` from outside and CW's App stands in the way, not providing access to it. It's implementation is specialized for their own `WasmKeeper` structure and nothing else.
If you know a better solution, please let me know as this feedback would be useful for the future too!

tests/src/lease/safe_delivery.rs
KirilMihaylov: For the boolean, I could've used atomic or a channel, but I felt like that would be an overkill for this use.
For the lease address, same thing applies but most likely will have to include unsafe Rust if channel or a `Mutex` is not used.

KirilMihaylov: Also, that's the tests and not production code. I just feel like it would have been a bit easier to not complicate it and keep it simple.

gmanev7 (commented):
the tests in contracts/profit/src/tests/integration_tests.rs still use CwApp ... please refactor.

on the other hand, please assess their value

tests/src/common/test_case/wasm.rs
```diff
@@ -0,0 +1,168 @@
...
+
+pub(crate) trait Wasm: CwWasm<InterChainMsg, Empty> + Sized {
+    type CounterPart;
+
+    fn store_code(
+        app: &mut CwApp<Self>,
+        counter_part: &mut Self::CounterPart,
+        code: Box<dyn CwContract<InterChainMsg, Empty>>,
+    ) -> u64;
+}
```
gmanev7, line 20: could you elaborate on the need for this trait?

KirilMihaylov: I made the trait instead of directly plugging the new Wasm module to ensure compatibility with the original one. That way we won't need to use ours when not needed.

tests/src/lease/safe_delivery.rs
gmanev7: exactly the same as what we need ... and having interior mutable data and reference counting data is a step in the opposite direction. Could we `CwApp::store` each time we want to change the behavior?

KirilMihaylov: `CwApp::store` is specialized to only work with their own WasmKeeper. Thus the need for this PR.

PR #145, opened 2023-08-01
fix: Oracle dispatch error & minor macro refactor

KirilMihaylov:
Addition: Minor refactor to `oracle` macro to add the base currency as a parameter.

### DIFF
--- a/contracts/oracle/src/contract/oracle/mod.rs
+++ b/contracts/oracle/src/contract/oracle/mod.rs
@@ -257,10 +257,12 @@ mod tests {
             .store(storage)
             .unwrap();
 
-        SupportedPairs::<BaseCurrency>::new(swap_tree!(Usdc::TICKER, (1, Nls::TICKER)).into_tree())
-            .unwrap()
-            .save(storage)
-            .unwrap();
+        SupportedPairs::<BaseCurrency>::new(
+            swap_tree!({ base: Usdc::TICKER }, (1, Nls::TICKER)).into_tree(),
+        )
+        .unwrap()
+        .save(storage)
+        .unwrap();
     }
 
     #[track_caller]
--- a/contracts/oracle/src/contract/oracle/mod.rs
+++ b/contracts/oracle/src/contract/oracle/mod.rs
@@ -157,14 +157,9 @@ where
 }
 
 #[cfg(test)]
-mod tests {
+mod test_normalized_price_not_found {
     use currency::{lpn::Usdc, native::Nls, Currency as _};
-    use finance::{
-        coin::{Amount, Coin},
-        duration::Duration,
-        percent::Percent,
-        price,
-    };
+    use finance::{coin::Coin, duration::Duration, percent::Percent, price};
     use marketprice::{config::Config as PriceConfig, SpotPrice};
     use sdk::cosmwasm_std::{
         testing::{MockApi, MockQuerier, MockStorage},
@@ -185,60 +180,29 @@ mod tests {
     type NlsCoin = Coin<Nls>;
     type UsdcCoin = Coin<Usdc>;
 
+    const NOW: Timestamp = Timestamp::from_seconds(1);
+
     #[test]
     fn test() {
-        const SAMPLE_PERIOD_SECS: u32 = 3;
-        const _: () = if SAMPLE_PERIOD_SECS < 3 {
-            panic!("Bug is reproduced with minimum of 3 seconds for the sample period.");
-        };
-
         let mut storage: MockStorage = MockStorage::new();
 
         let price_config: PriceConfig = PriceConfig::new(
             Percent::HUNDRED,
-            Duration::from_secs(SAMPLE_PERIOD_SECS),
+            Duration::from_secs(1),
             1,
             Percent::HUNDRED,
         );
 
-        let mut now_seconds: u64 = 1;
-
         init(&mut storage, &price_config);
 
-        add_higher_alarm(&mut storage);
+        add_alarm(&mut storage);
 
-        feed_below_price(&price_config, &mut storage, now_seconds);
+        feed_price(&price_config, &mut storage);
 
-        now_seconds += 1;
-
-        dispatch_1(&mut storage, now_seconds);
-
-        add_lower_alarm(&mut storage);
-
-        now_seconds += 1;
-
-        for _ in 0..SAMPLE_PERIOD_SECS.checked_add(1).unwrap() {
-            dispatch_0(&mut storage, now_seconds);
-
-            now_seconds += 1;
-        }
-
-        feed_normal_price(&price_config, &mut storage, now_seconds);
-
-        now_seconds += 1;
-
-        dispatch_1(&mut storage, now_seconds);
-
-        now_seconds += 1;
+        dispatch_and_deliver(&mut storage, 1);
 
         // Bug happens on this step.
-        dispatch_0(&mut storage, now_seconds);
-
-        feed_below_price(&price_config, &mut storage, now_seconds);
-
-        now_seconds += 1;
-
-        dispatch_0(&mut storage, now_seconds);
+        dispatch_and_deliver(&mut storage, 0);
     }
 
     #[track_caller]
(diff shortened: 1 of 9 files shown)


packages/marketprice/src/alarms/mod.rs
```diff
@@ -395,6 +395,68 @@ pub mod tests {
...
     }
 
+    #[test]
+    fn test_out_for_delivery_removes_above() {
+        const PRICE_BASE: Amount = 1;
+        const PRICE_QUOTE: Amount = if let Some(quote) = PRICE_BASE.checked_add(1) {
+            quote
+        } else {
+            panic!()
+        };
```
gmanev7, line 405: [everywhere] It's an overkill ... please use the operator + and - ... after all we always panic when overflow

contracts/oracle/src/contract/oracle/mod.rs
```diff
@@ -155,3 +155,214 @@ where
...
+    #[track_caller]
+    fn add_higher_alarm(storage: &mut dyn Storage) {
+        add_alarm(storage, 1, 3)
+    }
+
+    #[track_caller]
+    fn add_lower_alarm(storage: &mut dyn Storage) {
+        add_alarm(storage, 2, 2)
+    }
+
```
gmanev7, line 297: replace this code with price constants and pass them to `add_alarm` not `Amount`-s

contracts/oracle/src/contract/oracle/mod.rs
```diff
@@ -155,3 +155,214 @@ where
...
+    #[track_caller]
+    fn feed_below_price(price_config: &PriceConfig, storage: &mut dyn Storage, now_seconds: u64) {
+        feed_price(price_config, storage, now_seconds, 1, 1)
+    }
+
+    #[track_caller]
+    fn feed_normal_price(price_config: &PriceConfig, storage: &mut dyn Storage, now_seconds: u64) {
+        feed_price(price_config, storage, now_seconds, 1, 2)
+    }
```
gmanev7, line 326: get rid of this chunk simmilarly to the `add_alarm` comment above

contracts/oracle/src/contract/oracle/mod.rs
```diff
@@ -155,3 +155,214 @@ where
...
+        now_seconds += 1;
+
+        // Bug happens on this step.
+        dispatch_0(&mut storage, now_seconds);
+
+        feed_below_price(&price_config, &mut storage, now_seconds);
+
+        now_seconds += 1;
+
+        dispatch_0(&mut storage, now_seconds);
```
gmanev7, line 241: please structure it further ... now it is ~ 50 lines long and is difficult to read

packages/marketprice/src/alarms/mod.rs
```diff
@@ -395,6 +395,68 @@ pub mod tests {
...
+
+        alarms.last_delivered().unwrap();
+
+        alarms.ensure_no_in_delivery().unwrap();
+
+        // Catch below
+        assert_eq!(alarms.alarms(lower_price).count(), 0);
+
+        // Catch above or equal
+        assert_eq!(alarms.alarms(price).count(), 0);
```
gmanev7, line 457: please structure it further ... now it is ~ 60 lines long and is difficult to read

gmanev7 (commented):
LGTM except how easy it is to read some long functions ... and a few constants spread here and there

contracts/oracle/src/contract/oracle/mod.rs
```diff
@@ -155,3 +155,137 @@ where
...
+            Oracle::<'_, _, BaseCurrency>::load(storage).unwrap();
+
+        let alarms: u32 = oracle.try_notify_alarms(NOW, 16).unwrap().0;
+
+        assert_eq!(alarms, expected_count);
+    }
+
+    #[track_caller]
+    fn deliver(storage: &mut dyn Storage, count: u32) {
+        let storage: &mut dyn Storage = storage;
```
gmanev7, line 276: ???

KirilMihaylov: Leftover from the minimization. Removing it now.

packages/finance/src/price/mod.rs
```diff
@@ -66,23 +66,45 @@ where
...
     }
 
+    pub const fn base_amount(&self) -> Coin<C> {
+        self.amount
+    }
+
+    pub const fn quote_amount(&self) -> Coin<QuoteC> {
+        self.amount_quote
+    }
+
```
gmanev7, line 107: Price is not a bag of two amounts everyone should know about. please remove. Moreover you use them only for tests

packages/finance/src/coin/mod.rs
```diff
@@ -77,16 +80,21 @@ where
...
 
-        debug_assert_eq!(self.amount % gcd, 0);
-        debug_assert_eq!(other.amount % gcd, 0);
+        #[cfg(debug_assertions)]
+        if self.amount % gcd != 0 {
+            panic!("LHS-value's amount is not divisible by the GCD!");
+        } else if other.amount % gcd != 0 {
+            panic!("RHS-value's amount is not divisible by the GCD!");
+        }
+
```
gmanev7, line 97: a simple `debug_assert!()`  would suffice!

packages/finance/src/coin/mod.rs
```diff
@@ -40,10 +39,14 @@ where
...
 
-    pub fn is_zero(&self) -> bool {
+    pub const fn is_zero(&self) -> bool {
         self.amount == Zero::ZERO
     }
 
+    pub const fn into_amount(self) -> Amount {
+        self.amount
+    }
+
```
gmanev7, line 49: please remove it, I understand the motivation ... but that should not happen at this price

packages/finance/src/coin/mod.rs
KirilMihaylov: With all my respect, it does not. I am adding `const` modifier which doesn't play nice with `{|debug_}assert{|_eq|_ne}!` macros. If I didn't have to change it, I wouldn't have.

packages/finance/src/coin/mod.rs
KirilMihaylov: We don't have a way to neither do arithmetic, nor use trait methods in constant time. What do you suggest that I do?

packages/finance/src/price/mod.rs
KirilMihaylov: That would reintroduce more constants in the tests again, but I'll do as you wish.

packages/finance/src/price/mod.rs
```diff
@@ -66,23 +66,45 @@ where
...
+        #[cfg(debug_assertions)]
+        res.invariant_held().unwrap();
+
+        res
+    }
+
+    #[cfg(feature = "testing")]
+    pub const fn unchecked(amount: Coin<C>, amount_quote: Coin<QuoteC>) -> Self {
+        Self::new_inner(amount, amount_quote)
+    }
```
gmanev7, line 80: there is almost the same alternative, why have you added this?

KirilMihaylov: `invariant_held` can not be ran in compile time without further refactoring, thus the need for this method.

packages/finance/src/coin/mod.rs
gmanev7: a local compilation passes with not errors or warnings

packages/finance/src/coin/mod.rs
```diff
@@ -77,16 +76,26 @@ where
...
 
-        debug_assert_eq!(self.amount % gcd, 0);
-        debug_assert_eq!(other.amount % gcd, 0);
+        debug_assert!(
+            self.amount % gcd == 0,
+            "LHS-value's amount is not divisible by the GCD!"
+        );
+        debug_assert!(
+            other.amount % gcd == 0,
+            "RHS-value's amount is not divisible by the GCD!"
```
gmanev7, line 96: the description should be inverted

gmanev7 (approved):

PR #149, opened 2023-08-03
ci: Fix size error report & update limit to match

KirilMihaylov:


.github/actions/optimized_build/action.yaml
```diff
@@ -29,17 +29,20 @@ runs:
     - name: "Check for WASM binaries larger than allowed limit"
       shell: "sh"
       run: |
-        find_files_cmd="find \"artifacts/\" -size +1500k | wc -l"
+        large_files=`find "artifacts/" -size +3M`
 
-        large_files_count=$(eval "${find_files_cmd}")
+        large_files_count=`wc -l < EOF
+          ${large_files}
+          EOF`
```
gmanev7, line 36: what about
large_files=`find "artifacts/" -size +3M | wc -l`

gmanev7: `` is the same as `eval`

KirilMihaylov: I want to collect the file names too, which is half the point of this PR.

gmanev7 (approved):

PR #150, opened 2023-08-03
ci: Fix file size limit check

KirilMihaylov:


.github/actions/optimized_build/action.yaml
```diff
@@ -29,19 +29,14 @@ runs:
...
-        # Temporarily disable check
-        exit 0
+        large_files="`find "artifacts/" -size +3M | sed 's/ /\n/g'`"
 
-        large_files=`find "artifacts/" -size +3M`
-
-        large_files_count=`wc -l < EOF
-          ${large_files}
-          EOF`
+        large_files_count="`echo -n "${large_files}" | wc -l`"
```
gmanev7, line 34: have you tested with `find "artifacts/" -size +3M | wc -w` ?

.github/actions/optimized_build/action.yaml
```diff
@@ -29,19 +29,14 @@ runs:
...
 
         if ! test "${large_files_count}" = "0";
           then (
             (
-              echo "There are ${large_files_count} files with size larger than allowed limit!"
+              echo "There are files with size larger than allowed limit!"
```
gmanev7, line 39: why don't we print the number?

.github/workflows/smart_contracts.yaml
```diff
@@ -32,24 +37,51 @@ env:
...
+      - if: "github.event_name == 'workflow_dispatch' && fromJSON(github.event.inputs.optimized_build)"
+        shell: "sh"
+        run: 'echo "## Building optimized binaries" >> "${GITHUB_STEP_SUMMARY}"'
   check_formatting:
     runs-on: "ubuntu-latest"
+    needs:
+      - "preparation"
     steps:
       - uses: "actions/checkout@v3"
+        if: "github.event_name != 'workflow_dispatch' || !fromJSON(github.event.inputs.skip_checks)"
```
gmanev7, line 58: a lot of repetition!!! please figure out a solution to avoid it! In addition, why it's part of this PR?

.github/actions/optimized_build/action.yaml
KirilMihaylov: Why would we? It doesn't bring any meaningful information. On the other hand, the two lines below it I am printing the meaningful one, the file names.

.github/actions/optimized_build/action.yaml
KirilMihaylov: I haven't, yet it's a bit prettier when they are on a new line each in the step report. It's a bit more clear to read.

The point of separating it was not only to not lose this information as that happened when we encountered the problem with the file size limits.

TL;DR: This is intentionally split into two evals.

KirilMihaylov:
Reverted last commit.

.github/workflows/build_optimized.yaml
```diff
@@ -0,0 +1,190 @@
...
+        if: "fromJSON(inputs.perform_checks)"
+  linting:
+    runs-on: "ubuntu-latest"
+    needs:
+      - "check_formatting"
+      - "check_codebase"
+      - "audit_dependencies"
+    steps:
+      - uses: "actions/checkout@v3"
+        if: "github.event_name != 'workflow_dispatch' || !fromJSON(github.event.inputs.skip_checks)"
```
gmanev7, line 70: it looks like the last refactoring did not reduce the repetitions of this line?

KirilMihaylov: It's still a bit of a mess and it's not ready quite yet.

gmanev7 (approved):

PR #152, opened 2023-08-08
feat: LPP utilization-based open/close mechanism

### LINKED ISSUE #128: [LPP] Open/Close LPP Deposits on a certain % utilization threshold
Close LPP new deposits when the utilisation level falls below a `%` (ex. 24%) threshold. Resume accepting LPP deposits when the ratio goes above the threshold;

KirilMihaylov:
Closes #128


contracts/lpp/src/lpp.rs
```diff
@@ -796,8 +858,46 @@ mod test {
         assert_eq!(withdraw, Coin::new(1110));
     }
 
-    fn coin_cw(amount: u128) -> CwCoin {
-        coin_legacy::to_cosmwasm::<TheCurrency>(amount.into())
```
gmanev7, line 800: this change triggers a lot of updated lines of code ... what's your motivation? If it was to accept Amount and Coin<TheCurrency> then define the arg as `Into<Coin<Currency>>`

scripts/deploy-contracts-genesis.sh
```diff
@@ -32,7 +32,7 @@ add_wasm_messages() {
   _add_wasm_message "$genesis_home_dir" "$wasm_code_path" "lease" "$((++id))" \
     "$ADMIN_CONTRACT_ADDRESS" "" "--instantiate-anyof-addresses $LEASER_ADDRESS"
 
-  local -r lpp_init_msg='{"lpn_ticker":"'"$lpp_native"'","lease_code_admin":"'"$LEASER_ADDRESS"'","borrow_rate":{"base_interest_rate":100,"utilization_optimal":750,"addon_optimal_interest_rate":20}}'
+  local -r lpp_init_msg='{"lpn_ticker":"'"$lpp_native"'","lease_code_admin":"'"$LEASER_ADDRESS"'","borrow_rate":{"base_interest_rate":100,"utilization_optimal":750,"addon_optimal_interest_rate":20},"min_utilization":"0"}'
```
gmanev7, line 35: why the new value is a string?

contracts/lpp/src/msg.rs
```diff
@@ -20,10 +20,13 @@ pub struct InstantiateMsg {
     pub lpn_ticker: SymbolOwned,
     pub lease_code_admin: Addr,
     pub borrow_rate: InterestRate,
+    pub min_utilization: Percent,
```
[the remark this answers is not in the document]
KirilMihaylov: `LimitedPercent<const UPPER_LIMIT: Units>`?

scripts/deploy-contracts-genesis.sh
KirilMihaylov: I think I was mislead by the CW variants of the integers which used strings instead.

contracts/lpp/src/state/config.rs
```diff
@@ -53,41 +59,50 @@ impl Config {
...
+        }
+    };
+}
+
+update_param_impl!(
+    update_lease_code(lease_code_id: Uint64),
+    update_borrow_rate(borrow_rate: InterestRate),
+    update_min_utilization(min_utilization: BoundToHundredPercent),
+);
+
```
gmanev7, line 97: we have discussed many times that macro-ses are not welcome where we may provide functions, please rework

contracts/lpp/src/contract/config.rs
gmanev7, line 1: no much value left of this module,  please consider inlining functions into the calling sites

packages/finance/src/percent.rs
```diff
@@ -167,6 +168,69 @@ impl<'a> Sub<&'a Percent> for Percent {
...
+
+#[derive(Debug, Error)]
+#[error(
+    "[Finance] [Percent] Upper bound has been crossed! Upper bound is: {bound}, but got: {value}!"
+)]
+pub struct UpperBoundCrossed {
+    bound: Units,
+    value: Units,
+}
+
```
gmanev7, line 223: please stick to error::Error

packages/finance/src/percent.rs
```diff
@@ -167,6 +168,69 @@ impl<'a> Sub<&'a Percent> for Percent {
...
+pub type BoundToHundredPercent = BoundPercent<{ Percent::HUNDRED.units() }>;
+
+impl BoundToHundredPercent {
+    pub const HUNDRED: Self = if let Ok(percent) = Self::try_from_percent(Percent::HUNDRED) {
+        percent
+    } else {
+        panic!()
+    };
+}
+
```
gmanev7, line 233: seems unused ... please remove

packages/finance/src/percent.rs
```diff
@@ -167,6 +168,69 @@ impl<'a> Sub<&'a Percent> for Percent {
     }
 }
 
+impl<const UPPER_BOUND: Units> From<BoundPercent<UPPER_BOUND>> for Percent {
+    fn from(value: BoundPercent<UPPER_BOUND>) -> Self {
+        value.percent()
+    }
+}
```
gmanev7, line 175: please move it down before the `impl TryFrom`

contracts/lpp/src/lpp.rs
```diff
@@ -679,14 +740,15 @@ mod test {
 
         grant_admin_access(deps.as_mut(), &admin);
         Config::new(
-            balance_mock.denom,
+            String::from(TheCurrency::BANK_SYMBOL),
```
gmanev7, line 743: [everywhere] please use `Into::into()` where possible. Specifying types as in this case clutter the code.

contracts/lpp/src/lpp.rs
```diff
@@ -796,8 +858,49 @@ mod test {
...
+            MockQuerier::new(&[(mock_env.contract.address.as_str(), &[coin_cw(LPP_BALANCE)])]);
+        let mock_querier: QuerierWrapper<'_> = QuerierWrapper::new(&mock_querier);
+
+        assert_eq!(
+            lpp.deposit_limit(&mock_querier, &mock_env).unwrap(),
+            Some(EXPECTED_DEPOSIT_LIMIT)
+        );
+    }
+
+    fn coin_cw(coin: impl Into<Coin<TheCurrency>>) -> CwCoin {
```
gmanev7, line 902: Kiril, please comply with all the practices we've established so far! Never ever we have used `impl T` at arg definition!

contracts/lpp/src/lpp.rs
```diff
@@ -796,8 +858,49 @@ mod test {
...
+    fn test_deposit_limit() {
+        const BORROWED: Coin<TheCurrency> = Coin::new(200);
+        const LPP_BALANCE: Coin<TheCurrency> = Coin::new(100);
+        const MIN_UTILIZATION: BoundToHundredPercent = if let Ok(percent) =
+            BoundToHundredPercent::try_from_percent(Percent::from_permille(500))
+        {
+            percent
+        } else {
+            panic!()
+        };
```
gmanev7, line 871: [everywhere] 7 lines just to have `const` - not acceptable

contracts/lpp/src/lpp.rs
```diff
@@ -796,8 +858,49 @@ mod test {
         assert_eq!(withdraw, Coin::new(1110));
     }
 
-    fn coin_cw(amount: u128) -> CwCoin {
-        coin_legacy::to_cosmwasm::<TheCurrency>(amount.into())
+    #[test]
+    fn test_deposit_limit() {
```
gmanev7, line 862: a single positive test is not enough at all! I've spotted a bug only reading the code. You should have caught it with a test!

contracts/lpp/src/state/config.rs
KirilMihaylov: While I do agree, we also have discussed that code duplication is unacceptable too, and that's a lot of it.

packages/finance/src/percent.rs
KirilMihaylov: Pretty sure it is used. The compiler personally made sure to complain at one point that it's missing.

gmanev7: successfully compiled with `RELEASE_VERSION=d NET_NAME=dev cargo test` without this chunk

gmanev7:
there is a failed test lpp_tests::config_update_parameters
Do you run tests before updating the PR?

packages/finance/src/percent.rs
KirilMihaylov: Ok, my bad in that case.

KirilMihaylov:
I'm sorry if I have left you with the impression I was ready. I just wanted to push my changes in case I have to work from a different machine. That is why I usually just notify you in DMs that it is ready.

contracts/lpp/src/lpp.rs
KirilMihaylov: With all due respect, I have a another test that does both positive and negative testing. It's in `lender.rs`.

KirilMihaylov:
It is ready now. :)

contracts/lpp/src/contract/lender.rs
```diff
@@ -17,18 +17,20 @@ use crate::{
...
-    LPN: 'static + Currency + DeserializeOwned + Serialize,
+    Lpn: 'static + Currency + DeserializeOwned + Serialize,
 {
     let lender_addr = info.sender;
     let amount = bank::received_one(info.funds)?;
 
-    let lpp = LiquidityPool::<LPN>::load(deps.storage)?;
+    let lpp = LiquidityPool::<Lpn>::load(deps.storage)?;
+
+    lpp.check_utilization_rate(&deps.querier, &env)?;
```
gmanev7, line 33: The decision of whether a deposit would accepted or not should be based on the rule of maintaining the minimum utilization limit. So this check is incorrect. Rather you should compare the amount to the remaining deposit capacity.

contracts/lpp/src/msg.rs
```diff
@@ -67,6 +75,7 @@ pub enum QueryMsg {
     },
     LppBalance(),
     Price(),
+    DepositLimit(),
```
gmanev7, line 78: [everywhere] Please rename to 'deposit capacity'

gmanev7 (commented):
only new findings. did not validate the previous

contracts/lpp/src/lpp.rs
```diff
@@ -78,15 +76,41 @@ where
...
 
-    pub fn total_lpn(&self, deps: &Deps<'_>, env: &Env) -> Result<Coin<LPN>> {
-        let res = self.balance(&env.contract.address, &deps.querier)?
-            + self.total.total_principal_due()
-            + self.total.total_interest_due_by_now(env.block.time);
+    pub fn deposit_capacity(
+        &self,
+        querier: &QuerierWrapper<'_>,
+        env: &Env,
+        pending_deposit: Option<Coin<Lpn>>,
```
gmanev7, line 83: Why do we complicate the logic? Do not think we need this argument ... we shall calculate how much deposit amount this Lpp can accept.

contracts/lpp/src/contract/lender.rs
```diff
@@ -17,18 +17,26 @@ use crate::{
...
     let lender_addr = info.sender;
     let amount = bank::received_one(info.funds)?;
 
-    let lpp = LiquidityPool::<LPN>::load(deps.storage)?;
+    let lpp = LiquidityPool::<Lpn>::load(deps.storage)?;
+
+    if lpp
+        .deposit_capacity(&deps.querier, &env, Some(amount))?
+        .map(|limit| amount > limit)
+        .unwrap_or_default()
```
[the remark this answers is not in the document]
KirilMihaylov: Not true, as I am checking whether it's above and return an error. The default of `bool` is `false`, thus never going to enter when the minimum is zero. Verified by tests in `lender.rs`.

contracts/lpp/src/lpp.rs
KirilMihaylov: By the time this code gets executed, the funds have already arrived. That is why I was checking whether it is **above** the limit **when they have arrived**, the previous implementation.

contracts/lpp/src/contract/lender.rs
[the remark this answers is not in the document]
gmanev7: right ... and that is the problem. Capacity None means endless, right? And you should accept a deposit of any amount.

Please point to the test with min_utilization = 0?

contracts/lpp/src/contract/lender.rs
```diff
@@ -251,7 +269,69 @@ mod test {
         let balance_nlpn = query_balance(deps.as_ref().storage, Addr::unchecked("lender2"))
             .unwrap()
             .balance;
-        assert_eq!(balance_nlpn.u128(), zero);
+        assert_eq!(balance_nlpn.u128(), 0);
+    }
+
+    #[test]
+    fn test_min_utilization() {
```
gmanev7, line 276: please provide short and single-purpose tests

contracts/lpp/src/contract/lender.rs
```diff
@@ -125,6 +144,7 @@ mod test {
     const BASE_INTEREST_RATE: Percent = Percent::from_permille(70);
     const UTILIZATION_OPTIMAL: Percent = Percent::from_permille(700);
     const ADDON_OPTIMAL_INTEREST_RATE: Percent = Percent::from_permille(20);
+    const DEFAULT_MIN_UTILIZATION: BoundToHundredPercent = BoundToHundredPercent::ZERO;
 
     #[test]
     fn test_deposit() {
```
gmanev7, line 150: [nice to have, optional for this PR] please provide short and single-purpose tests

contracts/lpp/src/contract/migrate.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use finance::percent::bound::BoundToHundredPercent;
+use sdk::{
+    cosmwasm_std::{Storage, Uint64},
+    cw_storage_plus::Item,
+};
+
+use crate::{borrow::InterestRate, error::Result as ContractResult, state::Config};
+
+#[derive(Serialize, Deserialize)]
```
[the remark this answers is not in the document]
KirilMihaylov: Will keep that in mind for the future.

KirilMihaylov:
> right ... and that is the problem. Capacity None means endless, right? And you should accept a deposit of any amount.
> 
> Please point to the test with min_utilization = 0?

`contracts/lpp/contract/lender.rs`:`+325`

contracts/lpp/src/contract/lender.rs
KirilMihaylov: Ok. I'll split it.

contracts/lpp/src/contract/lender.rs
KirilMihaylov: I'll split this one too.

gmanev7 (approved):

PR #153, opened 2023-08-10
feat(leaser): Added purging capability

### LINKED ISSUE #24: [Leaser] purge closed leases
Unless that info is required for the web app consider removing them. That would reduce the cost of lease code migrations and the storage amount of the chain.

Potential approaches:
- an unrestricted command to the leaser going over all leases of all customers and removing closed ones from the internal leaser registry
- a command to the leaser removing closed leases pertaining *only* to the customer who sends the command

KirilMihaylov:
Closes #24

### DIFF
--- a/contracts/lease/src/api/query.rs
+++ b/contracts/lease/src/api/query.rs
@@ -12,7 +12,7 @@ use super::{DownpaymentCoin, LeaseCoin, LpnCoin};
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
 pub enum QueryMsg {
     State {},
-    IsClosed {},
+    Finished {},
 }
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
--- a/contracts/lease/src/api/query.rs
+++ b/contracts/lease/src/api/query.rs
@@ -10,7 +10,10 @@ use super::{DownpaymentCoin, LeaseCoin, LpnCoin};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
-pub struct StateQuery {}
+pub enum QueryMsg {
+    State {},
+    IsClosed {},
+}
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, JsonSchema)]
 #[cfg_attr(any(test, feature = "testing"), derive(Clone, Debug))]
--- a/contracts/lease/src/api/query.rs
+++ b/contracts/lease/src/api/query.rs
@@ -9,7 +9,7 @@ use sdk::{
 use super::{DownpaymentCoin, LeaseCoin, LpnCoin};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
-#[serde(rename_all = "snake_case")]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
 pub enum QueryMsg {
     State {},
     IsClosed {},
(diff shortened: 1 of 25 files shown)


contracts/lease/src/api/query.rs
```diff
@@ -10,7 +10,10 @@ use super::{DownpaymentCoin, LeaseCoin, LpnCoin};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
-pub struct StateQuery {}
+pub enum QueryMsg {
+    State {},
+    IsClosed {},
```
gmanev7, line 15: for the sake of avoiding naming confusion let's rename this to `Finished`

contracts/lease/src/contract/endpoins.rs
```diff
@@ -125,10 +125,18 @@ pub fn sudo(deps: DepsMut<'_>, env: Env, msg: SudoMsg) -> ContractResult<CwRespo
...
-        .and_then(|state| state.state(env.block.time, &deps.querier))
-        .and_then(|resp| to_binary(&resp).map_err(Into::into))
+        .and_then(|state| match msg {
+            QueryMsg::State {} => state
+                .state(env.block.time, &deps.querier)
+                .and_then(|resp| to_binary(&resp).map_err(Into::into)),
+            QueryMsg::IsClosed {} => to_binary(&matches!(
+                state,
+                State::Closed { .. } | State::Liquidated { .. },
+            ))
```
gmanev7, line 137: The knowledge of which lease states are considered final should be brought into the states themselves. Otherwise, if we add another state we must keep in mind updating this list too. So, please introduce the corresponding trait method with **no default implementation**.

contracts/leaser/src/msg.rs
```diff
@@ -55,6 +55,10 @@ pub enum ExecuteMsg {
     /// It migrates the next batch of up to `max_leases` number of Lease instances
     /// and emits the status as specified in `MigrateLeases`.
     MigrateLeasesCont { key: Addr, max_leases: MaxLeases },
+    DetachClosed {
```
gmanev7, line 58: 1. please add a documentation 
2. rename to `PurgeLeases`

contracts/leaser/src/msg.rs
```diff
@@ -55,6 +55,10 @@ pub enum ExecuteMsg {
...
+        max_leases: MaxLeases,
+        next_key: Option<Addr>,
```
gmanev7, line 60: let's keep the consistency with the above variants of the argument definition:
- flip the order
- `next_key` -> ``key`

gmanev7 (commented):
to be continued

contracts/lease/src/api/query.rs
```diff
@@ -10,7 +10,10 @@ use super::{DownpaymentCoin, LeaseCoin, LpnCoin};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
 #[serde(rename_all = "snake_case")]
-pub struct StateQuery {}
+pub enum QueryMsg {
+    State {},
```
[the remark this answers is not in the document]
KirilMihaylov: I can always rename the field through `serde`. I think that might be even more elegant solution, if we don't want breaking changes.

contracts/leaser/src/msg.rs
KirilMihaylov: Applying the change, but I'd like to remind you that to the JSON the order won't matter and that the optional field will be first. Just doesn't seem natural.

contracts/lease/src/api/query.rs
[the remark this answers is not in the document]
gmanev7: I was also about to recommend this ... however, it seems reasonable to migrate it to `State`

contracts/leaser/src/msg.rs
gmanev7: exactly, the format would not be an obstacle ... the goal is to reach better consistency among the variants on the API level

contracts/lease/src/api/query.rs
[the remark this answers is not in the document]
KirilMihaylov: I have bad news though. The previous query's definition was actually just "object" without specifying any fields, thus there is no way to avoid the breaking change.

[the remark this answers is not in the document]
KirilMihaylov: Update: Left is as state.

PR #154, opened 2023-08-11
refactor(lease): Set start time to contract opening one

### LINKED ISSUE #113: [Lease] Start a lease at the moment its opening begins
Currently, the initial period of a lease starts at the time the opening finishes. That results in extra interest for the loan taken 30-60 secs before. Or the opposite, we lose the margin interest for the time since starting the lease open procedure.

The solution is to start the initial period of a lease at the time the opening starts, which is equal to the time the loan is taken. In particular, that requires that time be carried through the opening process till its end when we start officially the lease.

KirilMihaylov:
Closes #113

### DIFF
--- a/contracts/lease/src/contract/state/opening/buy_asset.rs
+++ b/contracts/lease/src/contract/state/opening/buy_asset.rs
@@ -45,14 +45,14 @@ pub(in crate::contract::state::opening) fn start(
     downpayment: DownpaymentCoin,
     loan: OpenLoanRespResult,
     deps: (LppRef, OracleRef, TimeAlarmsRef),
-    opening_time: Timestamp,
+    start_opening_at: Timestamp,
 ) -> StartState {
     dex::start_local_remote::<_, BuyAsset>(OpenIcaAccount::new(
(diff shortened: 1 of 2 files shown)


contracts/lease/src/contract/state/opening/buy_asset.rs
```diff
@@ -45,8 +45,15 @@ pub(in crate::contract::state::opening) fn start(
...
     deps: (LppRef, OracleRef, TimeAlarmsRef),
+    opening_time: Timestamp,
 ) -> StartState {
-    dex::start_local_remote::<_, BuyAsset>(OpenIcaAccount::new(new_lease, downpayment, loan, deps))
+    dex::start_local_remote::<_, BuyAsset>(OpenIcaAccount::new(
+        new_lease,
+        downpayment,
+        loan,
+        deps,
+        opening_time,
```
gmanev7, line 55: [everywhere] `opening_time` -> `start_opening_at`

gmanev7 (approved):

PR #161, opened 2023-08-17
ci: Parallelize `check_features` job

KirilMihaylov:


.github/workflows/checks.yaml
```diff
@@ -41,11 +41,7 @@ jobs:
         shell: "sh"
         run: "./lint.sh"
   check_features:
-    runs-on: "ubuntu-latest"
-    steps:
-      - uses: "actions/checkout@v3"
-      - uses: "nolus-protocol/rust-ci-actions/rust_cache@main"
-      - uses: "nolus-protocol/rust-ci-actions/check_features@main"
```
gmanev7, line 48: don't you have to cleanup 'nolus-protocol/rust-ci-actions/check_features' then?

gmanev7 (approved):

.github/workflows/checks.yaml
KirilMihaylov: It should be updated, yes, but not removed as it's used within `check_features.yaml`, which is in turn called by `checks.yaml` /this file/.

PR #164, opened 2023-08-28
feat: Add `deny_unknown_fields` attribute to public interfaces

KirilMihaylov:


gmanev7 (commented):
are we sure we have not changed the representation of the existing APIs adding `rename_all = "snake_case"`?

KirilMihaylov:
Yes, because Rust already forces us to use snake case through lints. This was just a safeguard that we put and I wanted to "continue the tradition" so to say.

gmanev7 (approved):

PR #165, opened 2023-08-30
chore: Removed unnecessary Serialize

maneva3:


contracts/lease/src/lease/with_lease_deps.rs
```diff
@@ -72,7 +72,7 @@ where
 
     fn on<C>(self) -> AnyVisitorResult<Self>
     where
-        C: 'static + Currency + Serialize + DeserializeOwned,
+        C: 'static + Currency + DeserializeOwned,
```
gmanev7, line 75: I'm curious to see if there are `Deserialize` and `DeserializeOwner` extra trait boundaries.

gmanev7 (approved):

PR #167, opened 2023-08-31
refactor: `currency` package deduplication

KirilMihaylov:


KirilMihaylov:
* I believe that it's rather macro use, not abuse, as I only use it once in it's local file for something that otherwise wouldn't be de-duplicatable.
* I stand behind the general naming as those are interfaces behind a specific module names. Rust's idiomatic guidelines were influenced by OCaml on that.
* I also stand behind the added complexity as I believe this is one of the cases where Rust's strengths shines. "Write once and forget about it for the next 10 years."
* The "third copy", as I mentioned in DMs, is the addition of resolution by dex symbol. I would kindly ask that you take a look at `currency/src/lease.rs` between lines `-350:-398`.
As mentioned in DMs, I believe this already wasn't meeting the quality standards as it is very error-prone, especially when changes are needed.

I believe I answered to all your concerns listed above, but feel free to correct me, if I missed something.

PR #173, opened 2023-09-26
.price_alarm() as member of Level

maneva3:


platform/packages/finance/src/liability/level.rs
```diff
@@ -24,6 +32,24 @@ impl Level {
             Self::Max(_) => 4,
         }
     }
+
+    pub fn price_alarm<Asset, Lpn>(
```
gmanev7, line 36: Just realized this change we would introduce a dependency to higher-level abstractions, for example, Lpn, Asset, 'alarm'. In addition, I think the position is the best host of if in the form of `fn price_at(&self, level: Level, total_due: Coin<Lpn>)`

PR #175, opened 2023-09-28
Correct usage of min_asset and min_sell_asset

maneva3:


packages/finance/src/liability/mod.rs
```diff
@@ -882,13 +884,14 @@ mod test_check {
         let max_ltv = Percent::from_permille(751);
         let spec = liability_with_max(max_ltv);
         let min_asset = Coin::<TestCurrency>::new(1000);
+        let min_sell_asset = Coin::<TestCurrency>::new(1000);
```
gmanev7, line 887: adjust this value to simulate more test case variants

gmanev7 (approved):

PR #177, opened 2023-10-02
Improve tests for check_liability

maneva3:


contracts/lease/src/position/mod.rs
```diff
@@ -238,78 +239,50 @@ mod test_check {
...
-    const MIN_ASSET: Coin<TestLpn> = Coin::new(100);
-    const MIN_SELL_ASSET: Coin<TestLpn> = Coin::new(1);
     const PRICE_TEST_LPN: Coin<TestLpn> = Coin::new(1_000);
     const PRICE_TEST_CURRENCY: Coin<TestCurrency> = Coin::new(1_000);
 
     #[test]
     fn no_debt() {
         let warn_ltv = Percent::from_permille(11);
-        let spec = liability_with_first(warn_ltv, 100.into(), MIN_ASSET, MIN_SELL_ASSET);
+        let spec = liability_with_first(warn_ltv, 100.into(), 1.into(), 1.into());
```
gmanev7, line 248: please keep the fn names changed as their purpose does. `liability_with_*` no more creates Liability instances

gmanev7: do not remember that one should read the code as a chunk of prose

contracts/lease/src/position/mod.rs
```diff
@@ -509,91 +464,69 @@ mod test_check {
...
                     healthy_ltv: STEP
                 }
             ),
         );
+    }
+
+    #[test]
+    fn liquidate_partial() {
+        let max_ltv = Percent::from_permille(881);
+        let spec = liability_with_max(max_ltv, LEASE_AMOUNT, 100.into(), 1.into());
```
gmanev7, line 508: 100 -> 120, and add another assert when (asset - liquidate_amount) < 120

gmanev7 (approved):

PR #182, opened 2023-10-11
Add validate open amount check

maneva3:


contracts/lease/src/api/open/mod.rs
```diff
@@ -153,6 +153,16 @@ impl PositionSpec {
         obj
     }
 
+    pub fn validate_open_amount(self, amount: LpnCoin) -> ContractResult<()> {
```
gmanev7, line 156: 1. Why `self`?
2. The fn name implies validation of the position amount and that is not true as per the implementation

contracts/lease/src/api/open/mod.rs
```diff
@@ -153,6 +153,16 @@ impl PositionSpec {
...
+        if amount.amount() < self.min_sell_asset.amount() {
```
gmanev7, line 157: do not do this
do we know they are of the same currency? How do we compare then?

PR #190, opened 2023-10-19
Refactor versioning

KirilMihaylov:


platform/packages/versioning/src/lib.rs
```diff
@@ -21,6 +21,49 @@ pub struct SemVer {
...
+                b'.' => {
+                    segment_index += 1;
+
+                    if segment_index == segments.len() {
+                        panic!("Unexpected segment!");
+                    } else if version_index + 1 == version.len() {
+                        panic!("Version can't end with a dot!");
+                    }
+                }
+                _ => unreachable!(),
```
gmanev7, line 47: a description would provide valuable explanation of a failed parsing

platform/packages/versioning/src/lib.rs
```diff
@@ -21,6 +21,49 @@ pub struct SemVer {
...
+            }
+
+            version_index += 1;
+        }
+
+        if segment_index < segments.len() - 1 {
+            unreachable!()
```
gmanev7, line 54: a description would provide valuable explanation of a failed parsing

platform/packages/versioning/src/lib.rs
```diff
@@ -21,6 +21,49 @@ pub struct SemVer {
...
+                    segments[segment_index] += (digit - b'0') as VersionSegment;
+                }
+                b'.' => {
+                    segment_index += 1;
+
+                    if segment_index == segments.len() {
+                        panic!("Unexpected segment!");
+                    } else if version_index + 1 == version.len() {
+                        panic!("Version can't end with a dot!");
+                    }
```
gmanev7, line 45: consider replacing that with `assert`-s

platform/packages/versioning/src/lib.rs
```diff
@@ -34,47 +77,10 @@ impl Version {
     }
 }
 
-pub fn parse_semver(version: &str) -> SemVer {
```
gmanev7, line 37: why did you abandon this implementation? It was fine ... very fine ... much more readable and less error-prone than the new code!

protocol/contracts/lease/src/contract/endpoins.rs
```diff
@@ -22,6 +22,8 @@ use super::{finalize::FinalizerRef, state::Migrate};
 #[cfg(feature = "migration")]
 const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 5;
 const CONTRACT_STORAGE_VERSION: VersionSegment = 6;
+const CONTRACT_VERSION: SemVer = package_version!();
```
gmanev7, line 25: please rename EVERYWHERE `CONTRACT_VERSION` -> `PACKAGE_VERSION`

protocol/contracts/lease/src/contract/endpoins.rs
```diff
@@ -22,6 +22,8 @@ use super::{finalize::FinalizerRef, state::Migrate};
...
+const COMPLETE_VERSION: Version = version!(CONTRACT_STORAGE_VERSION, CONTRACT_VERSION);
```
gmanev7, line 26: please rename EVERYWHERE `COMPLETE_VERSION` -> `CONTRACT_VERSION`

platform/packages/versioning/src/lib.rs
KirilMihaylov: It isn't const-evaluatable; The reasoning behind the PR is that this should be done during compilation instead of parsing the version on-demand every time.

KirilMihaylov: There are plenty of tests that make sure it does in fact parse it correctly.

KirilMihaylov: Added more tests, specifically for checking every branch of the parsing.

platform/packages/platform/src/response.rs
```diff
@@ -8,6 +8,14 @@ pub fn empty_response() -> CwResponse {
     response_only_messages(MessageResponse::default())
 }
 
+pub fn response_consuming<T, E>(response: T) -> Result<CwResponse, E>
```
gmanev7, line 11: why do we need another variant of `fn response`? How have you decided which one to use? It looks like each of them alone is enough.

protocol/contracts/lpp/src/contract/mod.rs
```diff
@@ -29,6 +29,8 @@ mod rewards;
 #[cfg(feature = "migration")]
 const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 0;
 const CONTRACT_STORAGE_VERSION: VersionSegment = 1;
+const CONTRACT_VERSION: SemVer = package_version!();
+const COMPLETE_VERSION: Version = version!(CONTRACT_STORAGE_VERSION, CONTRACT_VERSION);
```
gmanev7, line 33: here the two issues are still present .... `lpp` was not in the find/replace scope? ;-)

KirilMihaylov: Somehow it wasn't ;) Will correct that for the next time.

platform/packages/platform/src/response.rs
KirilMihaylov: `origin/main` was broken because `response` expected a reference but `ReleaseLabel` was not behind one. The motivation behind this was to avoid having to write it as a closure just to make it a reference everywhere.

gmanev7: disagree it was "broken"
two versions of `fn response` are too much, please choose and stick to one.

KirilMihaylov: I don't know what do you disagree with. Please check the signatures.

I wanted to avoid repetition. If you think using lambdas/closures would be better, I will change it.

gmanev7 (approved):

PR #195, opened 2023-11-01
Validate downpayment and borrow amount

maneva3:


contracts/lease/src/position/spec.rs
```diff
@@ -66,7 +69,26 @@ where
...
+        if asset_amount < self.min_asset {
+            Err(ContractError::InsufficientAssetAmount(
+                self.min_asset.into(),
+            ))
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn check_asset_amount_on_liq<Trasactional>(
```
gmanev7, line 91: reuse `check_asset` please

contracts/lease/src/position/spec.rs
```diff
@@ -87,6 +109,38 @@ where
...
+        &self,
+        downpayment: Coin<Lpn>,
+        may_max_ltd: Option<Percent>,
+    ) -> ContractResult<Coin<Lpn>> {
+        let price: Price<Lpn, Lpn> = price::total_of(1.into()).is(1.into());
+
+        let _ = self
+            .check_trasaction_amount(downpayment, price)
+            .map_err(|err| match err {
+                ContractError::PositionCloseAmountTooSmall(min_amount) => {
```
gmanev7, line 127: it is not OK `check_transaction_amount` to return an error related to PolitionClose, is it?

contracts/lease/src/position/spec.rs
```diff
@@ -87,6 +109,38 @@ where
...
+    /// Return 'error::ContractError::InsufficientTrasactionAmount' when either the downpayment
+    /// or the borrow amount is with amount less than the minimum trasaction amount.
+    /// Return 'error::ContractError::InsufficientAssetAmount' when the lease (downpayment + borrow)
+    /// is with amount less than the minimum asset amount.
+    pub fn calc_borrow_amount(
+        &self,
+        downpayment: Coin<Lpn>,
+        may_max_ltd: Option<Percent>,
+    ) -> ContractResult<Coin<Lpn>> {
+        let price: Price<Lpn, Lpn> = price::total_of(1.into()).is(1.into());
```
gmanev7, line 122: 1. variable naming do not convey any meaning
2. the value uses constants .... which is not what the good practices recommend. Suggest using Price::identity()

protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -76,6 +79,26 @@ where
         }
     }
 
+    /// Calculate the borrow amount.
+    /// Return 'error::ContractError::InsufficientTrasactionAmount' when either the downpayment
+    /// or the borrow amount is with amount less than the minimum trasaction amount.
+    /// Return 'error::ContractError::InsufficientAssetAmount' when the lease (downpayment + borrow)
+    /// is with amount less than the minimum asset amount.
```
gmanev7, line 86: [nit] similar to the above

protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -76,6 +79,26 @@ where
...
+    pub fn calc_borrow_amount(
+        &self,
+        downpayment: Coin<Lpn>,
+        may_max_ltd: Option<Percent>,
+    ) -> ContractResult<Coin<Lpn>> {
+        let amount_in_lpn: Price<Lpn, Lpn> = Price::identity();
```
gmanev7, line 92: 1. let's not specify the type when the type inference manages to do it
2. the variable name is not correct, that is not an amount

protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -76,6 +79,26 @@ where
...
+
+        self.check_trasaction_amount(downpayment, amount_in_lpn)?;
+
+        let borrow = self.liability.init_borrow_amount(downpayment, may_max_ltd);
+        self.check_trasaction_amount(borrow, amount_in_lpn)
+            .and_then(|_| self.check_asset_amount(downpayment.add(borrow), amount_in_lpn))
+            .map(|_| borrow)
```
gmanev7, line 99: better to specify `()` as the closure arg

gmanev7, line 99: please combine it into a single expression

protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -91,3 +114,85 @@ where
...
+        assert_eq!(Coin::<TestLpn>::new(451), borrow_2.unwrap());
+
+        let borrow_3 = spec.calc_borrow_amount(870.into(), Some(Percent::from_percent(100)));
+        assert_eq!(Coin::<TestLpn>::new(870), borrow_3.unwrap());
+
+        let borrow_4 = spec.calc_borrow_amount(650.into(), Some(Percent::from_percent(150)));
+        assert_eq!(Coin::<TestLpn>::new(975), borrow_4.unwrap());
+    }
+
+    fn spec<Lpn>(min_asset: Lpn, min_trasaction_amount: Lpn) -> Spec<TestLpn>
```
gmanev7, line 183: with `Lpn` we name a currency, so not suitable here

protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -91,3 +114,85 @@ where
...
+
+        assert!(matches!(
+            borrow,
+            Err(ContractError::InsufficientTrasactionAmount(_))
+        ));
+    }
+
+    #[test]
+    fn borrow_less_than_min() {
+        let spec = spec(1_000, 100);
```
gmanev7, line 142: init here with the corner case value

protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -91,3 +114,85 @@ where
...
+            borrow_1,
+            Err(ContractError::InsufficientAssetAmount(_))
+        ));
+
+        let borrow_2 = spec.calc_borrow_amount(550.into(), Some(Percent::from_percent(81)));
+        assert!(matches!(
+            borrow_2,
+            Err(ContractError::InsufficientAssetAmount(_))
+        ));
+    }
```
gmanev7, line 165: please add a positive invocation using the same `spec` instance. We aim at catching the corner case values.

gmanev7 (approved):

PR #213, opened 2023-11-24
refactor: Select protocol using features

KirilMihaylov:


protocol/packages/currencies/src/lease/mod.rs
```diff
@@ -1,14 +1,17 @@
...
+#[cfg(feature = "astroport")]
 pub(crate) mod astroport;
+#[cfg(feature = "astroport")]
+use self::astroport as impl_mod;
 
-#[cfg(all(not(dex = "astroport"), dex = "osmosis"))]
+#[cfg(any(feature = "osmosis", feature = "testing"))]
 pub(crate) mod osmosis;
+#[cfg(any(feature = "osmosis", feature = "testing"))]
+use self::osmosis as impl_mod;
```
gmanev7, line 14: do not think we need `feature = "testing"`. The testing feature should just add `derive(Debug)`. The consumers of this package should ask for `--features osmosis, testing` or `--features astroport, testing`

protocol/packages/currencies/src/lease/mod.rs
```diff
@@ -18,21 +21,11 @@ pub struct LeaseGroup {}
...
-    #[cfg(dex = "astroport")]
-    fn maybe_visit<M, V>(matcher: &M, symbol: &SymbolSlice, visitor: V) -> MaybeAnyVisitResult<V>
-    where
-        M: Matcher + ?Sized,
-        V: AnyVisitor,
-    {
-        astroport::maybe_visit(matcher, symbol, visitor)
-    }
-
-    #[cfg(all(not(dex = "astroport"), dex = "osmosis"))]
```
gmanev7, line 30: well done!

protocol/packages/currencies/src/test.rs
gmanev7, line 1: we do need dex-specific currencies to be exposed as dex-neutral names, so please revert and add respective `mod astroport`

protocol/packages/marketprice/Cargo.toml
```diff
@@ -6,9 +6,12 @@ edition.workspace = true
 license.workspace = true
 
 [features]
+astroport = ["dep:currencies", "currencies/astroport"]
+osmosis = ["dep:currencies", "currencies/osmosis"]
```
gmanev7, line 10: `"dep:currencies"` looks unnecessary

KirilMihaylov: I added it just to hide `currencies` from the features list. I can remove it if you don't mind it being exposed there.

protocol/contracts/lease/src/contract/state/dex.rs
```diff
@@ -26,22 +26,15 @@ impl<H> State<H> {
...
-}
+// #[cfg(feature = "migration")]
+// impl<H> State<H> {
+//     pub fn map<MapFn, HTo>(self, map_fn: MapFn) -> State<HTo>
+//     where
+//         MapFn: FnOnce(H) -> HTo,
+//     {
+//         State::new(map_fn(self.handler))
+//     }
+// }
```
KirilMihaylov, line 37: This one was one of your commits, if I am not wrong. Do we really want to keep it commented? AFAIR, that was considered a code smell.

tests/src/common/oracle.rs
```diff
@@ -57,9 +57,11 @@ impl Instantiator {
...
             swap_tree: oracle::swap_tree!(
                 { base: StableC1::TICKER },
-                (1, PaymentC5::TICKER),
-                (3, PaymentC6::TICKER),
-                (13, PaymentC3::TICKER)
+                (1, LeaseC2::TICKER),
+                (3, LeaseC3::TICKER),
+                (7, LeaseC4::TICKER),
+                // (11, NativeC::TICKER), # FIX ME #216
+                (13, LeaseC1::TICKER),
```
gmanev7, line 64: in the oracle, we do deal with payment currencies. Is there any reason I cannot see for this change?

protocol/packages/swap/src/trx/mod.rs
```diff
@@ -3,8 +3,10 @@ use finance::coin::{Amount, CoinDTO};
 use platform::{ica::HostAccount, trx::Transaction};
 use sdk::cosmos_sdk_proto::cosmos::base::abci::v1beta1::MsgData;
 
-#[cfg(feature = "testing")]
-pub use self::test::*;
```
gmanev7: do not see `mod test`???

gmanev7: it's important the tests to see the request and response types not in `test` module

protocol/packages/marketprice/Cargo.toml
gmanev7: probably do not get you right 
> I added it just to hide currencies from the features list
I believe it's not necessary to have a separate feature. It's mandatory to have `currencies` in the `[dependencies]` though.

protocol/contracts/lease/src/contract/state/dex.rs
gmanev7: right, it is. I did it with the intention to uncomment right in the next commit. It's under review ... and will go in the repo soon.

Thanks for catching it

tests/src/common/oracle.rs
KirilMihaylov: I have renamed them in order to be able to distinguish Lease currencies.

protocol/packages/marketprice/Cargo.toml
KirilMihaylov: If an optional dependency is not mentioned by a feature, then a new feature with the name of the dependency is automatically created. By doing this, I am trying to avoid creating that feature as it is not functional by any means. I haven't tested whether I can only go with `currencies/protocol`, so I will try it out and report back.

protocol/packages/swap/src/trx/mod.rs
KirilMihaylov: Why not in there, when we explicitly show them on `testing`? If we do that, then `test` module should go or be private and only contain tests.

protocol/packages/marketprice/Cargo.toml
gmanev7: I am pretty sure we can go on;y with `currencies/protocol`

KirilMihaylov: No, unless there is `dep:` it creates it. What should I do in this case?

gmanev7: image

gmanev7: so no sense of specifying `dep:currencies` in our case

gmanev7: > No, unless there is `dep:` it creates it. What should I do in this case?

just leave `osmosis = [currencies/osmosis]` and the corresponding one for astroport

protocol/contracts/oracle/Cargo.toml
```diff
@@ -11,9 +11,11 @@ license.workspace = true
 crate-type = ["cdylib", "rlib"]
 
 [features]
-default = ["cosmwasm-bindings"]
+default = []
```
gmanev7, line 14: why have you decided to remove the default features for some contracts while still keeping them for the others?

gmanev7 (approved):

PR #215, opened 2023-11-27
refactor(admin): Decode response of `instatiate2`

KirilMihaylov:


platform/packages/platform/src/reply.rs
```diff
@@ -16,36 +21,43 @@ pub fn from_instantiate<T>(api: &dyn Api, reply: Reply) -> StdResult<Instantiate
...
-            })?)?,
+        address: api.addr_validate(&response.address)?,
         data: maybe_from_json(response.data)?,
     })
 }
 
+pub fn from_instantiate2_raw(
+    api: &dyn Api,
+    reply: Reply,
+) -> StdResult<InstantiateResponse<Vec<u8>>> {
```
gmanev7, line 47: The level of abstraction of this package is in contrast with the raw bytes that escape with the response. We need only the contract address, right? We do not have to interpret the `result.data` in any way. Remember that the reply may let the submessage result [go up](https://docs.cosmwasm.com/docs/smart-contracts/contract-semantics#handling-the-reply). Would suggest skipping the propagation of the data up. Is there a type the serde deserializes from any input?

platform/contracts/admin/src/lib.rs
```diff
@@ -188,42 +194,37 @@ fn migration_reply(msg: Reply, expected_release: String) -> ContractResult<CwRes
...
+    if reported_code_id != expected_code_id {
+        return Err(ContractError::DifferentInstantiatedCodeId {
             reported: reported_code_id,
             expected: expected_code_id,
-        })
+        });
+    }
+
+    if let Some(data) = data {
+        response::response(data)
```
gmanev7, line 225: see the comment below on the need to override the data in a reply

platform/packages/platform/src/reply.rs
KirilMihaylov: `serde_json`'s `Value` but we can't use that.

platform/contracts/admin/src/error.rs
```diff
@@ -17,11 +17,9 @@ pub enum Error {
...
     #[error("Contract returned wrong release string! \"{reported}\" was returned, but \"{expected}\" was expected!")]
     WrongRelease { reported: String, expected: String },
-    #[error(
-        "Contract might have been instantiated properly but contract address couldn't be found!"
-    )]
-    FindContractAddress {},
-    #[error("Contract address exists but code id is different! \"{reported}\" was returned, but \"{expected}\" was expected!")]
+    #[error("Contract returned wrong code id! Expected \"{expected}\", but got \"{reported}\"!")]
+    DifferentInstantiatedAddress { reported: Addr, expected: Addr },
+    #[error("Contract returned wrong code id! Expected \"{expected}\", but got \"{reported}\"!")]
```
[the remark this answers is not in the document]
KirilMihaylov: Must have missed it while editing the types and managing imports. Will fix it right away.

platform/Cargo.toml
```diff
@@ -43,6 +43,8 @@ cw-multi-test = "0.16.1"
 cosmos-sdk-proto = { version = "0.18", default-features = false }
 
 # Other
+serde-json-wasm = { version = "1", default-features = false }
+
```
gmanev7, line 47: I missed adding a comment on this before, why do you go directly with `serde`? We use `sdk::cosmwasm_std::from_json` and `sdk::cosmwasm_std::to_json_*`.

platform/contracts/admin/src/error.rs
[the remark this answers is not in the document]
gmanev7: hmmm. ... remain the same

platform/packages/platform/src/reply.rs
```diff
@@ -3,59 +3,83 @@
...
-        #[prost(bytes, tag = "2")]
-        pub data: Vec<u8>,
+impl InstantiateResponse<Vec<u8>> {
+    fn into_addr(self) -> Addr {
+        self.address
     }
+}
 
-    let response = decode::<ReplyData>(reply)?;
+pub fn from_instantiate_addr_only(api: &dyn Api, reply: Reply) -> StdResult<Addr> {
```
gmanev7, line 26: [all public functions] we shall return `platform::error::Error` not the cosmwasm one

platform/packages/platform/src/reply.rs
```diff
@@ -3,59 +3,83 @@
...
 }
 
 fn decode_raw<M>(message: &[u8]) -> StdResult<M>
 where
     M: Message + Default,
 {
-    M::decode(message).map_err(|_| {
-        StdError::generic_err("Data is malformed or doesn't comply with used protobuf format!")
+    M::decode(message).map_err(|error| {
+        StdError::generic_err(format!("[Platform] Data is malformed or doesn't comply with used protobuf format! Cause: [Protobuf] {error}"))
```
gmanev7, line 82: Platform != cosmwasm

platform/packages/platform/src/reply.rs
```diff
@@ -3,59 +3,83 @@
...
-        .map(|data| data.data)
-        .and_then(maybe_from_json)
+impl From<MsgInstantiateContract2Response> for UncheckedInstantiateResponse {
+    fn from(
+        MsgInstantiateContract2Response { address, data }: MsgInstantiateContract2Response,
+    ) -> Self {
+        Self { address, data }
+    }
+}
+
```
gmanev7, line 66: Added extra complexity handling `data`.  A simple trait a-la `Address` implemented for the two structures MsgInstantiateContractResponse and MsgInstantiate2ContractResponse would suffice

platform/packages/platform/src/reply.rs
KirilMihaylov: While true, that is an error that DOES within `platform`.

platform/packages/platform/src/reply.rs
```diff
@@ -23,78 +25,85 @@ impl InstantiateResponse<Vec<u8>> {
...
-pub fn from_execute<T>(reply: Reply) -> StdResult<Option<T>>
+pub fn from_execute<T>(reply: Reply) -> Result<Option<T>>
 where
     T: DeserializeOwned,
 {
     decode::<MsgExecuteContractResponse>(reply)
         .map(|data| data.data)
-        .map(Binary)
-        .and_then(from_json)
+        .and_then(|data| from_json(Binary(data)).map_err(From::from))
```
gmanev7, line 44: no need to instantiate `Binary`, just use `&data`

platform/packages/platform/src/reply.rs
```diff
@@ -23,78 +25,85 @@ impl InstantiateResponse<Vec<u8>> {
...
 
-fn decode_raw<M>(message: &[u8]) -> StdResult<M>
+fn decode_raw<M>(message: &[u8]) -> Result<M>
 where
     M: Message + Default,
 {
-    M::decode(message).map_err(|error| {
-        StdError::generic_err(format!("[Platform] Data is malformed or doesn't comply with used protobuf format! Cause: [Protobuf] {error}"))
-    })
+    M::decode(message).map_err(From::from)
```
gmanev7, line 94: Is the usage of `From` instead of `Into` intentional? As far as I remember the books say "define From, use Into"

platform/packages/platform/src/reply.rs
gmanev7: Spawning task is a nice instrument for tracking issues in the code that we have noticed. Here, as in the other cases before, we *introduce* issues.

platform/packages/platform/src/reply.rs
KirilMihaylov: Yes, it does say it's preferred. My train of thought was that it might introduce a second, nested, call but on closer inspection the implementation is indeed marked with `#[inline]`. The more you know. Will make it use `Into` right away.

platform/packages/platform/src/reply.rs
KirilMihaylov: Must be too scarred from the old API. Didn't even even notice. Thanks!

platform/packages/platform/src/reply.rs
KirilMihaylov: I think that it meant to prefer using `Into` when generics are involved, as that is where it might cause a problem.

gmanev7 (approved):

PR #218, opened 2023-11-29
refactor: Cleanup unnecessary impls and bounds

KirilMihaylov:
In addition to cleanup, a check on PriceDTO is added to ensure both currencies belong to their groups.

### DIFF
--- a/platform/packages/currency/src/group.rs
+++ b/platform/packages/currency/src/group.rs
@@ -1,8 +1,12 @@
+use std::fmt::Debug;
+
+use sdk::schemars::JsonSchema;
+
 use crate::SymbolSlice;
 
 use super::{matcher::Matcher, AnyVisitor, AnyVisitorResult};
 
-pub trait Group: PartialEq {
+pub trait Group: Debug + Copy + Eq + JsonSchema + 'static {
     const DESCR: &'static str;
 
     fn maybe_visit<M, V>(matcher: &M, symbol: &SymbolSlice, visitor: V) -> MaybeAnyVisitResult<V>
--- a/platform/packages/currency/src/lib.rs
+++ b/platform/packages/currency/src/lib.rs
@@ -1,6 +1,6 @@
 use std::{any::TypeId, fmt::Debug};
 
-use serde::{de::DeserializeOwned, Serialize};
+use sdk::schemars::JsonSchema;
 
 use crate::error::{Error, Result};
 
@@ -31,7 +31,7 @@ pub type SymbolOwned = String;
 // Not extending Serialize + DeserializeOwbed since the serde derive implementations fail to
 // satisfy trait bounds with regards of the lifetimes
 // Foe example, https://stackoverflow.com/questions/70774093/generic-type-that-implements-deserializeowned
-pub trait Currency: Copy + Ord + Default + Debug + 'static {
+pub trait Currency: Debug + Copy + Ord + JsonSchema + Sized + 'static {
     /// Identifier of the currency
     const TICKER: SymbolStatic;
 
@@ -83,7 +83,7 @@ pub fn maybe_visit_any<M, C, V>(
 ) -> MaybeAnyVisitResult<V>
 where
     M: Matcher + ?Sized,
-    C: Currency + Serialize + DeserializeOwned,
+    C: Currency,
     V: AnyVisitor,
 {
     if matcher.match_::<C>(symbol) {
--- a/platform/packages/finance/src/coin/dto/mod.rs
+++ b/platform/packages/finance/src/coin/dto/mod.rs
@@ -4,13 +4,13 @@ use std::{
     result::Result as StdResult,
 };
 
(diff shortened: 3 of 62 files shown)


platform/packages/currency/src/group.rs
```diff
@@ -1,8 +1,12 @@
+use std::fmt::Debug;
+
+use sdk::schemars::JsonSchema;
+
 use crate::SymbolSlice;
 
 use super::{matcher::Matcher, AnyVisitor, AnyVisitorResult};
 
-pub trait Group: PartialEq {
+pub trait Group: Debug + Copy + Eq + JsonSchema + 'static {
```
gmanev7, line 9: two things:
1. why do we impose more constraints adding more super-traits?
2. not agree on depending to `sdk` at this core module

platform/packages/currency/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub type SymbolOwned = String;
 // Not extending Serialize + DeserializeOwbed since the serde derive implementations fail to
 // satisfy trait bounds with regards of the lifetimes
 // Foe example, https://stackoverflow.com/questions/70774093/generic-type-that-implements-deserializeowned
-pub trait Currency: Copy + Ord + Default + Debug + 'static {
+pub trait Currency: Debug + Copy + Ord + JsonSchema + Sized + 'static {
```
gmanev7, line 34: [same like the issues above about Group]

platform/packages/finance/src/coin/dto/mod.rs
```diff
@@ -28,8 +28,11 @@ mod unchecked;
 /// carries also the currency ticker. The aim is to use it everywhere the cosmwasm
 /// framework does not support type parameterization or where the currency type
 /// is unknown at compile time.
-#[derive(Clone, PartialEq, Eq, Debug, Serialize, Deserialize, JsonSchema)]
-#[serde(try_from = "unchecked::CoinDTO")]
+#[derive(PartialEq, Eq, Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(
+    try_from = "unchecked::CoinDTO",
+    bound(serialize = "", deserialize = "")
```
gmanev7, line 34: what bounds do we override with this?

platform/packages/finance/src/coin/dto/mod.rs
```diff
@@ -40,9 +43,22 @@ where
     // keep a Coin<C> in a Box<Member<G>> replacing all the struct member variables
     ticker: SymbolOwned,
     #[serde(skip)]
+    #[schemars(skip)]
     _g: PhantomData<G>,
 }
 
+impl<G> Clone for CoinDTO<G>
+where
+    G: Group,
```
gmanev7, line 52: what is the reason the derived impl is not acceptable? The `Clone` requirement to G?

platform/packages/currency/src/group.rs
KirilMihaylov: I wouldn't have if all the implementations didn't require them. If you are fine with converting them to manual implementations instead of `#[derive(...)]` ones, there won't be a need for that.

The whole problem is that the derived implementations place those bounds on all generic types, not on the fields themselves.

KirilMihaylov: About (2), I'd like to ask why was it added in the first place, if we don't want it? I can remove it, but perhaps we'd like to explore ways to avoid `schemars`'s problems which I encountered during the refactoring. Another possibility is to drop it entirely and maybe find a better suiting one.

platform/packages/currency/src/lib.rs
KirilMihaylov: As mentioned in the above comment. Those are common bounds, thus can be moved up to here. If we don't want those, we'd have to convert to manual implementations.

platform/packages/finance/src/coin/dto/mod.rs
KirilMihaylov: As you might have noticed, without this override, the derived versions place additional bounds on all generics. Since currencies and groups are logical constructs, not functional ones, this code works exactly as we'd expect it to without them implementing, as you liked to refer to it when we talked about `#[cfg_attr(..., derive(Debug))]`, unnecessary code. If we are not making exceptions for Debug, whose implementation is dropped once linker strips unused symbols, then this is more than unacceptable.

platform/packages/finance/src/coin/dto/mod.rs
KirilMihaylov: Yes, indeed it is the bound on the generic. Those are, as mentioned above, logical constructs with no relation to the data itself. As already mentioned, this is unacceptable by the standards you laid out.

gmanev7: let's refrain from implementing Close by ourselves since we're going to turn CoinDTO into Copy-able

PR #226, opened 2023-11-30
feat: Added protocol listing & addresses queries

KirilMihaylov:


platform/contracts/admin/src/msg.rs
```diff
@@ -67,4 +67,7 @@ pub struct MigrateContracts {
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub enum QueryMsg {
     InstantiateAddress { code_id: CodeId, protocol: String },
+    Protocols {},
+    PlatformAddresses {},
```
gmanev7, line 71: just `Platform`

platform/contracts/admin/src/msg.rs
```diff
@@ -67,4 +67,7 @@ pub struct MigrateContracts {
...
+    ProtocolAddresses { protocol: String },
```
gmanev7, line 72: just `Protocol`

platform/contracts/admin/src/state/contracts.rs
```diff
@@ -48,7 +48,22 @@ pub(crate) fn add_protocol_set(
...
+    PLATFORM.load(storage).map_err(Into::into)
+}
+
+pub(crate) fn load_protocol(storage: &dyn Storage, protocol: String) -> Result<Protocol<Addr>> {
+    PROTOCOL.load(storage, protocol).map_err(Into::into)
+}
+
+pub(crate) fn load_all(storage: &dyn Storage) -> Result<ContractsGroupedByProtocol> {
     PLATFORM
         .load(storage)
```
gmanev7, line 68: call `load_platform` instead

platform/contracts/admin/src/msg.rs
gmanev7, line 72: 1. please document the response type
2. make sure response types are visible outside the crate. They should be consumed by the callers.

platform/contracts/admin/src/msg.rs
gmanev7, line 1: Past this PR the Admin contract would be consumed by other contracts, namely Dispatcher. We should provide an API-only and API+impl build. Would suggest you to:
- build the API-only with no feature provided
- build the impl when the feature "contract"  is provided

gmanev7 (approved):

PR #230, opened 2023-12-01
refactor!: Merge `cosmwasm-bindings` and `contract` features

KirilMihaylov:
As I investigated the `entry_point` attribute macro, it appears to only add exports only when the target is WebAssembly. As a consequence of that even when the integration tests pull the contract packages with the `contract` feature it won't affect the workflow, as the integration tests are in a way prohibited from targeting WebAssembly.

The result is that, `contract` and `cosmwasm-bindings` can be freely combined, which is the intent of this PR.

### DIFF
--- a/platform/contracts/admin/src/endpoints.rs
+++ b/platform/contracts/admin/src/endpoints.rs
@@ -1,14 +1,6 @@
 use access_control::ContractOwnerAccess;
 use platform::{batch::Batch, contract::CodeId, response};
-#[cfg(feature = "contract")]
-use sdk::cosmwasm_std::entry_point;
-use sdk::{
-    cosmwasm_ext::Response as CwResponse,
-    cosmwasm_std::{
-        ensure_eq, to_json_binary, Addr, Api, Binary, CodeInfoResponse, Deps, DepsMut, Env,
-        MessageInfo, QuerierWrapper, Reply, Storage, WasmMsg,
-    },
-};
+use sdk::{cosmwasm_std::{entry_point, ensure_eq, to_json_binary, Addr, Api, Binary, CodeInfoResponse, Deps, DepsMut, Env, MessageInfo, QuerierWrapper, Reply, Storage, WasmMsg}, cosmwasm_ext::Response as CwResponse};
 use versioning::{package_version, version, SemVer, Version, VersionSegment};
 
 use crate::{
@@ -29,7 +21,7 @@ const CONTRACT_STORAGE_VERSION: VersionSegment = 1;
 const PACKAGE_VERSION: SemVer = package_version!();
 const CONTRACT_VERSION: Version = version!(CONTRACT_STORAGE_VERSION, PACKAGE_VERSION);
 
-#[cfg_attr(feature = "contract", entry_point)]
+#[entry_point]
 pub fn instantiate(
     mut deps: DepsMut<'_>,
     _env: Env,
@@ -48,7 +40,7 @@ pub fn instantiate(
     state_contracts::store(deps.storage, contracts).map(|()| response::empty_response())
 }
 
-#[cfg_attr(feature = "contract", entry_point)]
+#[entry_point]
 pub fn migrate(
     mut deps: DepsMut<'_>,
     _env: Env,
@@ -71,7 +63,7 @@ pub fn migrate(
         .and_then(|(label, ())| response::response(label))
 }
 
-#[cfg_attr(feature = "contract", entry_point)]
+#[entry_point]
 pub fn execute(
     mut deps: DepsMut<'_>,
     env: Env,
@@ -117,7 +109,7 @@ pub fn execute(
     }
 }
 
-#[cfg_attr(feature = "contract", entry_point)]
+#[entry_point]
 pub fn sudo(deps: DepsMut<'_>, env: Env, msg: SudoMsg) -> ContractResult<CwResponse> {
     match msg {
         SudoMsg::ChangeDexAdmin { ref new_dex_admin } => ContractOwnerAccess::new(deps.storage)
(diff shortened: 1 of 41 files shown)


KirilMihaylov:
Full workflow run, including optimized builds: https://github.com/nolus-protocol/nolus-money-market/actions/runs/7058585327

KirilMihaylov:
Manually verified the entry points are exported.

platform/contracts/admin/src/endpoints.rs
```diff
@@ -1,6 +1,6 @@
 use access_control::ContractOwnerAccess;
 use platform::{batch::Batch, contract::CodeId, response};
-#[cfg(feature = "cosmwasm-bindings")]
+#[cfg(feature = "contract")]
```
gmanev7, line 3: The current module is already under `#[cfg(feature = "contract")]` so this is not needed

platform/contracts/admin/src/endpoints.rs
```diff
@@ -29,7 +29,7 @@ const CONTRACT_STORAGE_VERSION: VersionSegment = 1;
 const PACKAGE_VERSION: SemVer = package_version!();
 const CONTRACT_VERSION: Version = version!(CONTRACT_STORAGE_VERSION, PACKAGE_VERSION);
 
-#[cfg_attr(feature = "cosmwasm-bindings", entry_point)]
+#[cfg_attr(feature = "contract", entry_point)]
```
gmanev7, line 32: [all] only `#[entry_point]` suffices

platform/contracts/dispatcher/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license.workspace = true
 crate-type = ["cdylib", "rlib"]
 
 [features]
-cosmwasm-bindings = []
+contract = []
```
gmanev7, line 14: I argue that this feature is not necessary

platform/contracts/dispatcher/src/contract.rs
```diff
@@ -28,7 +28,7 @@ const CONTRACT_STORAGE_VERSION: VersionSegment = 1;
 const PACKAGE_VERSION: SemVer = package_version!();
 const CONTRACT_VERSION: Version = version!(CONTRACT_STORAGE_VERSION, PACKAGE_VERSION);
 
-#[cfg_attr(feature = "cosmwasm-bindings", entry_point)]
+#[cfg_attr(feature = "contract", entry_point)]
```
gmanev7, line 31: [all] again only `#[entry_point]` suffices

platform/contracts/timealarms/src/contract.rs
```diff
@@ -2,7 +2,7 @@ use platform::{
     batch::{Emit, Emitter},
     response,
 };
-#[cfg(feature = "cosmwasm-bindings")]
+#[cfg(feature = "contract")]
```
gmanev7, line 5: Similarly to the Admin, the module is already under `cfg(feature = contract)`. Noticed that it's in also when `test` ... have doubts it should. I believe it was added because the module contains tests .... but they are relevant to the contract only, so should be executed only when the feature is `contract`

gmanev7 (commented):
Did not list all occurrences of the issues already found. I believe you will get a taste of them and fix them accordingly.

platform/contracts/admin/src/endpoints.rs
KirilMihaylov: Good point, that didn't cross my mind.

protocol/contracts/lease/src/contract/mod.rs
```diff
@@ -5,11 +5,13 @@ use sdk::cosmwasm_std::QuerierWrapper;
 
 use crate::lease::{with_lease::WithLease, LeaseDTO};
 
+#[cfg(feature = "contract")]
 pub use self::endpoins::{execute, instantiate, migrate, query, reply, sudo};
 use self::finalize::FinalizerRef;
 
 mod api;
 mod cmd;
+#[cfg(feature = "contract")]
```
gmanev7, line 14: not sure if we need `contract` feature at the lease at all cause we have `#[cfg(any(feature = "astroport", feature = "osmosis"))]`

protocol/contracts/leaser/src/lib.rs
```diff
@@ -1,7 +1,10 @@
-pub use crate::error::ContractError;
+pub use self::error::ContractError;
+#[cfg(feature = "contract")]
+pub use self::contract::{execute, instantiate, migrate, query, reply, sudo};
 
 mod cmd;
-pub mod contract;
+#[cfg(feature = "contract")]
```
gmanev7, line 6: similarly to the lease, I do not think we need a dedicated `contract` feature

protocol/contracts/profit/src/lib.rs
```diff
@@ -6,20 +6,15 @@ pub mod typedefs;
...
-
-#[cfg(any(feature = "astroport", feature = "osmosis"))]
+#[cfg(all(feature = "contract", any(feature = "astroport", feature = "osmosis")))]
 pub mod profit;
-
 #[cfg(any(feature = "astroport", feature = "osmosis"))]
 pub mod result;
-
-#[cfg(any(feature = "astroport", feature = "osmosis"))]
+#[cfg(all(feature = "contract", any(feature = "astroport", feature = "osmosis")))]
```
gmanev7, line 19: I guess adding `contract` as a dependent feature to `astroport` and `osmosis` would turn this change into useless.

KirilMihaylov: Not necessarily. `error` module doesn't require it.

PR #232, opened 2023-12-04
fix(tests): Fixed and unified profit int. tests

### LINKED ISSUE #216: Profit integration tests fail when adding NLS to oracle's swap tree

KirilMihaylov:
Closes #216

### DIFF
--- a/tests/src/common/swap.rs
+++ b/tests/src/common/swap.rs
@@ -1,4 +1,5 @@
-use currency::SymbolSlice;
+use std::ops::Deref;
+
 use finance::coin::Amount;
 use sdk::{
     cosmos_sdk_proto::traits::Message,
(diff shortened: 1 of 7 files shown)


tests/src/common/swap.rs
```diff
@@ -17,6 +18,28 @@ use super::{
     ADMIN,
 };
 
+#[repr(transparent)]
+#[derive(Debug, Eq)]
+pub struct DexDenom<'r>(&'r str);
```
gmanev7, line 23: could you outline the motivation behind your decision to define it?

gmanev7 (commented):
The test functions are too long, resembling other tests in their previous state. Please review, extract the common code in functions, and reuse it.

tests/src/common/swap.rs
KirilMihaylov: So it is clear a DEX symbol is passed down and not a bank one. The whole motivation was to avoid ambiguity.

gmanev7 (approved):

PR #240, opened 2023-12-07
 Remove serde alias attribute for the min_transaction…

maneva3:
… property


gmanev7:
We should double-check the locations where the old named field is being used and assess whether a migration is required. Initially, we have considered only the Lease-s ... but recently I have found another location.

gmanev7:
> We should double-check the locations where the old named field is being used and assess whether a migration is required. We initially considered only the Lease-s ... but recently I have found another location.
The other location was the Leaser config. After thorough analysis, we discovered that a regular lease codeId update in the Leaser config migrates the config content. Since all the networks have undergone such migration we can confirm that an explicit migration is unnecessary.

gmanev7 (approved):

PR #242, opened 2023-12-08
fix: Fix admin contract migration

KirilMihaylov:


platform/contracts/admin/src/msg.rs
```diff
@@ -6,7 +6,7 @@ use sdk::{
     schemars::{self, JsonSchema},
 };
 
-use crate::contracts::{
+pub use crate::contracts::{
```
gmanev7, line 9: not sure I understand the motivation behind this, could you elaborate?

KirilMihaylov: It was not possible to access the type from outside, thus any contract and/or test that had to interact with it was blocked from doing so.

PR #246, opened 2023-12-11
refactor!: Admin migration process

KirilMihaylov:


platform/contracts/admin/src/endpoints.rs
```diff
@@ -134,14 +137,23 @@ pub fn sudo(deps: DepsMut<'_>, env: Env, msg: SudoMsg) -> ContractResult<CwRespo
...
             post_migration_execute,
-        }) => crate::contracts::migrate(
-            deps.storage,
-            env.contract.address,
-            release,
-            admin_contract,
-            migration_spec,
-            post_migration_execute,
-        )
+        }) => if let Some(admin_contract) = admin_contract {
```
gmanev7, line 140: Let's not combine two scenarios under the same API. That makes it difficult to use.

Define a dedicated `SudoMsg::MigrateSelf` ... or even better, let's use the native migration for the Admin. Then remove the ability to migrate the Admin with a SudoMsg.

platform/contracts/admin/src/endpoints.rs
```diff
@@ -48,30 +48,33 @@ pub fn instantiate(
...
-                },
-                Into::into,
-            )
-        })
-        .and_then(|(label, ())| response::response(label))
+            ContractState::AwaitingSelfMigrationReply {
+                release,
+                migration_spec,
+                post_migration_execute,
+            }
```
gmanev7, line 71: why do we set to wait infinitely for a reply?

platform/contracts/admin/src/msg.rs
```diff
@@ -21,9 +19,8 @@ pub struct InstantiateMsg {
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub struct MigrateMsg {
-    pub protocol_name: String,
-    pub network_name: String,
-    pub dex_admin: Addr,
+    pub migration_spec: String,
+    pub post_migration_execute: String,
```
gmanev7, line 23: what migration messages are these for? The Admin itself or the contracts? You use them as the latter .... in the `migrate` endpoint that should be about the Admin only. 
In addition, if they were for the Admin then we would have them in a structured form, not in a regular json String.

platform/contracts/admin/src/endpoints.rs
```diff
@@ -160,7 +160,22 @@ fn register_protocol(
...
+            post_migration_execute,
+        } => migration_reply(msg, release).and_then(|release| {
+            crate::contracts::migrate(
+                deps.storage,
+                release,
+                from_json(migration_spec)?,
+                from_json(post_migration_execute)?,
+            )
+            .map(response::response_only_messages)
+        }),
```
gmanev7, line 175: not sure why we need to act in this case?

platform/packages/versioning/src/release.rs
```diff
@@ -9,6 +11,21 @@ use super::{Version, VersionSegment};
 #[serde(transparent)]
 pub struct ReleaseLabel(&'static str);
 
+impl Display for ReleaseLabel {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        f.write_str(self.0)
+    }
+}
+
+impl<S> PartialEq<S> for ReleaseLabel
```
gmanev7, line 20: not a good idea to allow `release == "some_txt"` ... in general, there should be a solid reason to justify such a case. Do we have?

KirilMihaylov: Yes, indeed we do. There is absolutely no other way to compare it with the passed string which represents the expected one, if we don't want to allow it to be built from outside it's package.

platform/contracts/admin/src/msg.rs
```diff
@@ -48,13 +46,13 @@ pub enum SudoMsg {
     ChangeDexAdmin { new_dex_admin: Addr },
     RegisterProtocol { name: String, protocol: Protocol },
     MigrateContracts(MigrateContracts),
+    ClearStorage {},
```
gmanev7, line 49: please add a doc string explaining the purpose of that its intended for internal usage

platform/contracts/admin/src/endpoints.rs
```diff
@@ -48,30 +48,34 @@ pub fn instantiate(
...
+                post_migration_execute,
             )
-        })
-        .and_then(|(label, ())| response::response(label))
+            .and_then(|messages| response::response_with_messages(label, messages))
+        } else {
+            Err(ContractError::WrongRelease {
+                reported: label.to_string(),
+                expected: release,
+            })
```
gmanev7, line 76: let's reuse the way we check and report that error in `fn migration_reply`

platform/contracts/admin/src/endpoints.rs
gmanev7, line 1: please move down the `fn register_protocol` as it appears before other public members

platform/packages/versioning/src/release.rs
gmanev7: as usually, the need of such a mild hacks come due to errors in the design. In our case, that is due to the misusage of `String` where `ReleaseLabel` should be.

platform/contracts/admin/src/endpoints.rs
```diff
@@ -60,7 +60,15 @@ pub fn migrate(
...
-        if label == release {
+        ensure_eq!(
+            label,
+            release,
+            ContractError::WrongRelease {
+                reported: label.to_string(),
+                expected: release,
+            }
+        );
+
```
gmanev7: please reuse in a real manner, define a `fn check_release(label: ..., expected: ...) -> ContractResult<CwResponse>`

gmanev7 (approved):

gmanev7 (approved):

PR #247, opened 2023-12-11
refactor(ci): Fail-fast on old depenendency versions

KirilMihaylov:


.github/workflows/smart_contracts.yaml
```diff
@@ -16,12 +16,28 @@ on:
...
+    runs-on: "ubuntu-latest"
+    if: "github.event_name == 'tag'"
+    strategy:
+      fail-fast: true
+      matrix:
+        working_directory: ["platform", "protocol", "tests", "tools"]
+    steps:
+      - uses: "actions/checkout@v4.1.1"
+      - shell: "sh"
+        run: "cargo +stable update --locked"
```
gmanev7, line 29: why `+stable`? why not the one we specify?

gmanev7 (approved):

.github/workflows/smart_contracts.yaml
KirilMihaylov: I have noticed it downloads the Rust toolchain even for `cargo clean`, so instead of adding the caching action, which would add /even tho little, nonetheless/ overhead, I just opted for the minimal approach.

PR #261, opened 2023-12-22
refactor(lease): Move .price_at() in position spec

maneva3:


protocol/contracts/lease/src/position/mod.rs
```diff
@@ -99,6 +104,21 @@ where
...
 
+    /// Calculate the price at which the lease reaches given ltv.
+    pub(crate) fn price_at(
+        &self,
+        level: Level,
+        total_due: Coin<Lpn>,
+    ) -> ContractResult<Price<Asset, Lpn>> {
+        debug_assert!(
+            !total_due.is_zero(),
+            "Loan already paid, no need of next alarms!"
```
gmanev7, line 115: the message is not in the right context, better remove it

gmanev7 (approved):

PR #269, opened 2024-01-11
fix(ci): Fix WASM binary checking

KirilMihaylov:


scripts/build-and-optimize.sh
```diff
@@ -2,9 +2,12 @@
...
-cd '/code/'
+cd '/code/' || exit
 
-cargo update --locked
+if ! test "${DANGER_SKIP_CARGO_LOCK_UPDATE}" = '1'
+then
+  cargo update --locked
+fi
 
 if ! test $? -eq 0
```
gmanev7, line 12: this should be moved up right after `cargo update --locked`

.github/actions/optimized_build/action.yaml
```diff
@@ -35,6 +38,13 @@ runs:
 
         echo "${{ inputs.release_version }}" >> "./artifacts/release_version"
 
+        DANGER_SKIP_CARGO_LOCK_UPDATE='0'
```
gmanev7, line 41: a couple of things:
- let's refrain from using extreme-meaning words, as here "DANGER"
- waaaaay easier to read and maintain when the naming and logic do not involve negations. Rename to `CHECK_DEPENDENCIES_UPDATED`.
- for value use words/strings rather than digits referring to semantics in some distinguished programming languages

scripts/build-and-optimize.sh
```diff
@@ -2,15 +2,18 @@
 
 RUSTFLAGS="-C link-arg=-s ${RUSTFLAGS}"
 
-cd '/code/'
+cd '/code/' || exit
 
-cargo update --locked
-
-if ! test $? -eq 0
+if ! test "${CHECK_DEPENDENCIES_UPDATED}" = 'false'
```
gmanev7, line 7: negation in the meaning, please fix

gmanev7 (approved):

PR #278, opened 2024-01-25
In tests call spec::validate_close_amount() instead of position::validate_close_amount()

maneva3:


protocol/contracts/lease/src/position/spec/mod.rs
```diff
@@ -1163,24 +1162,28 @@ mod test_validate_close {
 
     #[test]
     fn amount_as_min_transaction() {
-        let spec = position(100, 85, 15);
-        let result_1 = spec.validate_close_amount(15.into(), price(1, 1));
+        let spec = spec(85, 15);
+        let asset: Coin<TestCurrency> = 100.into();
```
gmanev7, line 1166: [everywhere in this mod] It's too much to specify each time the type. Alternatives?

gmanev7 (approved):

PR #281, opened 2024-02-02
feat: Add dex request parsing

KirilMihaylov:


protocol/packages/dex/src/swap.rs
```diff
@@ -6,12 +6,26 @@ use oracle::api::swap::SwapPath;
...
+    pub token_in: CoinDTO<GIn>,
+    pub swap_path: SwapPath,
+}
+
 pub trait ExactAmountIn {
+    #[cfg(feature = "testing")]
+    fn parse_request<GIn>(request: Any) -> SwapRequest<GIn>
+    where
+        GIn: Group;
+
```
gmanev7, line 23: the rule of ordering here was 'production/test' - first, 'build/parse' - second
so this fn should be move accordingly

protocol/packages/swap/src/utils.rs
gmanev7, line 1: this module lowers the code quality due to:
- the need for functionality under feature flags at so low/deep level' - remedy: move the code in the respective modules
- 'utils' along with a few other words are too-general to be used in our code base

protocol/packages/swap/src/utils.rs
```diff
@@ -0,0 +1,59 @@
...
+{
+    assert_eq!(
+        request.type_url, type_url,
+        "Different type URL than expected one encountered!"
+    );
+
+    T::decode(request.value.as_slice()).expect("Expected a swap request message!")
+}
+
+pub(crate) fn parse_token<G>(amount: &str, denom: String) -> CoinDTO<G>
```
gmanev7, line 28: since it is only one fn we may not be able to introduce a module with a name to classify the functionality defined in, then move it up to the lib.rs

protocol/packages/swap/src/lib.rs
```diff
@@ -1,11 +1,37 @@
...
+            .get(),
+        denom,
+    )
+    .expect("Expected swap-in token to be part of selected group!")
+}
+
+// #[cfg(any(test, feature = "testing"))] revert TODO report a cargo bug that 'test' cfg is not applied
+#[cfg(feature = "testing")]
+#[cold]
+pub(crate) fn pattern_match_else(message_name: &str) -> ! {
```
gmanev7, line 33: Too high visibility, right, technically it is the same as if it was without `pub(crate)` but if it happens to be moved then it will be not the same. Prefer expressing the intention explicitly ... which in our case is module-private, i.e. without `pub(crate)`

protocol/packages/swap/src/lib.rs
```diff
@@ -1,11 +1,37 @@
...
-pub use self::osmosis::*;
 
 #[cfg(feature = "astroport")]
 mod astroport;
 #[cfg(feature = "osmosis")]
 mod osmosis;
+
+// #[cfg(any(test, feature = "testing"))] revert TODO report a cargo bug that 'test' cfg is not applied
+#[cfg(feature = "testing")]
+pub(crate) fn parse_token<G>(amount: &str, denom: String) -> finance::coin::CoinDTO<G>
```
gmanev7, line 16: same as for the next fn

protocol/packages/swap/src/osmosis.rs
```diff
@@ -73,7 +83,41 @@ impl ExactAmountIn for Impl {
...
+            sender: _,
+            routes,
+            token_in: Some(token_in),
+            token_out_min_amount,
+        } = parse_request_from_any_and_type_url(request, RequestMsg::TYPE_URL)
+        else {
+            crate::pattern_match_else(type_name::<RequestMsg>())
+        };
+
+        assert_eq!({ token_out_min_amount }, "1");
```
gmanev7, line 100: what is the purpose of this extra block?

protocol/packages/swap/src/astroport.rs
```diff
@@ -23,37 +33,42 @@ use sdk::{
     },
     cosmwasm_std::{self, Coin as CwCoin, Decimal},
 };
-use serde::{Deserialize, Serialize};
 
 pub type RequestMsg = MsgExecuteContract;
```
gmanev7, line 37: [here and on Osmosis' impl] just to remind you that 'pub' should be removed once migrate int.tests

protocol/packages/swap/src/astroport.rs
```diff
@@ -282,15 +401,17 @@ mod test {
         ));
     }
 
+    // revert TODO report a cargo bug that 'test' cfg is not applied
+    #[cfg(feature = "testing")]
     #[test]
     fn resp() {
```
gmanev7, line 407: please add a corresponding test(s) for the request part: a few different cases of build and parse

protocol/packages/swap/src/lib.rs
```diff
@@ -1,11 +1,37 @@
+#[cfg(feature = "osmosis")]
+pub use self::osmosis::*;
```
gmanev7, line 2: let's specify the Impl only

protocol/packages/swap/src/osmosis.rs
KirilMihaylov: Nothing else than to force the value to be moved out and dropped on the spot after the comparison. Most likely I did that unintentionally as there isn't a loop or a long-running function afterwards.

protocol/packages/swap/src/astroport.rs
KirilMihaylov: Will put a TODO for that in this case!

protocol/packages/swap/src/lib.rs
KirilMihaylov: Not yet possible. Can be done only after refactoring the integration tests.

protocol/packages/swap/src/astroport.rs
```diff
@@ -23,37 +30,43 @@ use sdk::{
     },
     cosmwasm_std::{self, Coin as CwCoin, Decimal},
 };
-use serde::{Deserialize, Serialize};
 
+// TODO change visibility to private
 pub type RequestMsg = MsgExecuteContract;
```
gmanev7, line 35: I believe this type is not used outside of this module so please resolve the TODO

KirilMihaylov: That's true. Thanks!

protocol/packages/swap/tests/request.rs
```diff
@@ -0,0 +1,60 @@
...
+use sdk::{
+    cosmos_sdk_proto::Any as CosmosAny, cosmwasm_std::Binary,
+    neutron_sdk::bindings::types::ProtobufAny as NeutronAny,
+};
+use swap::Impl;
+
+#[test]
+fn build_and_parse() {
+    let expected_token_in =
+        finance::coin::from_amount_ticker::<SubGroup>(20, SubGroupTestC1::TICKER.into()).unwrap();
```
gmanev7, line 17: we have got a much more preferred way of creating Coin<DTO>
`let expected_token_in: CoinDTO<SubGroup> = Coin::<SubGroupTestC1>::from(20).into()`

gmanev7 (approved):

PR #293, opened 2024-02-21
feat: Added testing utilities for loading raw contract state

KirilMihaylov:


platform/Cargo.toml
```diff
@@ -29,10 +29,11 @@ tree = { path = "packages/tree", default-features = false }
 versioning = { path = "packages/versioning", default-features = false }
 
 # General
-serde = { version = "1", default-features = false }
-thiserror = { version = "1", default-features = false }
 anyhow = "1"
 base64 = { version = "0.21", default-features = false }
+data-encoding = "2.5.0"
```
gmanev7, line 34: is not there an overlap in the functionalities of the two?

platform/packages/sdk/src/testing.rs
```diff
@@ -79,6 +87,127 @@ pub fn new_app(message_sender: InterChainMsgSender) -> CwAppBuilder {
...
+
+pub fn try_load_into_storage<S, I, E>(storage: &mut S, mut iter: I) -> Result<(), E>
+where
+    S: Storage + ?Sized,
+    I: Iterator<Item = Result<KvPair, E>>,
+{
+    iter.try_for_each(|kv_pair| {
+        kv_pair.map(|kv_pair| storage.set(kv_pair.key.as_ref(), kv_pair.value.as_ref()))
+    })
+}
```
gmanev7, line 155: could you shed some light on these functions' intended use case?

platform/Cargo.toml
KirilMihaylov: Yes, they do indeed overlap but I haven't tested them against each other to see which one would be more performant and/or efficient. The intent here is to use `data-encoding` only for testing purposes until we are sure that we don't want to replace `base64` altogether.

platform/packages/sdk/src/testing.rs
```diff
@@ -79,6 +87,127 @@ pub fn new_app(message_sender: InterChainMsgSender) -> CwAppBuilder {
...
+    #[error("Decoding error has occurred! Cause: {0}")]
+    Decode(DecodeError),
+}
+
+/// CSV format should be as follows:
+/// ```plaintext
+/// KEY,VALUE
+/// KEY,VALUE
+/// ...
+/// KEY,VALUE
```
gmanev7, line 172: please document what is the expected encoding

export-state-to-csv.sh
```diff
@@ -0,0 +1,161 @@
...
+check_too_few_args() {
+  if [ "${1}" -lt 2 ]
+  then
+    echo "Too few parameters! Expected ${2}!"
+
+    exit 1
+  fi
+}
+
+ensure_not_is_set() {
```
gmanev7, line 34:
```suggestion
ensure_not_set() {
```

export-state-to-csv.sh
gmanev7, line 1: it's not too general to stay at the root of the file hierarchy. What about 'scripts\'?

platform/packages/sdk/src/testing.rs
```diff
@@ -79,6 +87,141 @@ pub fn new_app(message_sender: InterChainMsgSender) -> CwAppBuilder {
...
+    iter.for_each(|kv_pair| storage.set(kv_pair.key.as_ref(), kv_pair.value.as_ref()))
+}
+
+pub fn try_load_into_storage<S, I, E>(storage: &mut S, mut iter: I) -> Result<(), E>
+where
+    S: Storage + ?Sized,
+    I: Iterator<Item = Result<KvPair, E>>,
+{
+    iter.try_for_each(|kv_pair| {
+        kv_pair.map(|kv_pair| storage.set(kv_pair.key.as_ref(), kv_pair.value.as_ref()))
```
gmanev7, line 153: please define the common lambda as a first-class named function and reuse

platform/packages/sdk/src/testing.rs
gmanev7, line 1: the name is too generic ... let's rename to something more specific, for example cosmwasm or contract

platform/packages/sdk/src/testing/manage_state/decoded.rs
```diff
@@ -0,0 +1,23 @@
+use super::encoding::Encoding;
+
+pub struct Decoded<Encoding: self::Encoding>(Box<[u8]>, PhantomData<Encoding>);
```
gmanev7, line 3: I think the name of the struct is misleading .... is it rather `Encoded`?

protocol/Cargo.lock
gmanev7, line 1: not sure I understand how these changes got into the PR

tests/Cargo.lock
gmanev7, line 1: not sure I understand how these changes got into the PR

protocol/Cargo.lock
KirilMihaylov: My bad; Mixed the branches

tests/Cargo.lock
KirilMihaylov: My bad; Mixed the branches

protocol/Cargo.lock
KirilMihaylov: After further investigation I saw that they actually should be here. `data-encoding` is added down the other `Cargo.lock`s as it's part of `sdk/testing`.

tests/Cargo.lock
KirilMihaylov: After further investigation I saw that they actually should be here. `data-encoding` is added down the other `Cargo.lock`s as it's part of `sdk/testing`.

platform/packages/sdk/src/testing/manage_state/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub fn load_into_storage<S, I>(storage: &mut S, iter: I)
+where
+    S: Storage + ?Sized,
+    I: Iterator<Item = KvPair>,
+{
+    iter.for_each(|kv_pair| storage.set(kv_pair.key().as_ref(), kv_pair.value().as_ref()))
+}
+
+pub fn try_load_into_storage<S, I, E>(storage: &mut S, mut iter: I) -> Result<(), E>
```
gmanev7, line 22: locally used, so please reduce its visibility

scripts/export-state-to-csv.sh
```diff
@@ -0,0 +1,161 @@
...
+  then
+    echo "\"${2}\" is already set!"
+
+    exit 1
+  fi
+}
+
+while true
+do
+  if [ "${#}" -eq 0 ]
```
gmanev7, line 45: easier to read and maintain if this is moved as a condition of the `while`

gmanev7 (approved):

gmanev7 (approved):

PR #302, opened 2024-03-11
feat(reserve): intro Protocol Reserve

gmanev7:


protocol/contracts/reserve/src/contract.rs
```diff
@@ -0,0 +1,118 @@
...
+        .and_then(|()| Config::from(new_reserve).store(deps.storage))
+        .map(|()| response::empty_response())
+        .or_else(|err| platform_error::log(err, deps.api))
+}
+
+#[entry_point]
+pub fn migrate(deps: DepsMut<'_>, __env: Env, _msg: MigrateMsg) -> ContractResult<CwResponse> {
+    versioning::update_software(deps.storage, CONTRACT_VERSION, Into::into)
+        .and_then(response::response)
+        .or_else(|err| platform_error::log(err, deps.api))
```
KirilMihaylov, line 57: Since `platform::error::log` is infallible, I think the behavior we desire is `Result::{map|inspect}_err`. This goes in a very roundabout way about it when internally it's equivalent to `|error| Err(error)`.

protocol/contracts/reserve/src/contract.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+    .map(response::response_only_messages)
+    .or_else(|err| platform_error::log(err, deps.api))
+}
+
+#[entry_point]
+pub fn query(deps: Deps<'_>, _env: Env, msg: QueryMsg) -> ContractResult<Binary> {
+    match msg {
+        QueryMsg::Config() => Config::load(deps.storage)
+            .map( Into::<ConfigResponse>::into)
```
KirilMihaylov: I feel like `rustfmt` wasn't ran here.

KirilMihaylov: If you need to specify the type, I think it's better to use `T::from` at this point. Using turbo-fish for such a simple case just hurts readability.

If the argument is that the documentation says `Into` should be preferred, it's meant for generics as it's the only place where that would actually matter. When ran with specific types, at most you'd get a compilation error **if** for some case only `Into` is implemented, which on it's own is against the documentation.

protocol/contracts/reserve/src/contract.rs
```diff
@@ -0,0 +1,118 @@
...
+            .and_then(|config| cosmwasm_std::to_json_binary(&config).map_err(Into::into))
```
KirilMihaylov, line 115: [Suggestion] I'd prefer using `sdk::cosmwasm_std` personally. In case we ever remove the direct dependency, it would be easier to handle.

protocol/contracts/reserve/src/error.rs
```diff
@@ -0,0 +1,26 @@
+use thiserror::Error;
+
+use sdk::cosmwasm_std::StdError;
+
+#[derive(Error, Debug, PartialEq)]
+pub enum ContractError {
+    #[error("[Reserve] [Std] {0}")]
+    Std(#[from] StdError),
+
```
KirilMihaylov, line 9: I'm not sure about all these empty lines but I feel like this either isn't set in the `rustfmt` configuration or we have problems with it reading said configuration.

protocol/contracts/reserve/src/state/config.rs
```diff
@@ -0,0 +1,63 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, JsonSchema)]
+pub struct Config {
+    lease_code_id: CodeId,
+}
+
+impl Config {
+    const STORAGE: Item<'static, Self> = Item::new("config");
+
+    #[cfg(test)]
+    pub const fn new(lease_code_id: CodeId) -> Self {
```
KirilMihaylov, line 21: My IDE doesn't report there being any uses of this. I'd say you should follow your advice and remove it if it's not used.

protocol/contracts/reserve/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+    Config(),
+}
+
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, JsonSchema)]
+pub struct ConfigResponse {
+    pub lpn_ticker: SymbolOwned,
+    pub lease_code_id: Uint64,
+}
+
+impl From<Config> for ConfigResponse {
```
KirilMihaylov, line 54: I'm not sure how I feel about this one. Perhaps a method with `lease_code_id` as a parameter would suffice.

protocol/contracts/reserve/src/state/config.rs
```diff
@@ -0,0 +1,63 @@
...
+        Self::STORAGE
+            .update(storage, |config: Self| Ok(f(config)))
+            .map(|_| ())
+    }
+}
+
+impl From<InstantiateMsg> for Config {
+    fn from(msg: InstantiateMsg) -> Self {
+        Self {
+            lease_code_id: msg.lease_code_id.into(),
```
KirilMihaylov, line 60: [Suggestion] Might be better to not include such implementation but rather make it explicit and include checks for whether the code ID is at the very least valid, as on first inspection CW doesn't provide a way to fetch additional information besides the creator's address and the binary checksum.

KirilMihaylov (changes requested):

protocol/contracts/reserve/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use finance::coin::CoinDTO;
+use sdk::{
+    cosmwasm_std::{Addr, Uint64},
+    schemars::{self, JsonSchema},
+};
+
+use crate::Config;
+
+pub(crate) type LpnCurrencies = Lpns;
```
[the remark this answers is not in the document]
gmanev7: good catch!

protocol/contracts/reserve/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(crate) type LpnCurrency = Lpn;
+pub type LpnCoin = CoinDTO<LpnCurrencies>;
```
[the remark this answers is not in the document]
gmanev7: albeit no warning is generated, aligned its visibility to the type arg

protocol/contracts/reserve/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct MigrateMsg {}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    NewLeaseCode { code_id: Uint64 },
+
```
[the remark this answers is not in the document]
gmanev7: not sure we want to chase rustfmt bugs

protocol/contracts/reserve/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::Config;
+
+pub(crate) type LpnCurrencies = Lpns;
+pub(crate) type LpnCurrency = Lpn;
+pub type LpnCoin = CoinDTO<LpnCurrencies>;
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct InstantiateMsg {
+    pub lease_code_admin: Addr,
```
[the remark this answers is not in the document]
gmanev7: agree

protocol/contracts/reserve/Cargo.toml
```diff
@@ -0,0 +1,57 @@
...
+# Required as a dependency by `entry_point` attribute macro
+cosmwasm-std = { workspace = true, optional = true }
+
+thiserror = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+
+[dev-dependencies]
+schema = { workspace = true }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
```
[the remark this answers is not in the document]
gmanev7: removed the whole dependency as we won't add integration tests in this package

protocol/contracts/reserve/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum SudoMsg {}
+
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, JsonSchema)]
+#[cfg_attr(feature = "testing", derive(Debug))]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum QueryMsg {
+    /// Return a `ConfigResponse`
```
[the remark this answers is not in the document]
gmanev7: [do not recall the source] The documenting convention does not require "it does ..."- like statements.

protocol/contracts/reserve/src/contract.rs
```diff
@@ -0,0 +1,118 @@
...
+                deps.storage.deref_mut(),
+                crate::access_control::LEASE_CODE_ADMIN_KEY,
+            )
+            .check(&info.sender)?;
+
+            Config::update_lease_code(deps.storage, new_code_id.into())
+                .map(|()| PlatformResponse::default())
+        }
+        ExecuteMsg::CoverLiquidationLosses { amount } => {
+            let lease = info.sender;
```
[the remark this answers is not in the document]
gmanev7: not sure I understand your point

protocol/contracts/reserve/src/state/config.rs
KirilMihaylov: Update: Since the `stub` feature is for the next PR, if this is also the case here, feel free to close this.

protocol/contracts/reserve/src/contract.rs
gmanev7: accept it although the difference is sooooo small

protocol/contracts/reserve/src/error.rs
gmanev7: the [config option](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#blank_lines_upper_bound) is still unstable

protocol/contracts/reserve/src/contract.rs
gmanev7: it's [a known flaw](https://github.com/nolus-protocol/nolus-money-market/blob/93c9485b4345551d55fa46472504b212db9e46ad/platform/packages/platform/src/error.rs#L80) which will be fixed once the Result::inspect_err get stabilized

protocol/contracts/reserve/src/contract.rs
gmanev7: I follow the Rust calling conventions for functions

protocol/contracts/reserve/src/state/config.rs
gmanev7: not anymore unused

protocol/contracts/reserve/src/contract.rs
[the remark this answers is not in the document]
KirilMihaylov: I meant that we're assigning `lease` before we check that it actually is a lease /+85/. While it is true that right now all the errors are chained logically with AND, it might not always be the case, e.g. introduction of `Result::or_else` calls.

protocol/contracts/reserve/src/access_control.rs
```diff
@@ -0,0 +1 @@
+pub(crate) static LEASE_CODE_ADMIN_KEY: &str = "lease_code_admin";
```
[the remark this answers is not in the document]
gmanev7: No
As all names should follow, here we name after the **role**, not the one who plays it. Tomorrow we may decide to assign that role to another actor.

protocol/contracts/reserve/src/contract.rs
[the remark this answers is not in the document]
gmanev7: here it's named for convenience, just a shorter name, ... that does not change the semantic

protocol/contracts/reserve/src/contract.rs
```diff
@@ -0,0 +1,118 @@
...
+    _env: Env,
+    _info: MessageInfo,
+    new_reserve: InstantiateMsg,
+) -> ContractResult<CwResponse> {
+    deps.api
+        .addr_validate(new_reserve.lease_code_admin.as_str())
+        .map_err(Into::<ContractError>::into)
+        .and_then(|_| {
+            SingleUserAccess::new(
+                deps.storage.deref_mut(),
```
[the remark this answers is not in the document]
gmanev7: if we do it that deserves a separate PR

protocol/contracts/reserve/src/api.rs
gmanev7: me neither
Why would we need such a method?

protocol/contracts/reserve/src/access_control.rs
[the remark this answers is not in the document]
KirilMihaylov: Yeah, okay. That's what I meant by future-proof. Not having to rename it if the organization changes.

protocol/contracts/reserve/src/contract.rs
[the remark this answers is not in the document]
KirilMihaylov: I don't have beef with the name, I wanted to emphasize that usually when we assign it it's usually the version of the value we'd like to use and is usually already checked.

Feel free to close this comment if you don't agree. It's just a minor concern of mine.

protocol/contracts/reserve/src/contract.rs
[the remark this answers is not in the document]
KirilMihaylov: For sure the overall refactoring would need a separate one. I'm just suggesting it as a kind of a heads up.

protocol/contracts/reserve/src/contract.rs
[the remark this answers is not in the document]
gmanev7: no, I see merit in it .... though the safer alternative, passing it as function arg, would be more complex

protocol/contracts/reserve/src/api.rs
gmanev7: compiled my solution - tightened the encapsulation and retained the link from the Config to ConfigResponse, Thanks!

PR #303, opened 2024-03-11
feat!: Introduce `cargo-each`

KirilMihaylov:


tools/cargo-each/src/check.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use crate::config::Config;
+
+pub(crate) fn configuration(package: &Package, config: &Config<'_>) -> Result<()> {
+    sets(package, config)?;
+
+    combinations(package, config)
+}
+
+fn sets(package: &Package, config: &Config<'_>) -> Result<()> {
```
gmanev7, line 12: reading the function name I do not have any idea what it does

tools/cargo-each/src/check.rs
```diff
@@ -0,0 +1,91 @@
...
+    config.sets.iter().try_for_each(|(name, set)| {
+        if set.members.is_empty() {
+            cold();
```
gmanev7, line 15: premature optimizations may only pollute the code and make reading it more difficult

tools/cargo-each/src/check.rs
```diff
@@ -0,0 +1,91 @@
...
+
+            bail!(
+                r#"Package "{}"'s configuration's set "{}" is empty!"#,
+                package.name,
+                name,
+            );
```
gmanev7, line 21: please follow the "single exit point" best practice

tools/cargo-each/src/check.rs
```diff
@@ -0,0 +1,91 @@
...
+fn combinations(package: &Package, config: &Config<'_>) -> Result<()> {
+    config.combinations.iter().try_for_each(|combination| {
+        if let Some(set) = combination.sets.iter().find(|&set| !config.sets.contains_key(set)) {
+            cold();
+
+            bail!(
+                r#"Package "{}"'s configuration contains combinations referring to undefined set "{}"!"#,
+                package.name,
+                set,
+            );
```
gmanev7, line 59: same comments as for the `sets`

tools/cargo-each/src/combinations_iter/mod.rs
```diff
@@ -0,0 +1,306 @@
...
+    groups: Option<&'r BTreeSet<&'r str>>,
+) -> Result<Option<impl Iterator<Item = String> + 'r>, ()> {
+    if let Some(config) = maybe_config {
+        check::configuration(package, config)
+            .map(move |()| {
+                Some(EitherIter::Left(configured_package_combinations(
+                    package, config, groups,
+                )))
+            })
+            .map_err(|error: Error| eprintln!("Configuration checks failed!\nError: {error}"))
```
gmanev7, line 28: the first time I see "error swallowing" in Rust
I hope it's not necessary to explain more what the problem is with this

tools/cargo-each/src/config.rs
```diff
@@ -0,0 +1,173 @@
...
+            |error| eprintln!("Deserializing configuration failed!\nError: {error}"),
+        )
+        .and_then(|maybe_config| {
+            maybe_config
+                .map(|config| try_resolve_config_generics(package, config))
+                .transpose()
+        })
+}
+
+fn try_resolve_config_generics<'r>(
```
gmanev7, line 85: ~90 LoC for a function is too much

tools/cargo-each/src/iter_or_else_iter.rs
```diff
@@ -0,0 +1,29 @@
+pub(crate) struct IterOrElseIter<Iter: Iterator, AltIter: Iterator<Item = Iter::Item>> {
+    iter: Iter,
+    alt_iter: Option<AltIter>,
+}
+
+impl<Iter: Iterator, AltIter: Iterator<Item = Iter::Item>> IterOrElseIter<Iter, AltIter> {
```
gmanev7, line 6: [everywhere] we put the trait bounds in the `where` clause

tools/cargo-each/src/check.rs
KirilMihaylov: While true, it's in the `check` module. If I named it `check_sets`, `clippy` complains when ran in pedantic mode.

tools/cargo-each/src/config.rs
KirilMihaylov: I have forgotten to break it down. My bad.

tools/Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = ["**/src/.."]
 resolver = "2"
 
 [workspace.dependencies]
-anyhow = "1.0.75"
+anyhow = "1.0.80"
```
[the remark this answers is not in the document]
gmanev7: It should be .... but its message is that it sticks to a particular version. I do not buy the argument of locking to an older version, after all, we regularly update versions. Then, if we agree that `=1.0.80` is semantically equivalent to `=1.0` then the latter is shorter and prettier.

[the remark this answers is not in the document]
KirilMihaylov: I meant someone along the chain. We can't ensure a dependency doesn't drag **us** down because **it** locked down. And, no, `=X.Y.Z` was never semantically equivalent `=X.Y`.

If you want really want to use versions as short as possible, okay, I'll follow that. I just want to make sure you are aware of the risks.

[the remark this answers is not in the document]
KirilMihaylov: As an example of a similar situation, I had to turn off the dependency auditing for quite a while because a dependency locked it's dependency to a specific version. Mind you, that locked dependency was `webpki` and had a high-risk CVE.

Things like this do happen, and people do lock their dependencies sometimes.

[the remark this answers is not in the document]
gmanev7: I do not insist, just wanna follow the same convention throughout the repo

gmanev7 (approved):

PR #305, opened 2024-03-12
refactor(platform): Intro Code

gmanev7:
The aim of this PR is to abstract Cosmwasm Code for the sake of improved validation and better domain representation. The next will be `Contract` that would hide Addr, its construction and validation.

### DIFF
--- a/platform/packages/platform/src/contract.rs
+++ b/platform/packages/platform/src/contract.rs
@@ -1,3 +1,5 @@
+use std::mem;
+
 use sdk::{
     cosmwasm_std::{Addr, CodeInfoResponse, ContractInfoResponse, QuerierWrapper, WasmQuery},
     schemars::{self, JsonSchema},
@@ -9,8 +11,7 @@ use crate::{error::Error, result::Result};
 pub type CodeId = u64;
 
 #[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq, JsonSchema)]
-#[serde(deny_unknown_fields, rename_all = "snake_case")]
-#[serde(transparent)]
+#[serde(deny_unknown_fields, rename_all = "snake_case", transparent)]
 /// A valid Cosmwasm code that may be stored and transferred
 /// Not indended to be used in external APIs since there is no way to integrate validation on deserialization!
 /// Instead, use [CodeId] in APIs and [Code::try_new] to validate the input.
@@ -20,9 +21,8 @@ pub struct Code {
 
 impl Code {
     pub fn try_new(id: CodeId, querier: &QuerierWrapper<'_>) -> Result<Self> {
-        let raw = WasmQuery::CodeInfo { code_id: id }.into();
         querier
-            .query(&raw)
+            .query(&WasmQuery::CodeInfo { code_id: id }.into())
             .map_err(Error::from)
             .map(|resp: CodeInfoResponse| Self { id: resp.code_id })
     }
@@ -40,7 +40,7 @@ impl From<Code> for CodeId {
 }
 
 pub fn validate_addr(querier: QuerierWrapper<'_>, contract_address: &Addr) -> Result<()> {
-    query_info(querier, contract_address).map(|_| ())
+    query_info(querier, contract_address).map(mem::drop)
 }
 
 pub fn validate_code_id(
--- a/platform/packages/platform/src/contract.rs
+++ b/platform/packages/platform/src/contract.rs
@@ -1,8 +1,43 @@
-use sdk::cosmwasm_std::{Addr, ContractInfoResponse, QuerierWrapper, WasmQuery};
+use sdk::{
+    cosmwasm_std::{Addr, CodeInfoResponse, ContractInfoResponse, QuerierWrapper, WasmQuery},
+    schemars::{self, JsonSchema},
+};
+use serde::{Deserialize, Serialize};
 
 use crate::{error::Error, result::Result};
 
 pub type CodeId = u64;
+
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[serde(transparent)]
+/// A valid Cosmwasm code that may be stored and transferred
+/// Not indended to be used in external APIs since there is no way to integrate validation on deserialization!
+/// Instead, use [CodeId] in APIs and [Code::try_new] to validate the input.
+pub struct Code {
+    id: CodeId,
+}
+
+impl Code {
+    pub fn try_new(id: CodeId, querier: &QuerierWrapper<'_>) -> Result<Self> {
+        let raw = WasmQuery::CodeInfo { code_id: id }.into();
+        querier
+            .query(&raw)
+            .map_err(Error::from)
+            .map(|resp: CodeInfoResponse| Self { id: resp.code_id })
+    }
+
+    #[cfg(any(test, feature = "testing"))]
+    pub const fn unchecked(id: CodeId) -> Self {
+        Self { id }
+    }
+}
+
+impl From<Code> for CodeId {
+    fn from(value: Code) -> Self {
+        value.id
+    }
+}
 
 pub fn validate_addr(querier: QuerierWrapper<'_>, contract_address: &Addr) -> Result<()> {
     query_info(querier, contract_address).map(|_| ())
@@ -11,14 +46,14 @@ pub fn validate_addr(querier: QuerierWrapper<'_>, contract_address: &Addr) -> Re
 pub fn validate_code_id(
     querier: QuerierWrapper<'_>,
     contract_address: &Addr,
-    expected_code_id: CodeId,
+    expected_code: Code,
 ) -> Result<()> {
     query_info(querier, contract_address).and_then(|info| {
-        if info.code_id == expected_code_id {
+        if info.code_id == expected_code.id {
             Ok(())
         } else {
             Err(Error::unexpected_code(
-                expected_code_id,
+                expected_code.into(),
                 contract_address.clone(),
             ))
         }
@@ -38,11 +73,14 @@ fn query_info(
 
 #[cfg(test)]
 pub mod tests {
-    use sdk::cosmwasm_std::{testing::MockQuerier, Addr, QuerierWrapper};
+    use sdk::cosmwasm_std::{self, testing::MockQuerier, Addr, QuerierWrapper};
 
-    use crate::contract::testing::{self, CODE_ID};
+    use crate::contract::{
+        testing::{self, CODE},
+        Code,
+    };
 
-    use super::validate_addr;
+    use super::{validate_addr, CodeId};
 
     #[test]
     fn validate_invalid_addr() {
@@ -69,7 +107,16 @@ pub mod tests {
         let querier = QuerierWrapper::new(&mock_querier);
 
         let address = Addr::unchecked("some address");
-        assert!(super::validate_code_id(querier, &address, CODE_ID).is_ok());
+        assert!(super::validate_code_id(querier, &address, CODE).is_ok());
+    }
+
+    #[test]
+    fn transparent_serde() {
+        let id: CodeId = 13;
+        assert_eq!(
+            Code::unchecked(id),
+            cosmwasm_std::from_json(cosmwasm_std::to_json_string(&id).unwrap()).unwrap()
+        );
     }
 }
 
@@ -80,16 +127,16 @@ pub mod testing {
         WasmQuery,
     };
 
-    use super::CodeId;
+    use super::Code;
 
-    pub const CODE_ID: CodeId = 20;
+    pub const CODE: Code = Code::unchecked(20);
 
     pub fn valid_contract_handler(_query: &WasmQuery) -> QuerierResult {
         SystemResult::Ok(ContractResult::Ok(
             to_json_binary(&{
                 let mut response = ContractInfoResponse::default();
 
-                response.code_id = CODE_ID;
+                response.code_id = CODE.into();
                 response.creator = "some data".into();
 
                 response
--- a/platform/packages/sdk/Cargo.toml
+++ b/platform/packages/sdk/Cargo.toml
@@ -14,7 +14,7 @@ storage = ["dep:cw-storage-plus"]
 testing = [
     "dep:anyhow",
     "dep:cosmwasm-schema",
-    "dep:cw-multi-test",
+    "cw-multi-test/cosmwasm_1_2",
     "dep:data-encoding",
     "dep:serde",
     "dep:thiserror",
--- a/platform/packages/sdk/src/testing/mod.rs
+++ b/platform/packages/sdk/src/testing/mod.rs
@@ -1,7 +1,7 @@
 use cosmwasm_std::{
     testing::{mock_dependencies, MockApi, MockQuerier, MockStorage},
-    Binary, ContractResult, Empty, GovMsg, IbcMsg, IbcQuery, OwnedDeps, SystemError, SystemResult,
-    WasmQuery,
+    CodeInfoResponse, ContractInfoResponse, ContractResult, Empty, GovMsg, HexBinary, IbcMsg,
+    IbcQuery, OwnedDeps, SystemError, SystemResult, WasmQuery,
 };
 use cw_multi_test::{
     BankKeeper, BasicAppBuilder as BasicCwAppBuilder, DistributionKeeper, FailingModule,
@@ -57,10 +57,9 @@ pub fn customized_mock_deps_with_contracts<const N: usize>(
         WasmQuery::ContractInfo { contract_addr }
             if contracts.contains(&contract_addr.as_str()) =>
         {
-            SystemResult::Ok(ContractResult::Ok(Binary(Vec::from(
-                br#"{"code_id":0,"creator":"","admin":null,"pinned":false,"ibc_port":null}"#
-                    as &[u8],
-            ))))
+            SystemResult::Ok(ContractResult::Ok(
+                cosmwasm_std::to_json_binary(&ContractInfoResponse::default()).unwrap(),
+            ))
         }
         WasmQuery::Smart { contract_addr, .. }
         | WasmQuery::Raw { contract_addr, .. }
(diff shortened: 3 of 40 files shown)


protocol/contracts/leaser/src/contract.rs
```diff
@@ -45,10 +48,9 @@ pub fn instantiate(
 
     ContractOwnerAccess::new(deps.storage.deref_mut()).grant_to(&info.sender)?;
 
-    let lease_code = msg.lease_code_id;
-    Config::new(msg).store(deps.storage)?;
-
-    leaser::update_lpp(deps.storage, lease_code.into(), Batch::default())
+    Config::try_new(msg, &deps.querier)
+        .and_then(|config| config.store(deps.storage).map(|()| config))
+        .and_then(|config| leaser::update_lpp(deps.storage, config.lease_code, Batch::default()))
```
KirilMihaylov, line 53: [Suggestion] The two `Result::and_then` calls could be nested to avoid the need for the `Result::map` call.

protocol/contracts/leaser/src/contract.rs
```diff
@@ -199,7 +197,7 @@ fn validate_customer(
 fn validate_lease(lease: Addr, deps: Deps<'_>) -> ContractResult<Addr> {
     Leaser::new(deps)
         .config()
-        .map(|config| config.config.lease_code_id)
+        .map(|config| config.config.lease_code)
         .and_then(|lease_code_id| {
```
KirilMihaylov, line 201: This can be combined into one `Result::and_then` call to avoid complications. `lease_code_id` is used only once either way, so aliasing the value doesn't matter in this case.

protocol/contracts/leaser/src/state/config.rs
```diff
@@ -67,10 +71,10 @@ impl Config {
             .map_err(Into::into)
     }
 
-    pub fn update_lease_code(storage: &mut dyn Storage, new_code: CodeId) -> ContractResult<()> {
+    pub fn update_lease_code(storage: &mut dyn Storage, new_code: Code) -> ContractResult<()> {
         Self::STORAGE
             .update(storage, |mut c| -> ContractResult<Config> {
-                c.lease_code_id = new_code;
+                c.lease_code = new_code;
```
KirilMihaylov, line 77: I'd suggest using functional style update. In my opinion, it could look a bit better especially in cases like the lines above in the `update` method.

```suggestion
                Ok(Self {
                    lease_code: new_code,
                    ..c
                })
```

protocol/contracts/lpp/Cargo.toml
```diff
@@ -27,7 +27,8 @@ currencies = { workspace = true, optional = true }
 currency = { workspace = true }
 finance = { workspace = true }
 lpp-platform = { workspace = true }
-platform = { workspace = true }
+# TODO remove the testing feature once initialize the Lease code on instanciation
```
KirilMihaylov, line 30:
```suggestion
# TODO remove the testing feature once the Lease code is set during instantiation
```

protocol/contracts/lpp/src/contract/rewards.rs
```diff
@@ -112,7 +112,7 @@ mod test {
             deps.as_mut().storage,
             Config::new(
                 TheCurrency::TICKER.into(),
-                1000u64.into(),
+                Code::unchecked(1000u64),
```
KirilMihaylov, line 115: This might be on me for not following it when I came, but I feel like we should stick to one magic numbers for a given purpose.

protocol/contracts/lpp/src/msg.rs
```diff
@@ -36,7 +37,8 @@ where
     Lpns: Group,
 {
     NewLeaseCode {
-        lease_code_id: Uint64,
+        /// This is an internal system API and we use [Code]
```
KirilMihaylov, line 40: Perhaps this should be a normal comment, rather than a doc comment, as it doesn't explain nor give other information about the field.

protocol/contracts/lpp/src/state/config.rs
```diff
@@ -1,19 +1,18 @@
...
-};
+use sdk::{cosmwasm_std::Storage, cw_storage_plus::Item};
 
 use crate::{borrow::InterestRate, error::Result, msg::InstantiateMsg};
 
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub struct Config {
     lpn_ticker: String,
-    lease_code_id: Uint64,
+    #[serde(alias = "lease_code_id")]
```
KirilMihaylov, line 14: Perhaps it'd be good to put a TODO here referencing the next release, after which the alias should be removed.

platform/packages/platform/src/contract.rs
```diff
@@ -1,24 +1,59 @@
...
+};
+use serde::{Deserialize, Serialize};
 
 use crate::{error::Error, result::Result};
 
 pub type CodeId = u64;
 
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[serde(transparent)]
```
KirilMihaylov, line 13: These could be combined.

KirilMihaylov, line 13: We might also like to put `#[repr(transparent)]` to tell the compiler to inherit the inner type's memory layout. In the best case it would provide us with an improvement and in the worst case nothing would change.

protocol/contracts/lpp/src/state/config.rs
```diff
@@ -104,9 +103,10 @@ impl Config {
 impl From<InstantiateMsg> for Config {
     fn from(msg: InstantiateMsg) -> Self {
         // 0 is a non-existing code id
```
KirilMihaylov, line 105: If that really is the case, I feel like `Code` should encapsulate `std::num::NonZeroU64` and `Option<Code>` should be used here. That would provide proper guarantees and there wouldn't be any loss as `Option<NonZeroXYZ>` has the same memory layout as `XYZ` due to special optimizations.

protocol/contracts/reserve/Cargo.toml
```diff
@@ -51,6 +51,6 @@ thiserror = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
+platform = {workspace = true, features = ["testing"]}
 schema = { workspace = true }
-sdk = { workspace = true, features = ["testing"] }
-
+sdk = { workspace = true, features = ["testing"] }
```
KirilMihaylov, line 56:
```suggestion
sdk = { workspace = true, features = ["testing"] }

```

protocol/contracts/reserve/Cargo.toml
```diff
@@ -51,6 +51,6 @@ thiserror = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
+platform = {workspace = true, features = ["testing"]}
 schema = { workspace = true }
```
KirilMihaylov, line 55: [Suggestion]
```suggestion
schema.workspace = true
```

protocol/contracts/reserve/src/api.rs
```diff
@@ -26,9 +26,14 @@ pub struct MigrateMsg {}
...
-    NewLeaseCode { code_id: Uint64 },
+    /// This is an internal system API and we use [Code]
+    NewLeaseCode {
+        code: Code,
+    },
 
-    CoverLiquidationLosses { amount: LpnCoin },
+    CoverLiquidationLosses {
+        amount: LpnCoin,
+    },
```
KirilMihaylov, line 36: [Suggestion] We could simplify the serialized version if we use a a tuple variant.
```suggestion
    CoverLiquidationLosses(LpnCoin),
```
This would produce `{"cover_liquidation_losses": "1000unls"}` in JSON instead of nested objects.

protocol/contracts/reserve/src/contract.rs
```diff
@@ -65,24 +69,21 @@ pub fn execute(
...
-        }
+        ExecuteMsg::NewLeaseCode { code } => SingleUserAccess::new(
+            deps.storage.deref_mut(),
+            crate::access_control::LEASE_CODE_ADMIN_KEY,
+        )
+        .check(&info.sender)
+        .map_err(Into::into)
+        .and_then(|()| {
+            Config::update_lease_code(deps.storage, code).map(|()| PlatformResponse::default())
+        }),
```
KirilMihaylov, line 80:
```suggestion
        .and_then(|()| Config::update_lease_code(deps.storage, code))
        .map(|()| PlatformResponse::default()),
```

protocol/contracts/reserve/src/state/config.rs
```diff
@@ -1,29 +1,28 @@
...
+    lease_code: Code,
 }
 
 impl Config {
     const STORAGE: Item<'static, Self> = Item::new("config");
 
-    #[cfg(test)]
-    pub const fn new(lease_code_id: CodeId) -> Self {
-        Self { lease_code_id }
+    pub fn new(lease_code: Code) -> Self {
```
KirilMihaylov, line 20: When it doesn't cost us anything, it's nicer to have it as `const fn` so we don't have to make massive changes later. ;)
```suggestion
    pub const fn new(lease_code: Code) -> Self {
```

protocol/contracts/reserve/src/state/config.rs
```diff
@@ -34,40 +33,32 @@ impl Config {
...
     use cosmwasm_std::{testing::MockStorage, Storage};
-    use platform::contract::CodeId;
+    use platform::contract::{Code, CodeId};
 
     use crate::Config;
 
     #[test]
     fn store_load() {
-        let lease_code_id = 12;
+        let lease_code = Code::unchecked(12);
```
KirilMihaylov, line 52: I'd like to reiterate on the magic numbers, as the others at least were `>= 1000`.

KirilMihaylov (changes requested):
Mostly minor changes with the exception of the comment about usage of `NonZeroU64` if the said statement holds true.

platform/packages/platform/src/contract.rs
```diff
@@ -1,24 +1,59 @@
...
+}
+
+impl From<Code> for CodeId {
+    fn from(value: Code) -> Self {
+        value.id
+    }
+}
+
 pub fn validate_addr(querier: QuerierWrapper<'_>, contract_address: &Addr) -> Result<()> {
     query_info(querier, contract_address).map(|_| ())
```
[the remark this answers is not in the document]
gmanev7: not sure that `mem::drop` would be easier to comprehend than `|_| ()`

platform/packages/sdk/Cargo.toml
```diff
@@ -14,7 +14,7 @@ storage = ["dep:cw-storage-plus"]
 testing = [
     "dep:anyhow",
     "dep:cosmwasm-schema",
-    "dep:cw-multi-test",
```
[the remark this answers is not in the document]
gmanev7: cargo check complains with:
feature `testing` includes `dep:cw-multi-test/cosmwasm_1_2` with both `dep:` and `/`
  To fix this, remove the `dep:` prefix.
  If the intent is to avoid creating an implicit feature `cw-multi-test` for an optional dependency, then consider replacing this with two values:
      "dep:cw-multi-test", "cw-multi-test/cosmwasm_1_2"

platform/packages/platform/src/contract.rs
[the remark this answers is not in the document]
KirilMihaylov: As I said, it's less cryptic as it's English, not just symbols. In my personal opinion, it is easier as on first pass I usually focus on reading the variables and functions.

While it's in the `std`'s prelude for as an artifact from the 2015 edition, it's still useful for such reasons. Another thing is, I wrote `drop` and not `mem::drop` for that reason.

platform/packages/sdk/src/testing/mod.rs
```diff
@@ -57,10 +57,9 @@ pub fn customized_mock_deps_with_contracts<const N: usize>(
         WasmQuery::ContractInfo { contract_addr }
             if contracts.contains(&contract_addr.as_str()) =>
         {
-            SystemResult::Ok(ContractResult::Ok(Binary(Vec::from(
-                br#"{"code_id":0,"creator":"","admin":null,"pinned":false,"ibc_port":null}"#
-                    as &[u8],
-            ))))
+            SystemResult::Ok(ContractResult::Ok(
+                cosmwasm_std::to_json_binary(&ContractInfoResponse::default()).unwrap(),
+            ))
```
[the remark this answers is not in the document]
gmanev7: the `unwrap` call is not to the instance created with `ContractResult::Ok`. It unwraps the `to_json_binary`

protocol/contracts/leaser/src/contract.rs
gmanev7: prefer to keep the functional expressions as shallow as possible, up to some complexity extent for sure :-)

platform/packages/sdk/Cargo.toml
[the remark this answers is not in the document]
KirilMihaylov: That's strange because the whole CI passes on my fork. Granted I ran only `clippy`, but that is really strange. I'll check my fork now with `cargo check`.

platform/packages/sdk/src/testing/mod.rs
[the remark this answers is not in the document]
KirilMihaylov: While I agree, it still has to fail but this time it wouldn't with a panic. I think that the less panics we have when we can avoid them easily, the better.

protocol/contracts/leaser/src/contract.rs
gmanev7: indeed it can be but prefer to keep it separate. it's similar to the question how big a function should be

platform/packages/sdk/Cargo.toml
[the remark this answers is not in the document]
KirilMihaylov: I have this line in the `sdk` package and `cargo check` doesn't complain. Not even when I pass `--all-features`.

`cosmos_ibc = ["dep:cosmos-sdk-proto", "cosmos", "cosmwasm-std/ibc3", "cosmos-sdk-proto/cosmwasm"]`

protocol/contracts/leaser/src/state/config.rs
gmanev7: this is not in the scope of this PR, but I'll accept this suggestion, moreover prefer immutability.

protocol/contracts/lpp/src/contract/rewards.rs
gmanev7: do not get what you say

platform/packages/platform/src/contract.rs
gmanev7: if there is no reason,would prefer not to tell the compiler how to represent it better in the memory

protocol/contracts/lpp/src/state/config.rs
gmanev7: this is a logical error, and I'm going to fix it right after this PR is merged. The lease code is going to be provided at contract instantiation.

gmanev7: Re: NonZero - no guarantees that it really holds true

platform/packages/platform/src/contract.rs
KirilMihaylov: There is a reason, and it's optimization. If `NonZero*` type is plugged inside it can help when it's used inside an `Option`. Instead of being two "words" it would be just one. It's something we'd generally want on single field/single non-zero-type field structs.

protocol/contracts/reserve/src/state/config.rs
gmanev7: I do not understand the problem here - it's just a test that should get some input data.

protocol/contracts/lpp/src/contract/rewards.rs
gmanev7: this is a test code, after all

protocol/contracts/leaser/src/state/config.rs
```diff
@@ -61,21 +62,25 @@ impl Config {
         lease_due_period: Duration,
     ) -> ContractResult<()> {
         Self::STORAGE
-            .update(storage, |mut c| -> ContractResult<Config> {
-                c.lease_interest_rate_margin = lease_interest_rate_margin;
-                c.lease_position_spec = lease_position_spec;
-                c.lease_due_period = lease_due_period;
-                Ok(c)
+            .update(storage, |c| -> ContractResult<Config> {
```
KirilMihaylov, line 65: I thought you were pushing against explicit types. ;)
```suggestion
            .update(storage, |c| {
```

protocol/contracts/leaser/src/state/config.rs
```diff
@@ -61,21 +62,25 @@ impl Config {
...
+                Ok(Self {
+                    lease_interest_rate_margin,
+                    lease_position_spec,
+                    lease_due_period,
+                    ..c
+                })
             })
             .map(|_| ())
```
KirilMihaylov, line 73: I didn't write it for every single instance, but if you'd be accepting of using `drop`, it might be good to be consistent.

protocol/contracts/reserve/src/state/config.rs
KirilMihaylov: My point is that, the black box integration tests /`./tests`/ for example uses the normal numbering because it goes through real API. In this case, we're adding magic numbers which if not hard to distinguish from other components, could be brushed off. That's the reason when I use magic numbers, I use `0xDEADC0DE`. It's enormous and there isn't a way to mistake it.

The other tests at least used larger integers.

platform/packages/platform/src/contract.rs
gmanev7: I follow the best practices on optimizations and do not do premature ones.
Just to note here that we are talking about `#repr` and not about NonZero*

protocol/contracts/leaser/src/state/config.rs
gmanev7: right, this code is not in the scope of the PR ... but I am going to fix it

protocol/contracts/leaser/src/state/config.rs
gmanev7: hahaha
expanding the scope of this PR may go endless

platform/packages/platform/src/contract.rs
KirilMihaylov: They are closely related if you think about it. Even if we want to call it a premature optimization, it is a zero cost one. The worst case is that we'd add a new non-zero-sized field, which is highly unlikely for such specialized type, that would require removing it.

protocol/contracts/leaser/src/state/config.rs
KirilMihaylov: True, yet it was just a suggestion after all. We could've just left it as a future work. I didn't really know how much it would extend the scope of the PR.

PR #313, opened 2024-03-20
feat: Extend currency definitions & add definitions query

### LINKED ISSUE #228: [Oracle] Provide info for the currencies pertaining to the protocol
The use case to enable is the following: The App obtains the protocols from the Admin, incl. the oracles. Then query them to get the info subject of this issue.

We need the (ticker, symbol@Nolus, symbol@DexNetwork, decimals/exponent) of each of the Lpn, lease currencies, and extra payment currencies.

Payment currencies = union of the Lpn, lease currencies, and the extra payment currencies (if any, for example, Nls).

The ticker must be the same as the one in `currencies.json` at the Dex network. The only exception, cause it needs a migration, is USDC at Osmosis.

KirilMihaylov:
Note: Implementation only lists currently configured currencies rather than all software supported.

Closes #228

### DIFF
--- a/platform/packages/currency/src/lib.rs
+++ b/platform/packages/currency/src/lib.rs
@@ -40,6 +40,12 @@ pub trait Currency: Copy + Ord + Default + Debug + 'static {
 
     /// Symbol at the Dex network
     const DEX_SYMBOL: SymbolStatic;
+
+    /// Exponent on which the whole unit was raised to get the currency's base
+    /// unit represented by the trait.
+    ///
+    /// Example: `(10 ^ DECIMAL_DIGITS) uUSDC = 1 USDC`
+    const DECIMAL_DIGITS: u8;
 }
 
 pub fn equal<C1, C2>() -> bool
--- a/protocol/contracts/oracle/src/api/contract.rs
+++ b/protocol/contracts/oracle/src/api/contract.rs
@@ -76,8 +76,10 @@ pub enum QueryMsg {
     Price {
         currency: SymbolOwned,
     },
-    // returns a list of supported denom pairs
+    /// Lists configured swap pairs
     SupportedCurrencyPairs {},
+    /// Lists configured currencies
+    Currencies {},
     /// Provides a path in the swap tree between two arbitrary currencies
     ///
     /// Returns `oracle::api::swap::SwapPath`
@@ -112,6 +114,44 @@ pub enum ExecuteAlarmMsg {
 pub struct DispatchAlarmsResponse(pub AlarmsCount);
 
 pub type SupportedCurrencyPairsResponse = Vec<SwapLeg>;
+
+pub type CurrenciesResponse = Vec<Currency>;
+
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "testing"), derive(Debug, PartialEq, Eq))]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct Currency {
+    pub ticker: String,
+    pub bank_symbol: String,
+    pub dex_symbol: String,
+    pub decimal_digits: u8,
+    pub group: CurrencyGroup,
+}
+
+impl Currency {
+    pub fn new<C>(group: CurrencyGroup) -> Self
+    where
+        C: currency::Currency,
+    {
+        Self {
+            ticker: C::TICKER.into(),
+            bank_symbol: C::BANK_SYMBOL.into(),
+            dex_symbol: C::DEX_SYMBOL.into(),
+            decimal_digits: C::DECIMAL_DIGITS,
+            group,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "testing"), derive(Debug, PartialEq, Eq))]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum CurrencyGroup {
+    Native,
+    Lpn,
+    Lease,
+    PaymentOnly,
+}
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, JsonSchema)]
 #[cfg_attr(any(test, feature = "testing"), derive(Debug))]
--- a/protocol/contracts/oracle/src/state/supported_pairs.rs
+++ b/protocol/contracts/oracle/src/state/supported_pairs.rs
@@ -230,7 +230,7 @@ impl AnyVisitor for CurrencyVisitor {
 mod tests {
     use std::cmp::Ordering;
 
-    use currencies::test::StableC;
+    use currencies::test::{LeaseC1, LeaseC2, NativeC, StableC};
     use currency::Currency;
     use sdk::cosmwasm_std::{self, testing};
     use tree::HumanReadableTree;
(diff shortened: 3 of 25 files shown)


platform/packages/currency/src/lib.rs
```diff
@@ -40,6 +40,12 @@ pub trait Currency: Copy + Ord + Default + Debug + 'static {
 
     /// Symbol at the Dex network
     const DEX_SYMBOL: SymbolStatic;
+
+    /// Exponent on which the whole unit was raised to get the currency's base
+    /// unit represented by the trait.
+    ///
+    /// Example: `(10 * DECIMAL_PLACES) uUSDC = 1 USDC`
+    const DECIMAL_PLACES: u8;
```
gmanev7, line 48: [everywhere] "places" is not so suitable, please rename to "digits"

protocol/contracts/oracle/src/api/contract.rs
```diff
@@ -113,6 +115,17 @@ pub struct DispatchAlarmsResponse(pub AlarmsCount);
 
 pub type SupportedCurrencyPairsResponse = Vec<SwapLeg>;
 
+pub type CurrencyDefinitionsResponse = Vec<CurrencyDefinition>;
+
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct CurrencyDefinition {
```
gmanev7, line 122: -> Currency

protocol/contracts/oracle/src/api/contract.rs
```diff
@@ -76,8 +76,10 @@ pub enum QueryMsg {
     Price {
         currency: SymbolOwned,
     },
-    // returns a list of supported denom pairs
+    /// Lists configured swap pairs
     SupportedCurrencyPairs {},
+    /// Lists configured currency definitions
+    CurrencyDefinitions {},
```
gmanev7, line 82: -> Currencies

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -178,6 +189,26 @@ where
...
+struct DefinitionVisitor;
+
+impl AnyVisitor for DefinitionVisitor {
+    type Output = CurrencyDefinition;
+
+    type Error = platform::never::Never;
+
+    fn on<C>(self) -> AnyVisitorResult<Self>
+    where
+        C: Currency + Serialize + DeserializeOwned,
```
gmanev7, line 201: Are you sure we need `Serialize + Deserialize`

protocol/contracts/oracle/src/api/contract.rs
```diff
@@ -113,6 +115,17 @@ pub struct DispatchAlarmsResponse(pub AlarmsCount);
...
+pub type CurrencyDefinitionsResponse = Vec<CurrencyDefinition>;
+
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct CurrencyDefinition {
+    pub ticker: String,
+    pub bank_symbol: String,
+    pub dex_symbol: String,
+    pub decimal_places: u8,
+}
```
gmanev7, line 127: we also need to know the group it directly is a member of. They are Native, Lpns, Lease, and PaymentOnlyGroup

protocol/packages/currencies/src/lease/osmosis.rs
```diff
@@ -274,7 +274,7 @@ define_symbol! {
         },
     }
 }
-define_currency!(StkAtom, STK_ATOM);
+define_currency!(StkAtom, STK_ATOM, 8);
```
gmanev7, line 277: please double-check it is eight! The definition in currencies.json is 6

protocol/contracts/oracle/src/state/supported_pairs.rs
KirilMihaylov: It came with the signature as far as I am aware. Most likely the IDE generated the method for me, thus it's already a bound on the trait. I can remove it here, but it won't make a big difference overall.

platform/packages/currency/src/lib.rs
```diff
@@ -40,6 +40,12 @@ pub trait Currency: Copy + Ord + Default + Debug + 'static {
 
     /// Symbol at the Dex network
     const DEX_SYMBOL: SymbolStatic;
+
+    /// Exponent on which the whole unit was raised to get the currency's base
+    /// unit represented by the trait.
+    ///
+    /// Example: `(10 * DECIMAL_DIGITS) uUSDC = 1 USDC`
```
gmanev7, line 47: * -> ^

protocol/contracts/oracle/src/api/contract.rs
```diff
@@ -76,8 +76,10 @@ pub enum QueryMsg {
     Price {
         currency: SymbolOwned,
     },
-    // returns a list of supported denom pairs
+    /// Lists configured swap pairs
     SupportedCurrencyPairs {},
+    /// Lists configured currency definitions
```
gmanev7, line 81: currency definitions -> currencies

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -160,6 +160,38 @@ where
...
+                        )
+                    })
+                    .or_else(|_| {
+                        Tickers.maybe_visit_any::<currencies::LeaseGroup, _>(
+                            &node.value().target,
+                            CurrencyVisitor(api::CurrencyGroup::Lease),
+                        )
+                    })
+                    .or_else(|_| {
+                        Tickers.maybe_visit_any::<currencies::PaymentGroup, _>(
```
gmanev7, line 185: currencies::PaymentOnly

protocol/contracts/oracle/src/state/supported_pairs.rs
gmanev7: how big the difference is we cannot measure but for sure we have to limit the required bounds to their minimum

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -160,6 +160,38 @@ where
             })
     }
 
+    pub fn currency_definitions(&self) -> impl Iterator<Item = api::Currency> + '_ {
```
gmanev7, line 163: this function deserves at least a few unit tests

protocol/contracts/oracle/src/state/supported_pairs.rs
KirilMihaylov: I'll remove it from here, but it's the trait that needs to be reworked is what I meant. ;)

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -178,11 +211,32 @@ where
...
+    where
+        C: Currency,
+    {
+        Ok(api::Currency {
+            ticker: C::TICKER.into(),
+            bank_symbol: C::BANK_SYMBOL.into(),
+            dex_symbol: C::DEX_SYMBOL.into(),
+            decimal_digits: C::DECIMAL_DIGITS,
+            group: self.0,
+        })
```
gmanev7, line 231: define this as a fn of api::Currency and reuse in the tests below

gmanev7 (approved):

PR #317, opened 2024-03-25
refactor: Refactor `dex` api

KirilMihaylov:


protocol/packages/dex/src/impl_/out_remote.rs
```diff
@@ -670,12 +670,12 @@ mod impl_contract {
...
-        SwapTask: SwapTaskT
-            + ContractInSwap<TransferOutState, <SwapTask as SwapTaskT>::StateResponse>
-            + ContractInSwap<SwapState, <SwapTask as SwapTaskT>::StateResponse>,
+        OpenIca: Contract,
+        SwapTask: SwapTaskT<StateResponse = OpenIca::StateResponse>
+            + ContractInSwap<TransferOutState, StateResponse = OpenIca::StateResponse>
+            + ContractInSwap<SwapState, StateResponse = OpenIca::StateResponse>,
     {
-        type StateResponse = SwapTask::StateResponse;
+        type StateResponse = OpenIca::StateResponse;
```
gmanev7, line 678: why does not apply `X as Y` here?

protocol/packages/dex/src/impl_/out_local.rs
```diff
@@ -769,13 +769,18 @@ mod impl_contract {
...
+                TransferInInitState,
+                StateResponse = <SwapTask as SwapTaskT>::StateResponse,
+            > + ContractInSwap<
+                TransferInFinishState,
+                StateResponse = <SwapTask as SwapTaskT>::StateResponse,
+            >,
         ForwardToInnerMsg: ForwardToInner,
     {
-        type StateResponse = SwapTask::StateResponse;
+        type StateResponse = <SwapTask as SwapTaskT>::StateResponse;
```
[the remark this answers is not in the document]
KirilMihaylov: Because it becomes ambiguous when the generic parameter becomes an associated type. That can be "mitigated"/worked around if the two associated types have different names.

protocol/packages/dex/src/impl_/out_remote.rs
KirilMihaylov: Because I used `OpenIca` instead of `SwapTask` to avoid it. The reason this is equivalent is that `SwapTask` is bound to have the same `StateResponse`.

gmanev7: my question was not about SwapTask -> OpenIca, but of not applying the disambiguation technique we discussed above, i.e. `<SwapTask as SwapTaskT>::StateResponse`

gmanev7 (approved):

PR #321, opened 2024-03-26
chore: Lints

KirilMihaylov:
At this stage, all newly tracked lint issues are allowed on per-item basis as they require medium to big refactoring.
Said refactoring shall be done in a future PR.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,4 +4,5 @@
 /.dockerignore
 **/rust-toolchain
 **/rust-toolchain.toml
+**/clippy.toml
 **/rustfmt.toml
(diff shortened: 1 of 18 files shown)


gmanev7 (commented):
is it possible to move the allow attributes from a function to a statement level to reduce their scope?

KirilMihaylov:
I tried but it didn't seem to work at least for the `unwrap-in-result` one. I haven't tried putting just `unwrap-used` on the statements while `unwrap-in-result` stays as a function-level attribute.

KirilMihaylov:
The only place where I couldn't move `clippy::unwrap_used` was for the `debug_assert{|_eq|ne}!`s. All other are moved down.

gmanev7 (approved):

PR #322, opened 2024-03-27
fix(admin): Migrate contract storage properly

KirilMihaylov:


platform/contracts/admin/src/endpoints.rs
```diff
@@ -22,8 +22,9 @@ use crate::{
 };
 
 // version info for migration info
-// const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 0;
-const CONTRACT_STORAGE_VERSION: VersionSegment = 1;
+#[cfg(feature = "migrate")]
```
gmanev7, line 25: I believe the package won't be built without the "migrate" feature, ever. Suggest removing it as we did for all the other contracts.

PR #323, opened 2024-03-27
feat(lease): migrate v8 -> v9

gmanev7:


protocol/contracts/lease/src/api/open/mod.rs
```diff
@@ -136,6 +138,9 @@ pub struct PositionSpecDTO {
     pub min_asset: LpnCoinDTO,
     /// The minimum amount to liquidate or close. Any attempt to liquidate a smaller
     /// amount would be postponed until the amount goes above this limit
+    // TODO remove past v0.6.0 migration. Necessary to migrate nolus1suz0vsqe8c8anckaer98awhqs8r4hu7wsm8a49acdl39x6ylfypsqywxwh and
```
KirilMihaylov, line 141: The current version of the contract is v0.9. Is this a typo?

protocol/contracts/lease/src/contract/state/opened/close/customer_close/full.rs
```diff
@@ -22,7 +22,8 @@ use crate::{
 };
 
 type Spec = FullClose;
-pub(super) type RepayableImpl = Close<Spec>;
+pub(in crate::contract::state::opened::close) type RepayableImpl = Close<Spec>;
```
KirilMihaylov, line 25: [multiple] I think this should work fine and would be a bit easier to read.
```suggestion
pub(in super::super) type RepayableImpl = Close<Spec>;
```

protocol/contracts/lease/src/contract/state/paid/v8.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+impl SwapTask for TransferIn {
+    type OutG = LeaseGroup;
+    type Label = Type;
+    type StateResponse = ContractResult<StateResponse>;
+    type Result = SwapResult;
+
+    fn label(&self) -> Self::Label {
+        unreachable!()
```
KirilMihaylov, line 64: Just as a side note, this heavily violates OO principles, specifically the Liskov substitution principle (L in SOLID).

While this is just a mock/stub used for migration, it might be good to rethink the trait bounds or at least the places where they are put. This should of course happen as future work.

protocol/contracts/lease/src/contract/endpoins.rs
```diff
@@ -17,10 +17,14 @@ use crate::{
...
-use super::state::{self, Response, State};
+use super::state::{
+    self,
+    v8::{Migrate, State as State_v8},
+    Response, State,
+};
 
-// const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 7;
-const CONTRACT_STORAGE_VERSION: VersionSegment = 8;
+const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 8;
```
KirilMihaylov, line 26: Maybe introducing a `migration` feature like in admin contract and others could prove beneficial here. This won't require us to keep commented code. :)
```suggestion
#[cfg(feature = "migration")]
const CONTRACT_STORAGE_VERSION_FROM: VersionSegment = 8;
```

KirilMihaylov (commented):
I haven't looked deep into it. The comments are mostly minor and a side note for future work.

protocol/contracts/lease/src/contract/state/opened/close/customer_close/full.rs
gmanev7: there is no such thing as `super::super`. You cannot have anything after the first `super`.  The Rust compiler complains with a `Syntax error: expected an item`

protocol/contracts/lease/src/contract/state/paid/v8.rs
gmanev7: I have tried to go without this implementation but my effort ended up stuck on a bigger refactoring related to relaxing trait bounds ...  that is out of the scope of this PR.
In addition, as a true believer in these principles, I do not have any concerns providing such impl-s as they are required only technically for satisfying the compiler's requirements and are *not* used in practice.

protocol/contracts/lease/src/contract/state/opened/close/customer_close/full.rs
KirilMihaylov: This works perfectly fine when the module `s` is not at the root level (lib.rs/main.rs). I'm guessing you missed the `in ` prefix.
image

gmanev7: thanks for the remark, that was the case

PR #324, opened 2024-03-28
refactor(versioning): Use non-serde type for output

KirilMihaylov:


platform/packages/versioning/src/lib.rs
```diff
@@ -118,21 +118,26 @@ where
         .map_err(map_error)
 }
 
+pub struct SoftwareAndStorageMigrationOutput<MigrateStorageOutput> {
```
gmanev7, line 121: -> FullUpdateOutput?

gmanev7 (approved):

PR #327, opened 2024-04-02
feat(oracle)!: Add stable currency ticker field

KirilMihaylov:


protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -212,6 +218,59 @@ where
...
+        .binary_search(&stable_currency)
+        .is_err()
+    {
+        return Err(ContractError::StableCurrencyNotInTree {});
+    }
+
+    if (0..supported_currencies.len() - 1)
+        .any(|index| supported_currencies[index] == supported_currencies[index + 1])
+    {
+        return Err(ContractError::DuplicatedNodes {});
```
gmanev7, line 251: let's follow the single exit function point idiom and refrain from exiting in the middle of the function. I'd suggest merging all if-statements.

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -30,49 +32,14 @@ where
...
-            }
-        }
-
-        for swap in self.tree.iter() {
-            Tickers.visit_any::<PaymentGroup, _>(&swap.value().target, TickerChecker)?;
-        }
-
-        Ok(self)
+    pub fn new(tree: Tree, stable_currency: SymbolOwned) -> Result<Self, ContractError> {
+        check_tree_tickers(&tree, B::TICKER, &stable_currency)
```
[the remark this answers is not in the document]
KirilMihaylov: I'm thinking of renaming it to `check_tree` as it has little-to-nothing to do with the actual tickers.

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -234,26 +232,22 @@ fn check_tree(
         {
             Err(ContractError::DuplicatedNodes {})
         } else {
-            Ok(())
+            tree.iter().try_for_each(|swap| {
+                Tickers.visit_any::<PaymentGroup, _>(&swap.value().target, EmptyVisitor)
```
gmanev7, line 236: please reuse `fn currency::validate<G>(ticker)`

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -234,26 +232,22 @@ fn check_tree(
         {
             Err(ContractError::DuplicatedNodes {})
         } else {
-            Ok(())
+            tree.iter().try_for_each(|swap| {
```
gmanev7, line 235: what about switching to `Iterator::all` - it sounds closer to the intent here, ah?

protocol/contracts/oracle/src/state/supported_pairs.rs
```diff
@@ -376,6 +376,28 @@ mod tests {
         SupportedPairs::<TheCurrency>::new(tree, TheCurrency::TICKER.into()).unwrap();
     }
 
+    #[test]
+    fn test_not_included_stable_currency() {
```
gmanev7, line 387: `unknown_stable_currency`

protocol/contracts/oracle/src/state/supported_pairs.rs
KirilMihaylov: `Iterator::all` doesn't support `Result`s, only `bool`s. While it might be more suitable for duplication check, I won't be able to use it here. At most I might be able to use `Iterator::find`.

gmanev7: no, then let's continue with `try_for_each`

gmanev7 (approved):

PR #328, opened 2024-04-03
refactor(oracle): Remove `OracleBase` & variations generic parameters

KirilMihaylov:


gmanev7 (commented):
It was not a good idea to spread the dependency to a particular currency or group down to any entity/type that needs them. It limits the chance for the code to evolve.

PR #329, opened 2024-04-05
fix(currencies): DEX-side IBC denominators

KirilMihaylov:


gmanev7 (approved):
looks good
Double-verified with the info from the Slack messages

PR #330, opened 2024-04-05
refactor(admin)!: Reduce migration granularity

### LINKED ISSUE #291: [Admin] Specify protocols before the migration specification of individual contracts
Currently, the focus is on a specific contract, and the protocol is on a second level. A few drawbacks:
- it is difficult to ensure all contracts of a protocol are migrated
- it is difficult to state and validate that all protocols are migrated

We want to flip them.

In addition, the migration API should accept the migration of some or all of the protocols, and (optionally) the platform. The some/all protocols choice should be explicitly stated.

KirilMihaylov:
Closes #291

### DIFF
--- a/platform/contracts/admin/src/contracts/mod.rs
+++ b/platform/contracts/admin/src/contracts/mod.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 
 use ::platform::contract::CodeId;
 use sdk::{
-    cosmwasm_std::Addr,
+    cosmwasm_std::{Addr, Uint64},
     schemars::{self, JsonSchema},
 };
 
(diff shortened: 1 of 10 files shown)


platform/contracts/admin/src/msg/mod.rs
```diff
@@ -53,9 +59,29 @@ pub enum ExecuteMsg {
     EndOfMigration {},
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct MigrateLeases
+where
+    Uint64: Into<CodeId>,
+    CodeId: Into<Uint64>,
```
gmanev7, line 66: what's the reason of specifying this?

platform/contracts/admin/src/msg/mod.rs
```diff
@@ -70,20 +96,25 @@ pub enum SudoMsg {
     /// it should start as Admin contract migration which would then
     /// continue with the migration of the other contracts.
     MigrateContracts(MigrateContracts),
+    MigrateLeases(BTreeMap<String, MigrateLeases>),
+    MigrateLeasesCont(BTreeMap<String, MigrateLeasesCont>),
```
gmanev7, line 100: not a good idea due to:
- the introduced knowledge for Leases in the Admin
- broken migration atomicity

platform/contracts/admin/src/msg/mod.rs
KirilMihaylov: Static assertions that the types are indeed interchangeable.

platform/contracts/admin/src/contracts/mod.rs
```diff
@@ -21,22 +21,58 @@ mod impl_mod;
 mod platform;
 mod protocol;
 
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
-#[serde(rename_all = "snake_case", deny_unknown_fields)]
-pub struct ContractsTemplate<T, U = ProtocolTemplate<BTreeMap<String, T>>> {
-    pub platform: PlatformTemplate<T>,
-    pub protocol: U,
+pub trait Hkt {
```
gmanev7, line 24: what does Hkt stand for?

KirilMihaylov: Higher-kind type. Another name for it is higher-order type. I can rename it to not use the abbreviation, if you think it'd be easier to understand.

gmanev7: this abbreviature is not so widely known, at least from my point of view, what short-and-descriptive names would you come with?

platform/contracts/admin/src/contracts/mod.rs
```diff
@@ -21,22 +21,58 @@ mod impl_mod;
...
+        deserialize = "OutmostHkt::Hkt<PlatformTemplate<PlatformUnit>>: Deserialize<'de>, \
+            OutmostHkt::Hkt<Protocol>: Deserialize<'de>",
+    )
+)]
+#[schemars(bound = "OutmostHkt: JsonSchema, \
+    OutmostHkt::Hkt<PlatformTemplate<PlatformUnit>>: JsonSchema, \
+    OutmostHkt::Hkt<Protocol>: JsonSchema, \
+    PlatformUnit: JsonSchema, \
+    Protocol: JsonSchema")]
+pub struct ContractsTemplate<OutmostHkt: Hkt, PlatformUnit, Protocol> {
```
gmanev7, line 58: please use `where ...` notation

platform/contracts/admin/src/contracts/mod.rs
KirilMihaylov: `TypeConstructor` is one I can think of as a replacement.

gmanev7 (approved):

PR #331, opened 2024-04-09
feat(admin): Add DEX-describing data fields

KirilMihaylov:


platform/contracts/admin/src/contracts/protocol/mod.rs
```diff
@@ -22,5 +22,7 @@ pub struct ProtocolTemplate<T> {
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub struct Protocol {
     pub network: String,
+    pub dex_type: String,
+    pub dex_metadata: String,
```
gmanev7, line 26: we've discussed alternative naming, please use it

platform/contracts/admin/src/contracts/protocol/mod.rs
```diff
@@ -22,5 +22,7 @@ pub struct ProtocolTemplate<T> {
 #[serde(rename_all = "snake_case", deny_unknown_fields)]
 pub struct Protocol {
     pub network: String,
+    pub dex_type: String,
```
gmanev7, line 25: using `String`s here is not adequate .... we have to restrict to the supported ones

platform/contracts/admin/src/endpoints.rs
```diff
@@ -49,13 +48,12 @@ pub fn instantiate(
...
 }
 
-#[cfg(feature = "migrate")]
 #[entry_point]
 pub fn migrate(
     deps: DepsMut<'_>,
     env: Env,
     MigrateMsg {
-        reserve_contracts,
+        dex_data,
```
gmanev7, line 56: do you believe that all existing protocols share the same dex_data? Or I am mistaken here?

gmanev7: disregard it, just saw the type

platform/contracts/admin/src/msg.rs
```diff
@@ -89,6 +87,13 @@ pub enum QueryMsg {
     Protocol(String),
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct DexData {
+    pub dex_type: String,
+    pub dex_params: Option<String>,
```
gmanev7, line 94: please remove the prefix `dex_`

platform/contracts/admin/src/contracts/protocol/mod.rs
```diff
@@ -21,6 +21,57 @@ pub struct ProtocolTemplate<T> {
...
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(
+    rename_all = "PascalCase",
+    rename_all_fields = "snake_case",
+    deny_unknown_fields
+)]
+pub enum Network {
+    #[serde(alias = "NEUTRON")]
+    Neutron,
+    #[serde(alias = "OSMOSIS")]
```
gmanev7, line 38: what's the idea of having aliases? They do allow supporting a second name, but only on deserialization.

platform/contracts/admin/src/contracts/protocol/mod.rs
```diff
@@ -21,6 +21,57 @@ pub struct ProtocolTemplate<T> {
...
-    pub network: String,
+    pub network: Network,
+    pub dex: Dex,
     pub contracts: ProtocolTemplate<Addr>,
 }
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(
+    rename_all = "PascalCase",
+    rename_all_fields = "snake_case",
```
gmanev7, line 32: as per the [doc](https://serde.rs/container-attrs.html) these two do the same when applied on enums, as it is in our case ..... and you have specified two different styles which would confuse the reader

platform/contracts/admin/src/contracts/protocol/mod.rs
```diff
@@ -21,6 +21,57 @@ pub struct ProtocolTemplate<T> {
...
+    #[serde(alias = "NEUTRON")]
+    Neutron,
+    #[serde(alias = "OSMOSIS")]
+    Osmosis,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
+#[serde(
+    rename_all = "PascalCase",
+    rename_all_fields = "snake_case",
```
gmanev7, line 45: same here

platform/contracts/admin/src/contracts/mod.rs
```diff
@@ -32,7 +32,7 @@ pub type ContractsMigration = ContractsTemplate<Option<MigrationSpec>>;
 
 pub type ContractsPostMigrationExecute = ContractsTemplate<Option<String>>;
 
-pub(crate) type ContractsGroupedByProtocol = ContractsTemplate<Addr, BTreeMap<String, Protocol>>;
+pub type ContractsGroupedByProtocol = ContractsTemplate<Addr, BTreeMap<String, Protocol>>;
```
gmanev7, line 35: who is gonna use it outside the crate?

platform/contracts/admin/src/state/contracts.rs
```diff
@@ -76,12 +74,9 @@ pub(crate) fn load_all(storage: &dyn Storage) -> Result<ContractsGroupedByProtoc
     })
 }
 
-#[cfg(feature = "migrate")]
 pub(super) fn migrate_protocols(
     storage: &mut dyn Storage,
-    api: &dyn Api,
-    querier: QuerierWrapper<'_>,
-    mut reserve_contracts: BTreeMap<String, String>,
+    mut dex_types: BTreeMap<String, Dex>,
```
gmanev7, line 79: [globally] please rename `dex_types` to `dexes`

platform/contracts/admin/src/state/contracts.rs
```diff
@@ -91,56 +86,30 @@ pub(super) fn migrate_protocols(
...
-    pub network: String,
-    pub contracts: OldProtocolTemplate<Addr>,
+    pub network: Network,
+    pub contracts: ProtocolTemplate<Addr>,
 }
 
-#[cfg(feature = "migrate")]
 impl OldProtocol {
-    fn convert(self, reserve: Addr) -> Protocol {
+    fn convert(self, dex: Dex) -> Protocol {
```
gmanev7, line 108: `convert` -> `migrate`

platform/contracts/admin/src/contracts/protocol/mod.rs
KirilMihaylov: I put them specifically for this migration. Later on we'll remove them, but right now we need them in order to automatically convert them from `SCREAMING_SNAKE_CASE` to `PascalCase`.

KirilMihaylov: I deemed that testing well-known `serde` attributes is beyond our scope. I had put tests on the mixed casing as we haven't used that before and it is after all mixing them, that behaviors is not explicitly documented.

gmanev7: OK, then put a TODO around to not forget their purpose and when supposed to be removed

platform/contracts/admin/src/contracts/mod.rs
KirilMihaylov: It is used outside, in the `InstantiateMsg`, if I recall correctly. The compiler was livid about it.

platform/contracts/admin/src/contracts/protocol/mod.rs
KirilMihaylov: I have a test covering it. It absolutely does NOT do the same thing when they are combined. I've applied them both in order to ensure consistency of the casing of the fields themselves.

gmanev7: my bad reading through the doc

gmanev7 (approved):

PR #345, opened 2024-05-27
Handle overflow

maneva3:


protocol/contracts/lpp/src/lpp.rs
```diff
@@ -891,6 +891,135 @@ mod test {
...
+        let loan1_annual_interest_rate = Percent::from_permille(137);
+        let mut loan1 = Loan {
+            principal_due: borrow_loan1,
+            annual_interest_rate: loan1_annual_interest_rate,
+            interest_paid: env.block.time,
+        };
+
+        Loan::open(&mut deps.storage, lease_addr_1.clone(), &loan1).unwrap();
+
+        lpp.total
```
gmanev7, line 936: what interface do we test against? why do we break the encapsulation reaching to `lpp.total`?

protocol/contracts/lpp/src/lpp.rs
```diff
@@ -891,6 +891,135 @@ mod test {
         assert_eq!(withdraw, Coin::new(1110));
     }
 
+    #[test]
+    fn open_and_repay_with_overflow() {
```
gmanev7, line 895: This is not a unit test - it is an unnecessarily broad functional one that does not focus on the test case. Instead, it deals, unnecessarily, with persistence, configuration, access rights, and even with a whole liquidity pool. 
We need to test only Total!
Its length - that smells a lot!

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -81,14 +81,24 @@ where
     ) -> Result<&Self, ContractError> {
         self.total_interest_due = self.total_interest_due_by_now(&ctime);
 
+        let updated_total_principal_due =
```
gmanev7, line 84: the naming is not correct - the amount is not yet "updated"
replacing it just with "new_total" would tell us more

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -81,14 +81,24 @@ where
...
+
         // TODO: get rid of fully qualified syntax
         self.annual_interest_rate = Rational::new(
             Fraction::<Coin<Lpn>>::of(&self.annual_interest_rate, self.total_principal_due)
-                + loan_interest_rate.of(amount),
-            self.total_principal_due + amount,
+                .checked_add(loan_interest_rate.of(amount))
+                .ok_or(ContractError::OverflowError(
+                    "Annual interest rate calculation overflow".to_string(),
+                ))?,
```
gmanev7, line 97: the code may become clearer if you extract the nominator as a separate variable - and name it appropriately - that is the most important! The goal is to end up with code that is easy to read and comprehend

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +112,21 @@ where
         loan_principal_payment: Coin<Lpn>,
         loan_interest_rate: Percent,
     ) -> &Self {
-        self.total_interest_due = self.total_interest_due_by_now(&ctime) - loan_interest_payment;
+        self.total_interest_due = self
+            .total_interest_due_by_now(&ctime)
+            .saturating_sub(loan_interest_payment);
```
gmanev7, line 117: the use of arithmetics that changes the value deserves at least a comment on why that is safe

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +112,21 @@ where
...
                 Fraction::<Coin<Lpn>>::of(&self.annual_interest_rate, self.total_principal_due)
                     - loan_interest_rate.of(loan_principal_payment),
                 self.total_principal_due - loan_principal_payment,
-            )
-        };
-
-        self.total_principal_due -= loan_principal_payment;
+            );
+            self.total_principal_due -= loan_principal_payment;
+        }
```
gmanev7, line 129: do not support this rewrite of the logic from functional to procedural style

let's keep the functional style intact

hint: you complicate the logic around  self.total_principal_due

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +112,21 @@ where
...
         loan_interest_rate: Percent,
     ) -> &Self {
-        self.total_interest_due = self.total_interest_due_by_now(&ctime) - loan_interest_payment;
+        self.total_interest_due = self
+            .total_interest_due_by_now(&ctime)
+            .saturating_sub(loan_interest_payment);
 
-        self.annual_interest_rate = if self.total_principal_due == loan_principal_payment {
-            zero_interest_rate()
+        if loan_principal_payment >= self.total_principal_due {
```
gmanev7, line 119: when reading code it's always easier to comprehend < or <= than > or >= because the ordering of the arguments match their total order

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +112,21 @@ where
...
+            self.annual_interest_rate = zero_interest_rate();
+            self.total_principal_due = Coin::ZERO;
         } else {
-            Rational::new(
+            self.annual_interest_rate = Rational::new(
                 Fraction::<Coin<Lpn>>::of(&self.annual_interest_rate, self.total_principal_due)
                     - loan_interest_rate.of(loan_principal_payment),
                 self.total_principal_due - loan_principal_payment,
```
gmanev7, line 126: the code here deviates too much from the corresponding code at `borrow` - haven't we missed anything?

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +110,25 @@ where
         loan_principal_payment: Coin<Lpn>,
         loan_interest_rate: Percent,
     ) -> &Self {
-        self.total_interest_due = self.total_interest_due_by_now(&ctime) - loan_interest_payment;
+        // Due to rounding in the calculations, there is a possibility of overflow in the arithmetic operations.
```
gmanev7, line 113: the comment should shed light on:
rounding of what exact calculation?
why it's safe to saturate to 0?

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -171,4 +187,74 @@ mod test {
         let interest_due = total.total_interest_due_by_now(&env.block.time);
         assert_eq!(interest_due, 500u128.into());
     }
+
+    #[test]
+    fn borrow_and_repay_with_overflow() {
+        let mut env = testing::mock_env();
+        env.block.time = Timestamp::from_nanos(0);
```
gmanev7, line 194: why do you use a mocked Env? just to pass a Timestamp? Too complex

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -81,14 +81,22 @@ where
...
         // TODO: get rid of fully qualified syntax
-        self.annual_interest_rate = Rational::new(
+        let annual_interest =
             Fraction::<Coin<Lpn>>::of(&self.annual_interest_rate, self.total_principal_due)
-                + loan_interest_rate.of(amount),
-            self.total_principal_due + amount,
-        );
+                .checked_add(loan_interest_rate.of(amount))
+                .ok_or(ContractError::OverflowError(
+                    "Annual interest rate calculation overflow",
```
gmanev7, line 94: "rate" is not correct to stay here

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -81,14 +81,22 @@ where
...
         self.total_interest_due = self.total_interest_due_by_now(&ctime);
 
+        let new_total_principal_due = self
+            .total_principal_due
+            .checked_add(amount)
+            .ok_or(ContractError::OverflowError("Total principal due overflow"))?;
+
         // TODO: get rid of fully qualified syntax
-        self.annual_interest_rate = Rational::new(
+        let annual_interest =
```
gmanev7, line 90: for the sake of uniformity, let's prefix the variable name with "new_"

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +110,27 @@ where
...
                 self.total_principal_due - loan_principal_payment,
             )
+        } else {
+            zero_interest_rate()
         };
 
-        self.total_principal_due -= loan_principal_payment;
+        self.total_principal_due = self
+            .total_principal_due
+            .saturating_sub(loan_principal_payment);
```
gmanev7, line 133: here we do not have any rounding that may cause an overflow

It is a bug if we get an overflow. Therefore we should panic, not error, if that happens.
Hint: Use `checked_sub` and `expect` to panic with a custom message explaining the case

protocol/contracts/lpp/src/state/total.rs
```diff
@@ -102,19 +110,27 @@ where
...
+            .saturating_sub(loan_interest_payment);
+
+        self.annual_interest_rate = if loan_principal_payment < self.total_principal_due {
             Rational::new(
                 Fraction::<Coin<Lpn>>::of(&self.annual_interest_rate, self.total_principal_due)
                     - loan_interest_rate.of(loan_principal_payment),
                 self.total_principal_due - loan_principal_payment,
             )
+        } else {
+            zero_interest_rate()
```
gmanev7, line 128: It would get to this `else` if loan_principal_payment >= self.total_principal_due. These are two cases, equal and greater-than. The `else` should only be done on the former case while the latter should not get into it.

Please do not insert another `if-else`.

gmanev7 (approved):

PR #347, opened 2024-06-05
fix(oracle): migrate the config

gmanev7:


protocol/contracts/oracle/src/state/config.rs
```diff
@@ -35,4 +36,32 @@ impl Config {
...
+            #[allow(dead_code)]
+            base_asset: SymbolOwned,
+            price_config: PriceConfig,
+        }
+        impl Serialize for OldConfig {
+            fn serialize<S>(&self, _serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: Serializer,
+            {
+                unreachable!("satisfy Item::load requirements")
```
KirilMihaylov, line 58: [Note] For reference, the documentation of [`Result::expect`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.expect) \(in the end\) hints towards saying what it is expected to be.  
In this case, perhaps leaving it as a comment on the `impl` explaining why it was done might be better, while having the `unreachable!` macro be replaced with `unimplemented!` explaining the type shouldn't be serialized.

While essentially both `unimplemented!` and `unreachable!` are just a `panic!` underneath, they do imply different things to the reader.

protocol/contracts/oracle/src/error.rs
```diff
@@ -37,6 +37,9 @@ pub enum ContractError {
     #[error("[Oracle] Failed to store configuration! Cause: {0}")]
     StoreConfig(StdError),
 
+    #[error("[Oracle] Failed to load the v0 configuration! Cause: {0}")]
+    LoadConfigV0(StdError),
```
KirilMihaylov, line 41: [Minor] Adding a `TODO` comment might be good, just to indicate that it's not needed after deployment.

protocol/contracts/oracle/src/contract/exec.rs
```diff
@@ -66,10 +66,8 @@ where
...
 {
-    let config = Config::load(storage).map_err(ContractError::LoadConfig)?;
-    let oracle = Feeds::<G, BaseC, QuoteG>::with(config.price_config);
-
-    oracle
-        .feed_prices(storage, block_time, &sender, &prices)
+    Config::load(storage)
+        .map(|cfg| Feeds::<G, BaseC, QuoteG>::with(cfg.price_config))
+        .and_then(|oracle| oracle.feed_prices(storage, block_time, &sender, &prices))
         .map(|()| Default::default())
```
KirilMihaylov, line 72: The mapping is `() -> ()`. Should be removed.
```suggestion
```

KirilMihaylov (approved):
Looks good.  
Only one suggestion, which might be optimized away anyway, a minor remark and a note.

protocol/contracts/oracle/src/state/config.rs
gmanev7: maybe I am missing something, but what is the relation between `expect` and `unimplemented!` except they panic under specific circumstances

KirilMihaylov: Well, both signify state that is for one way or another unsatisfied or unwanted.  
I referred to it as an example because it gave hints on better practices for error messages. That's all.

PR #349, opened 2024-06-07
feat(leaser): close protocol, API only

gmanev7:


protocol/contracts/leaser/src/msg.rs
```diff
@@ -79,6 +84,20 @@ pub enum SudoMsg {
         lease_position_spec: PositionSpecDTO,
         lease_due_period: Duration,
     },
+    CloseProtocol {
+        migration_spec: <HigherOrderGranularity<HigherOrderProtocolContracts, HigherOrderOption> as HigherOrderType>::Of<MigrationSpec>,
```
KirilMihaylov, line 88: I'd say that it should rather be: `ProtocolContracts<MigrationSpec>`, as all contracts should be migrated into their respective blank counterparts, rather than just some.

protocol/contracts/leaser/src/msg.rs
```diff
@@ -79,6 +84,20 @@ pub enum SudoMsg {
...
+    }
+}
+
+#[derive(Serialize, Deserialize, Default, Clone, PartialEq, Eq, JsonSchema)]
+#[cfg_attr(any(test, feature = "testing"), derive(Debug))]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum ForceClose {
+    #[default]
+    No,
+    KillProtocol,
```
KirilMihaylov, line 100: With this suggestion the API wouldn't change overall whilst still not including this variant on any other networks than the explicitly opt-in.
```suggestion
    #[cfg(any(feature = "net_dev", feature = "net_test"))]
    KillProtocol,
```

protocol/contracts/leaser/src/msg.rs
gmanev7: thanks, nice simplification .. and a better API

tests/src/common/test_case/builder.rs
```diff
@@ -272,10 +269,13 @@ where
...
             test_case.address_book.lpp().clone(),
-            test_case.address_book.time_alarms().clone(),
-            test_case.address_book.oracle().clone(),
+            (
+                test_case.address_book.time_alarms().clone(),
+                test_case.address_book.oracle().clone(),
+            ),
             test_case.address_book.profit().clone(),
             test_case.address_book.reserve().clone(),
+            test_case.address_book.protocols_registry().clone(),
```
KirilMihaylov, line 278: If we need so many parameters, it might be better to introduce a `struct` containing all addresses, rather than packing them in a tuple. At the very least it's really hard mix them up.

PR #350, opened 2024-06-07
feat(admin): Add protocol deregistration

KirilMihaylov:


platform/contracts/admin/src/endpoints.rs
```diff
@@ -209,7 +212,7 @@ pub fn query(deps: Deps<'_>, env: Env, msg: QueryMsg) -> ContractResult<Binary>
 
             let addr = deps.api.addr_humanize(&canonical_addr)?;
 
-            sdk::cosmwasm_std::to_json_binary(&addr).map_err(From::from)
+            to_json_binary(&addr).map_err(From::from)
```
gmanev7, line 215:
```suggestion
            cosmwasm_std::to_json_binary(&addr).map_err(From::from)
```

gmanev7: following the practice of referring functions by their package

platform/contracts/admin/src/endpoints.rs
```diff
@@ -273,6 +276,24 @@ fn register_protocol(
...
+) -> ContractResult<CwResponse> {
+    for name in state_contracts::protocols(storage)? {
+        let protocol = state_contracts::load_protocol(storage, name.clone())?.contracts;
+
+        if protocol.leaser == sender {
+            return Ok(response::response_only_messages(
+                protocol.migrate_standalone(migration_spec),
+            ));
+        }
+    }
```
gmanev7, line 292: I would suggest you refactor this in a functional style, a.k.a `findmap`
After all most of the other code, at least the recently modified, follows that style

gmanev7 (commented):
Overall looks good to me! Thanks for refactoring some aspects of the code .. and adding more abstractions. 
A few suggestions follow.

platform/contracts/admin/src/endpoints.rs
KirilMihaylov: That's a good idea. At first I searched for a `try_find` kind of method but I might have seen it in third-party packages extending iterators. I didn't think of `find_map`. Thanks!

platform/contracts/admin/src/endpoints.rs
KirilMihaylov: I was it was imported and didn't give it a second thought. Corrected now.

platform/contracts/admin/src/endpoints.rs
```diff
@@ -282,17 +282,21 @@ fn deregister_protocol(
...
+            match state_contracts::load_protocol(storage, name.clone())
+                .map(|protocol| protocol.contracts)
+            {
+                Ok(protocol) => (protocol.leaser == sender).then(|| {
+                    Ok(response::response_only_messages(
+                        protocol.migrate_standalone(migration_spec),
+                    ))
+                }),
+                Err(error) => Some(Err(error)),
+            }
```
gmanev7, line 297: I believe `.map_err(...).map(...)` would be more concise than `match`

gmanev7 (approved):

PR #355, opened 2024-06-17
fix(admin): Add expected void release label

KirilMihaylov:


platform/packages/versioning/src/release.rs
```diff
@@ -12,6 +12,12 @@ use super::{Version, VersionSegment};
 #[serde(transparent)]
 pub struct ReleaseLabel(String);
 
+impl ReleaseLabel {
+    pub fn void() -> Self {
+        Self("void".into())
```
gmanev7, line 17: please define "void" as a constant, similar to the way we do for "RELEASE_LABEL_DEV"

platform/packages/versioning/src/release.rs
```diff
@@ -12,6 +12,12 @@ use super::{Version, VersionSegment};
 #[serde(transparent)]
 pub struct ReleaseLabel(String);
 
+impl ReleaseLabel {
+    pub fn void() -> Self {
```
gmanev7, line 16: please make it consistent with `fn label()`

gmanev7 (approved):

PR #358, opened 2024-06-17
refactor(versioning): use factory member fn-s

gmanev7:


platform/packages/versioning/src/release.rs
```diff
@@ -21,7 +21,15 @@ impl ReleaseLabel {
...
-    const VOID_CONTRACT: &'static str = "void-contract";
+    const VOID_RELEASE: &'static str = "void-release";
+
+    pub fn label() -> Self {
+        Self(Self::RELEASE_LABEL.into())
+    }
+
+    pub fn void() -> Self {
+        Self(Self::VOID_RELEASE.into())
+    }
```
KirilMihaylov, line 32:
```suggestion
    pub fn void() -> Self {
        Self(Self::VOID_RELEASE.into())
    }

    pub(crate) fn label() -> Self {
        Self(Self::RELEASE_LABEL.into())
    }
```

KirilMihaylov (changes requested):

KirilMihaylov (approved):

PR #359, opened 2024-06-18
feat(leaser): migrate outstanding leases to a new code

gmanev7:


protocol/contracts/leaser/src/contract.rs
```diff
@@ -229,8 +240,15 @@ fn protocols_registry_load(storage: &dyn Storage) -> ContractResult<Addr> {
...
+where
+    C: Into<CodeId>,
+{
+    Code::try_new(new_code_id.into(), &querier).map_err(Into::into)
+}
+
 fn migrate_msg() -> impl Fn(Addr) -> LeaseMigrateMsg {
-    move |_customer| LeaseMigrateMsg {}
+    |_customer| LeaseMigrateMsg {}
 }
```
KirilMihaylov, line 252: I feel like we might have went a bit too far with this. 😂
If not inline the lambda, at least lets make this function just drop the argument /use `_: Addr`/ and return the message instead of a lambda.

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -121,28 +121,36 @@ where
     })
 }
 
-pub(super) fn try_close_protocol<ProtocolsRegistryLoader>(
-    storage: &dyn Storage,
+pub(super) fn try_close_protocol<ProtocolsRegistryLoader, MsgFactory>(
+    storage: &mut dyn Storage,
+    new_lease_code: Code,
+    max_leases: MaxLeases,
```
KirilMihaylov, line 127: Why is this a parameter to the function? As we discussed, the case where not all leases are migrated is invalid, thus rendering this parameter redundant.

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -121,28 +121,36 @@ where
...
+    migrate_msg: MsgFactory,
     protocols_registry: ProtocolsRegistryLoader,
     migration_spec: ProtocolContracts<MigrationSpec>,
     force: ForceClose,
 ) -> ContractResult<MessageResponse>
 where
+    MsgFactory: Fn(Addr) -> MigrateMsg,
     ProtocolsRegistryLoader: FnOnce(&dyn Storage) -> ContractResult<Addr>,
 {
     if force == ForceClose::No && has_lease(storage) {
```
KirilMihaylov, line 137: [Clippy; -W pedantic] Prefer using `matches!` when dealing with enums. This also removes the need for implementing `(Partial|)Eq`.
```suggestion
    if matches!(force, ForceClose::No) && has_lease(storage) {
```

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -185,46 +193,121 @@ fn emit_status(next_customer: Option<Addr>) -> Emitter {
...
+            migrate_msg(),
             |_| Ok(Addr::unchecked("Registry")),
             dummy_spec(),
-            ForceClose::No
+            ForceClose::No,
         )
-        .is_ok());
+        .unwrap();
+        let cw_resp = response::response_only_messages(resp);
+        let update_lpp_update_reserve_delete_protocol = 1 + 1 + 1;
```
KirilMihaylov, line 231: ???
If there is meaning implied by the two additions, I feel like writing `3` and just explaining where it comes from with a comment would make sense as I don't understand it myself like this.
```suggestion
        let update_lpp_update_reserve_delete_protocol = 3;
```

protocol/contracts/leaser/src/leaser.rs
KirilMihaylov, line 137: [Secure coding] Prefer matching against the safer case implicitly and explicitly against the more dangerous ones. That way when a new variant is added, if it happens for us to forget to update it, it will still go the safe route.  
Another way to avoid it is to use exhaustive pattern matching, but it might not be the perfect solution for this case.

In this case, the safer thing here is to ask "is it forced? if not, are there any more leases? if not, close".  
One way might be the following suggestion and inverting the `if` and `else` bodies:
```suggestion
    if matches!(force, ForceClose::KillProtocol) || !has_lease(storage) {
```

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -185,46 +193,121 @@ fn emit_status(next_customer: Option<Addr>) -> Emitter {
...
     use sdk::cosmwasm_std::testing::MockStorage;
 
-    use crate::{msg::ForceClose, state::leases::Leases, ContractError};
+    use crate::{
+        msg::{Config, ForceClose, InstantiateMsg, MaxLeases},
+        state::leases::Leases,
+        ContractError,
+    };
+
+    const MAX_LEASES: MaxLeases = 100_000;
```
KirilMihaylov, line 211: Is redundant when we keep in mind the above comment about the parameter.

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -241,4 +324,8 @@ mod test {
             reserve: migration_spec,
         }
     }
+
+    fn migrate_msg() -> impl Fn(Addr) -> MigrateMsg {
+        |_customer| MigrateMsg {}
+    }
```
KirilMihaylov, line 330: As mentioned above, it's going to be more compact and understandable with less jumping around if it's inlined.

protocol/contracts/leaser/src/msg.rs
```diff
@@ -84,7 +84,14 @@ pub enum SudoMsg {
         lease_due_period: Duration,
     },
     CloseProtocol {
+        // Since this is an external system API we should not use [Code].
+        new_lease_code_id: Uint64,
         migration_spec: ProtocolContracts<MigrationSpec>,
+        /// If `ForceClose::KillProtocol` closes the protocol even if it has not closed leases
```
KirilMihaylov, line 90: `If`? I'm guessing this happened due to rewording it later.

KirilMihaylov (changes requested):

protocol/contracts/leaser/src/contract.rs
gmanev7: A lambda has been used since we needed the migration message to be customized. I'd prefer to keep it as such to make further migrations easier.

protocol/contracts/leaser/src/leaser.rs
gmanev7: for consistency with the other functions
the parameter receives Uxx::MAX otherwise.

protocol/contracts/leaser/src/leaser.rs
gmanev7: hm, interesting, could you point me to the clippy item for more details?

protocol/contracts/leaser/src/leaser.rs
gmanev7: avoiding docs if the code may explain the stuff is what we want
here the name of the variable and its value convey the meaning

protocol/contracts/leaser/src/leaser.rs
gmanev7: Re: "In this case, the safer thing here is to ask "is it forced? if not, are there any more leases? if not, close"." This is what I do.
The rationale here is to use easier-to-read code and avoid negative checks.

gmanev7: in addition, that is covered with tests that validate the forcing logic

protocol/contracts/leaser/src/msg.rs
gmanev7: nice catch, thanks!

protocol/contracts/leaser/src/leaser.rs
gmanev7: refactored it in another way to address what you say and to avoid lease migration when force=no

protocol/contracts/leaser/src/leaser.rs
gmanev7: frankly, I do not understand what you mean

protocol/contracts/leaser/src/leaser.rs
KirilMihaylov: With all due respect, that's not a solution. The moment a 4th contract appears are we going to append it? A variable name that is longer than 2-3 words usually is not properly named. There are cases where it's unavoidable, but this isn't it. Why are we "scared" to write comments?

protocol/contracts/leaser/src/leaser.rs
KirilMihaylov: It appears I was thinking of a different lint, so I'll leave it to your discretion.  
[`clippy::match_like_matches_macro`](https://rust-lang.github.io/rust-clippy/master/#/match_like_matches_macro)

protocol/contracts/leaser/src/contract.rs
KirilMihaylov: Unless the lambda needs to capture context, becoming a closure, it is unnecessary and way less readable to do it this way.

protocol/contracts/leaser/src/leaser.rs
gmanev7: I am not "scared", the practice is one to prefer explanatory code than documentation.

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -134,24 +134,30 @@ where
...
+        ForceClose::KillProtocol => {
+            try_migrate_leases(storage, new_lease_code, max_leases, migrate_msg)
+        }
+        ForceClose::No => {
+            if has_lease(storage) {
+                Err(ContractError::ProtocolStillInUse())
+            } else {
+                Ok(MessageResponse::default())
+            }
+        }
```
KirilMihaylov, line 147:
```suggestion
        ForceClose::No if has_lease(storage) => Err(ContractError::ProtocolStillInUse()),
        ForceClose::No => Ok(MessageResponse::default()),
```

PR #367, opened 2024-07-15
feat(tools/graph): Add a graph package containing an acyclic double-edged graph implementation

KirilMihaylov:


tools/graph/src/acyclic/double_edged/insert.rs
```diff
@@ -0,0 +1,311 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /// # Safety
+    /// The index of the first vertex, `vertex_a`, has to be less than the
+    /// index of the second vertex, `vertex_b`.
```
gmanev7, line 137: this is something that deserves a debug_assert!

gmanev7 (commented):
At first sight, some abstractions are missing. Indication to that is the relation between some attributes you have described in the "Safety" doc sections.  They are USELESS unless implemented as debug_assert-s. 

I strongly recommend defining types that capture these data and enforcing those relations as their invariant at instance creation.

Other comments would arise once address the above one.

PR #370, opened 2024-07-17
Overflow on fraction of duration

maneva3:


platform/packages/finance/src/fractionable/coin.rs
```diff
@@ -37,27 +37,33 @@ mod test {
     };
 
     #[test]
-    fn safe_mul() {
+    fn checked_mul() {
         use crate::fractionable::Fractionable;
         assert_eq!(
             Coin::<SuperGroupTestC1>::new(30),
-            Coin::<SuperGroupTestC1>::new(3).safe_mul(&Percent::from_percent(1000))
+            Fractionable::<u32>::checked_mul(
```
gmanev7, line 44: [everywhere] not sure we need the extended syntax when calling `checked_mul`. More concise and readable is if we call it on the first operand

platform/packages/finance/src/fractionable/duration.rs
```diff
@@ -14,15 +14,13 @@ where
...
     where
         F: Ratio<Coin<C>>,
     {
-        let d128: u128 = self.into();
         // TODO re-assess the design of Ratio ... and whether it could be > 1
-        d128.safe_mul(fraction)
-            .try_into()
-            .expect("overflow computing a fraction of duration")
+        let d128: u128 = self.into();
+        Fractionable::<Coin<C>>::checked_mul(d128, fraction).and_then(|units| units.try_into().ok())
```
gmanev7, line 23: again, too generic naming - `units` - here we have `may_product`

platform/packages/finance/src/fractionable/mod.rs
```diff
@@ -41,20 +42,19 @@ where
...
-            self
+            Some(self)
         } else {
             let res_double: D = self.into() * ratio.parts().into();
             let res_double = res_double / ratio.total().into();
-            let res_intermediate: DIntermediate =
-                res_double.try_into().expect("unexpected overflow");
-            res_intermediate.into()
+            let res_intermediate: Option<DIntermediate> = res_double.try_into().ok();
+            res_intermediate.map(|val| val.into())
```
gmanev7, line 57: replace the  lambda function with the into function, e.g. `Into::into`

platform/packages/finance/src/fractionable/percent.rs
```diff
@@ -29,17 +29,14 @@ where
...
     {
         let p128: u128 = self.units().into();
         // TODO re-assess the design of Ratio ... and whether it could be > 1
-        let res: Units = p128
-            .safe_mul(fraction)
-            .try_into()
-            .expect("overflow computing a fraction of permille");
-        Self::from_permille(res)
+        Fractionable::<Coin<C>>::checked_mul(p128, fraction)
+            .and_then(|units| units.try_into().ok().map(Self::from_permille))
```
gmanev7, line 39: same 2 flaws as above - generic naming of `units` and extended call syntax

platform/packages/finance/src/price/mod.rs
```diff
@@ -174,9 +175,14 @@ where
     fn lossy_add(self, rhs: Self) -> Option<Self> {
         const FACTOR: Amount = 1_000_000_000_000_000_000; // 1*10^18
         let factored_amount = FACTOR.into();
```
gmanev7, line 177: the `fn Coin<C>::new()`  is const so please switch to using `Coin<C>` for `FACTOR`

platform/packages/finance/src/price/mod.rs
```diff
@@ -174,9 +175,14 @@ where
...
-            total(factored_amount, self).checked_add(total(factored_amount, rhs));
-        may_factored_total.map(|factored_total| total_of(factored_amount).is(factored_total))
+        total(factored_amount, self)
+            .ok()
+            .and_then(|total_self| {
+                total(factored_amount, rhs)
+                    .ok()
+                    .and_then(|total_rhs| total_self.checked_add(total_rhs))
+            })
+            .map(|factored_total| total_of(factored_amount).is(factored_total))
```
gmanev7, line 185: we complicate the expression by transforming the values from Result to Option. Is it impossible to stay and compute while having `Result` and later call `'ok()`?

gmanev7 (commented):
reviewed only finance without liability and interest modules
once address the findings , on the next iteration I'll continue with the rest

platform/packages/finance/src/fraction.rs
```diff
@@ -1,7 +1,9 @@
+use std::fmt::Display;
+
 use crate::fractionable::Fractionable;
 
 pub trait Fraction<U> {
-    fn of<A>(&self, whole: A) -> A
+    fn of<A>(&self, whole: A) -> Option<A>
     where
-        A: Fractionable<U>;
+        A: Fractionable<U> + Display + Clone;
```
gmanev7, line 8: we should not require `Display + Clone`

gmanev7: what is the motivation of that?

PR #372, opened 2024-07-30
feat(tools/topology): feat(tools/topology): Add a topology package for parsing structure and deriving currencies.

KirilMihaylov:


tools/topology/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+            .iter()
+            .try_for_each(|(ticker, currency)| {
+                self.resolve_currency(dex_network, &channels, &host_to_dex_path, ticker, currency)
+                    .map(|currency| currencies.push(currency))
+            })
+            .map(|()| currencies.into_boxed_slice())
+            .map_err(Into::into)
+    }
+
+    fn host_to_dex_path<'r>(
```
gmanev7, line 54: refuse to review functions longer than 20 loc, this one is more than 100!

tools/topology/src/symbol.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.path.push('/');
+    }
+
+    fn add_symbol(mut self, symbol: &str) -> Symbol {
+        self.path.push_str(symbol);
+
+        // "ibc/" + 64 hexadecimal digits
+        let mut symbol = String::new();
+
+        symbol.reserve_exact(68);
```
gmanev7, line 84: replace the constant 68 with a const statement that evaluates the size of "ibc/" + the size of the sha256 output.

tools/topology/src/symbol.rs
```diff
@@ -0,0 +1,106 @@
...
+            .flat_map(|byte| [byte >> 4, byte & 0xF])
+            .map(|half_byte| {
+                if half_byte < 10 {
+                    b'0' + half_byte
+                } else {
+                    b'A' + (half_byte - 10)
+                }
+            })
+            .map(char::from)
+            .for_each(|ch| symbol.push(ch));
```
gmanev7, line 99: this deserves to be tested ... with well known data

tools/topology/src/symbol.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        Symbol {
+            path: self.path.into_boxed_str(),
+            symbol: symbol.into_boxed_str(),
```
gmanev7, line 103: Why do we complicate the stuff working with `Box<str>` when we own the strings? In addition, `into_boxed_str` may reallocate and copy the raw bytes.

tools/topology/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+                unreachable!()
+            };
+
+            walked_endpoints.push(HostToDexPathChannel {
+                endpoint,
+                inverse_endpoint,
+            });
+
+            if next_network == dex_network {
+                break 'find_connection walked_endpoints;
```
gmanev7, line 148: [everywhere] Please refrain from using loop labels - they make the code hard to read and indicate the presence of hidden inline functions - please refactor by extracting code into functions and employing single function return best practice.

gmanev7 (commented):
not able to continue and finish the review due to the issues

platform/Cargo.toml
```diff
@@ -1,6 +1,10 @@
 [workspace]
-members = ["**/src/.."]
-exclude = ["target/**"]
+members = [
+    "contracts/*/src/..",
+    "packages/*/src/..",
+    "tests",
+]
+exclude = ["target"]
```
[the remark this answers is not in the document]
KirilMihaylov: I must have overlooked it. Thanks!

tools/topology/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "topology"
+version = "0.1.0"
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+
+[package.metadata.cargo-each]
+combinations = [
+    { tags = ["ci", "@agnostic"], include-rest = true }
+]
```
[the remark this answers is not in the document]
KirilMihaylov: We do and we should. Upholding quality standards is a must for me.

tools/topology/src/lib.rs
KirilMihaylov: The reason I did it like this was that there isn't an easy way to not waste time and resources on a `.clone()` call. I'll try to tackle the problem a bit more.

tools/topology/src/symbol.rs
KirilMihaylov: I'll comply but in my opinion it's a waste to do something at runtime that is known not only at compile time but also by specification.

tools/topology/src/symbol.rs
KirilMihaylov: The integration tests already test this against well-known data. I used the real topology of the dev-net.

If it's about having **unit** tests, okay but please say it like that. Don't beat around the bush, please.

tools/topology/src/symbol.rs
KirilMihaylov: It drops the excess capacity. While yes, it could reallocate, we won't be leaking memory that won't ever be needed as we won't even need to modify the string.

Please take take a look at this video: https://youtu.be/A4cKi7PTJSs. I'm not using `Rc<str>` or `Arc<str>` since probably it would be an overkill and I know your opinion on it, but `Box<str>` is sufficient. We don't clone this in the crate, nor do we modify it, so it's a nice compromise between the two.

tools/topology/src/lib.rs
KirilMihaylov: That's a fair criticism about the >100LoC, but I just want to be clear that not everything can be fit into <20LoC without passing 7+ parameters. 😉

tools/topology/Cargo.toml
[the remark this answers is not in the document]
gmanev7: let's reiterate why cargo-each was introduced .... it seems someone has missed the initial idea and what we agreed upon

tools/topology/src/lib.rs
gmanev7: no comment

tools/topology/src/symbol.rs
gmanev7: let's read carefully - I have asked you to replace it with `a const statement` ....

gmanev7: I am not wasting my resources if the comments are not read in their details

tools/topology/src/symbol.rs
gmanev7: I do not think a programmer have to be advised for such a basic stuff ... obviously you want it, OK

The best testing here is with **UNIT tests** - is that clear now?

tools/topology/src/symbol.rs
gmanev7: simplicity is always preferred than obscurity ... and I can send you 10x materials on that matter

tools/topology/Cargo.toml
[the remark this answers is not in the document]
KirilMihaylov: I'll ignore the passive-aggressive connotation and just say that UNTIL it goes away, we can't do much except not check, lint, run tests, etc. for said package.

If it were a random utility that I wrote purely for convenience, I'd still not agree but would be way more willing to compromise. Please, keep in mind that this will be fundamental part of the protocols, so not running tests is not desirable.

tools/topology/src/symbol.rs
KirilMihaylov: My bad. I must have not read it that carefully when I read it the first time. I apologize.

tools/topology/src/symbol.rs
KirilMihaylov: What you are saying is that if we were to use e.g. Java, you'd leave the whole application use `StringBuilder`. I hope you realize how fanatical that sounds.  
That is the literal definition of using the wrong tool for the job and I'm dying on this hill.

Would a type alias satisfy your requirement for simplicity and readability?

gmanev7 (commented):
This version is easier to read, though there are two flows:
- the code logic is data-driven, a lot of raw data structures are passed around, BtreeMap-s for example. This indicates missing abstractions, and
- a few parameters are used as output params of functions

tools/topology/src/lib.rs
```diff
@@ -0,0 +1,450 @@
...
+                dex_network,
+                &mut endpoints_deque,
+                &mut traversed_networks,
+                network,
+                endpoints,
+                &mut walked_channels,
+            ) {
+                break Ok(path);
+            }
+        }
```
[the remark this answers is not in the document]
KirilMihaylov: I'll look into `Iterator::scan` because it rings a bell but I haven't used it or at least don't recall how it works.

tools/topology/Cargo.toml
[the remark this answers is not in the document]
KirilMihaylov: Again, might not be possible because of how the DEX selection works right now. Moreover, it will be here to stay until I also make the generation of the sources too.

I know what the end goal is, but removing this is still out of reach for now.

tools/topology/src/inner_structure/channel.rs
```diff
@@ -0,0 +1,15 @@
...
+pub(crate) struct Channel {
+    pub a: ChannelEndpoint,
+    pub b: ChannelEndpoint,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub(crate) struct ChannelEndpoint {
+    pub network: String,
+    pub ch: String,
```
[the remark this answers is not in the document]
KirilMihaylov: I'd like to clarify, do you mean to make a type alias, `type T = U;`, or use the _newtype pattern_, `struct T(U)`?

tools/topology/src/lib.rs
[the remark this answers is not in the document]
gmanev7: Have you looked into `scan`?

gmanev7:
A recommendation: Print the currencies by groups to engage the end-user in double-checking. It's essential to print the path as well.

tools/topology/src/lib.rs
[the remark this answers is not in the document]
KirilMihaylov: I looked into the possibility but even in combination with `flat_map` for example, it would still have converted it to depth-first. Marking as resolved now.

gmanev7 (approved):

PR #379, opened 2024-08-15
fix(currencies): Add back tickers to currencies

KirilMihaylov:


protocol/packages/currencies/src/currency_macro.rs
```diff
@@ -8,9 +8,9 @@ pub use currency::{CurrencyDTO, CurrencyDef, Definition};
 macro_rules! define_currency {
     (
         $ident:ident,
-        $ticker: literal,
-        $bank_symbol: literal,
-        $dex_symbol: literal,
+        $ticker:expr,
```
gmanev7, line 11: why does its type differ from the bank and dex ones?

gmanev7 (approved):

PR #387, opened 2024-09-02
chore(scripts): Update configs

desislavva:


metodi96:
Looks good to me, the majority of the params usually stay the same, the difference is with the LPP base interest rate and the margin interest rate defined in the Leaser. It would be specific for each market that gets deployed.

metodi96 (approved):
Looks good to me, the majority of the params usually stay the same, the difference is with the LPP base interest rate and the margin interest rate defined in the Leaser. It would be specific for each market that gets deployed.

PR #389, opened 2024-09-11
fix(build): Remove dev- & test-net protocol combinations

KirilMihaylov:


protocol/contracts/void/src/contract.rs
gmanev7, line 1: what's the idea behind this removal?

gmanev7: ahaa, I saw, you just moved its content elsewhere

gmanev7 (approved):

PR #398, opened 2024-09-24
build: Forbid `unexpected_cfgs` lint on workspace's level

KirilMihaylov:


gmanev7:
What's the motivation behind it? Forbidding lints is not a good practice unless there is no other way to solve something. Could you present the problem, and some context, please?

platform/Cargo.toml
```diff
@@ -7,6 +7,9 @@ members = [
 
 resolver = "2"
 
+[workspace.lints.rust.unexpected_cfgs]
+level = "forbid"
```
gmanev7, line 11:
```suggestion
[workspace.lints.rust]
unexpected_cfgs = { level = "forbid" }
```

gmanev7: this is more readable and matches the examples provided in the books

gmanev7, line 11: I doubt this would work without the `check-cfg` attribute, see [here](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#explanation-90)

KirilMihaylov: Tested it just now locally by adding `#[cfg(abc)]` and it reported it as expected.

PR #404, opened 2024-09-24
ci: Ensure arguments are of boolean type.

KirilMihaylov:


.github/actions/cache-rust/action.yaml
```diff
@@ -78,6 +78,23 @@ runs:
...
+          ("false"|"true") ;;
+          (*)
+            "echo" "Expected boolean value for parameter, but got: \"${value}\"\
+        !"
+
+            error_code="1"
+            ;;
+          esac
+        done
+
```
gmanev7, line 97: Do you have access to the source tree at this point? If yes, let's define a function that checks a single boolean value and call it as many times as we need here.

PR #411, opened 2024-09-24
fix(protocol/lpp): Add required `cfg` on the `state` module and remove redundant ones from submodules.

KirilMihaylov:


gmanev7 (approved):
nice catch!

PR #414, opened 2024-09-24
fix(platform/tree): Linter-reported issue about change of behavior in `macro_rules!`'s `expr`.

KirilMihaylov:


gmanev7:
@KirilMihaylov FYI this module is going to be removed from main with the next commit

PR #420, opened 2024-09-24
refactor(platform/sdk): Split `cosmos_proto` off `cosmos_ibc` for the sake of granularity.

KirilMihaylov:


gmanev7:
at first sight, replacing `sdk/cosmos_ibc` with `sdk/cosmos_proto` would work

KirilMihaylov:
I kept the dependency of other packages the same, just needed a minimal `prost` re-export without pulling the rest or having to manage the `prost` version manually.

PR #422, opened 2024-09-24
refactor(protocol/dex+swap): Revamp structure to better utilize the type system for bug prevention.

KirilMihaylov:


KirilMihaylov:
CI shall be fixed in consequent commits to the branch.

protocol/Cargo.toml
```diff
@@ -48,6 +48,7 @@ base64 = { version = "0.22", default-features = false }
 serde = { version = "1", default-features = false }
 serde_json = "1"
 postcard = { version = "1", default-features = false }
+prost = { version = "0.12", default-features = false }
```
gmanev7, line 51: in opposite, please remove this dependency from platform/Cargo.toml, and migrate all imports `use prost::*` to use `sdk::cosmos_sdk_proto::prost::*`

protocol/packages/dex/src/swap.rs
```diff
@@ -21,9 +21,7 @@ pub trait ExactAmountIn {
         GIn: Group,
         GSwap: Group;
 
-    fn parse_response<I>(trx_resps: &mut I) -> Result<Amount>
-    where
-        I: Iterator<Item = Any>;
+    fn parse_response(response: CosmosAny) -> Result<Amount>;
```
gmanev7, line 24: do not buy this change without any motivation. Two reasons to reject this change:
- the parse is designed to be reflective of the build fn - the latter allows multiple messages to be sent as a transaction so parse should also be able to parse multiple responses
 - the API allowed communication to more complex DEX APIs

protocol/packages/swap/Cargo.toml
```diff
@@ -16,50 +16,77 @@ combinations = [
...
+]
 
 [dependencies]
-currency = { workspace = true, optional = true }
 dex = { workspace = true, optional = true }
+oracle = { workspace = true, optional = true }
+
+currency = { workspace = true, optional = true }
 finance = { workspace = true, optional = true }
-oracle = { workspace = true, optional = true, features = ["stub_swap"] }
```
gmanev7, line 53: are you sure "stub_swap" is not needed in all the features? I am

protocol/Cargo.toml
KirilMihaylov: As mentioned before, if I could, I would. The `Message` derivation macro expects `prost` as a dependency and not as an import.

protocol/packages/dex/src/swap.rs
KirilMihaylov: The old `parse_response` followed a bad practice, taking the iterator and polling it, which **is** defined as implementation dependent what it will do after the first `None`. This leads to error-proneness.

Another point I'd like to make is that, the iterator iterates over **multiple** responses, while the the function operates on the next *one*.

protocol/packages/swap/src/astroport/mod.rs
```diff
@@ -95,24 +73,35 @@ where
...
+        "neutron12jm24l9lr9cupufqjuxpdjnnweana4h66tsx5cl800mke26td26sq7m05p";
+}
+
+#[cfg(feature = "dex-astroport_main")]
+impl Router for Impl {
+    // Source: https://github.com/astroport-fi/astroport-changelog/blob/main/neutron/neutron-1/core_mainnet.json
+    const ADDRESS: &'static str =
+        "neutron1rwj6mfxzzrwskur73v326xwuff52vygqk73lr7azkehnfzz5f5wskwekf4";
+}
+
```
gmanev7, line 104: The more I look at this more I doubt the trait is needed. To say it straight - it's an overkill. In addition, the new way you suggest to be consumed is not acceptable - `<Self as Router>`

KirilMihaylov: Why wouldn't it be acceptable? And I wouldn't say it's an overkill as this **dis**allows different signatures while also evading `cfg` shenanigans in the constant's value definition.

protocol/packages/swap/Cargo.toml
KirilMihaylov: You were right, it is needed but only for the `testing` feature, thus I have moved it here:
https://github.com/nolus-protocol/nolus-money-market/pull/422/files/f338d2c5e7996e9a1d7d5d94ff1cb4ddc403f480#diff-eb2041f1f3c24c33db9764eb04a011b1818b14e34b3906f0fd713b4ae491124fR61

gmanev7: impossible, the imports [here ](https://github.com/nolus-protocol/nolus-money-market/blob/7522439e31d512f63ef5b05d144ae205938ea324/protocol/packages/swap/src/astroport/mod.rs#L12) are included only under "stub_swap"

KirilMihaylov: Then some other package must be including the feature. What I'm trying to say is that it still works without it. Saying "impossible" doesn't exactly contribute much when I know for a fact that it still works.

I will add it to the `dex-astroport_{main|test}` features, but it should probably be investigated who still pulls it.

protocol/packages/dex/src/swap.rs
```diff
@@ -21,9 +23,9 @@ pub trait ExactAmountIn {
         GIn: Group,
         GSwap: Group;
 
-    fn parse_response<I>(trx_resps: &mut I) -> Result<Amount>
+    fn parse_response<I>(trx_resps: &mut Fuse<I>) -> Result<Amount>
```
gmanev7, line 26: too much
let's not pollute the interface with an extra requirements on the Iter impl

protocol/packages/swap/src/astroport/mod.rs
```diff
@@ -95,7 +75,7 @@ where
         })
     }
 
-    fn parse_response<I>(trx_resps: &mut I) -> Result<Amount>
+    fn parse_response<I>(trx_resps: &mut Fuse<I>) -> Result<Amount>
```
gmanev7, line 78: same

protocol/packages/swap/src/osmosis/mod.rs
```diff
@@ -57,20 +61,19 @@ impl ExactAmountIn for Impl {
         Ok(())
     }
 
-    fn parse_response<I>(trx_resps: &mut I) -> Result<Amount>
+    fn parse_response<I>(trx_resps: &mut Fuse<I>) -> Result<Amount>
```
gmanev7, line 64: again

protocol/packages/dex/src/impl_/transfer_out.rs
```diff
@@ -301,18 +306,18 @@ where
...
         G: Group + MemberOf<Self::GIn>,
     {
-        let next_idx = self.last_index.map_or(CoinsNb::ZERO, |prev_idx| {
-            prev_idx
-                .checked_add(1)
-                .expect("The swap task exceeds the max number of coins `CoinsNb::MAX`")
+        self.last_index = Some(if let Some(last_index) = self.last_index {
+            last_index.checked_add(1).ok_or(const { TooManyCoins })?
+        } else {
+            CoinsNb::ZERO
```
gmanev7, line 318: could you revert the code to using `Option::map_or` - less branching, more concise code

protocol/packages/swap/Cargo.toml
```diff
@@ -18,47 +18,45 @@ combinations = [
...
     "dep:astroport",
-    "dep:currency",
-    "dep:dex",
-    "dep:finance",
-    "dep:oracle",
-    "dep:platform",
-    "dep:sdk",
-    "dep:serde",
+    "sdk/cosmos_ibc",
 ]
```
gmanev7, line 29: I am curious why the test dex would require `"sdk/cosmos_ibc"` and the main dex not.

protocol/packages/swap/Cargo.toml
```diff
@@ -18,47 +18,45 @@ combinations = [
...
-    "dep:oracle",
-    "dep:platform",
-    "dep:sdk",
-    "dep:serde",
+    "sdk/cosmos",
+]
+dex-test_impl = [
+    "dep:prost",
+    "sdk/cosmos",
 ]
```
gmanev7, line 37: we have agreed on not providing it

protocol/packages/swap/src/astroport/mod.rs
```diff
@@ -113,6 +91,24 @@ where
...
+}
+
+#[cfg(feature = "dex-astroport_test")]
+impl Router for Impl {
+    // Source: https://github.com/astroport-fi/astroport-changelog/blob/main/neutron/pion-1/core_testnet.json
+    const ADDRESS: &'static str =
+        "neutron12jm24l9lr9cupufqjuxpdjnnweana4h66tsx5cl800mke26td26sq7m05p";
+}
+
+#[cfg(any(feature = "dex-astroport_main", feature = "dex-test_impl"))]
```
gmanev7, line 105: we avoid `cfg` in the not-top-level modules hence we want the `lib.rs` to select the appropriate implementation

protocol/packages/swap/src/astroport/mod.rs
```diff
@@ -33,37 +31,17 @@ mod testing;
...
-    /// Source: https://github.com/astroport-fi/astroport-changelog/blob/main/neutron/pion-1/core_testnet.json
-    const ROUTER_ADDR: &'static str =
-        "neutron12jm24l9lr9cupufqjuxpdjnnweana4h66tsx5cl800mke26td26sq7m05p";
-}
-
 #[derive(Serialize, Deserialize)]
-pub struct RouterImpl<R>(PhantomData<R>);
+pub enum Impl
+where
+    Self: ExactAmountIn, {}
```
gmanev7, line 37: [all dex-es] We agreed on bounding to `Self: Router`

protocol/packages/swap/src/testing/mod.rs
```diff
@@ -55,13 +53,44 @@ where
     CurrencyDTO::<G>::from_symbol_testing::<DexSymbols<G>>(symbol).map_err(Into::into)
 }
 
+#[cfg(any(feature = "testing", test))]
```
gmanev7, line 56: no need to double this since it's identical to the one in `lib.rs` right above `mod testing`

protocol/packages/swap/src/astroport/mod.rs
KirilMihaylov: Yes, I remember that we did but the compiler/clippy complained that the trait was more private. Should I just make it public too?

gmanev7: nope, let's try with a where clause at `impl ExactAmountIn for Impl ` would that work?

gmanev7 (approved):

PR #427, opened 2024-09-24
feat(protocol/protocol): Add protocol JSON deserialization library.

KirilMihaylov:


KirilMihaylov:
CI shall be fixed in consequent commits to the branch.

gmanev7:
Could you share your motivation behind this new library package?

gmanev7:
what other packages are going to benefit from it? what use case do you envisage it would take part in?

KirilMihaylov:
The only beneficiary so far is the `currencies` package's build script, but can be extended to cover `dex` too in the future, if we decide we'd like to.

protocol/packages/protocol/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+use std::collections::BTreeSet;
+
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct Protocol {
+    pub dex_network: String,
+    pub dex: String,
+    pub liquidity_provider_currency_ticker: String,
```
gmanev7, line 10: LPN is a known abbreviation for the project and in its lowercase form is used throughout the source code, please rename

protocol/packages/protocol/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+    pub stable_currency_ticker: String,
+    pub lease_currencies_tickers: BTreeSet<String>,
```
gmanev7, line 12: We already have an abstraction that on the wire is presented with its ticker. That is CurrencyDTO<G>. Please use it.

To save a cycle - G is unknown if the package does not depend on currencies. Either move the type to another package or make it generic over G.

PR #430, opened 2024-09-25
feat(lease): no version checks on legacy lease migration

gmanev7:
in addition:
fix(currencies): osmosis-osmosis-osmo denoms

### DIFF
--- a/protocol/contracts/lease/src/api/mod.rs
+++ b/protocol/contracts/lease/src/api/mod.rs
@@ -26,8 +26,16 @@ pub type LpnCoinDTO = crate::finance::LpnCoinDTO;
 
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
-pub struct MigrateMsg {}
+pub struct MigrateMsg {
+    pub kind: MigrationKind,
+}
 
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum MigrationKind {
+    LEGACY,
+    REGULAR,
+}
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, JsonSchema)]
 #[cfg_attr(any(test, feature = "testing"), derive(Debug))]
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
--- a/protocol/contracts/lease/src/api/mod.rs
+++ b/protocol/contracts/lease/src/api/mod.rs
@@ -27,7 +27,7 @@ pub type LpnCoinDTO = crate::finance::LpnCoinDTO;
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
 pub struct MigrateMsg {
-    pub kind: MigrationKind,
+    pub kind: MigrationKind, // TODO remove it once the migration ot v0.7.3 is over
 }
 
 #[derive(Serialize, Deserialize, Clone)]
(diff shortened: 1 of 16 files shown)


protocol/contracts/lease/src/api/mod.rs
```diff
@@ -26,8 +26,16 @@ pub type LpnCoinDTO = crate::finance::LpnCoinDTO;
...
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(deny_unknown_fields, rename_all = "snake_case")]
-pub struct MigrateMsg {}
+pub struct MigrateMsg {
+    pub kind: MigrationKind,
+}
 
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum MigrationKind {
```
KirilMihaylov, line 35: Please add `// TODO` comment for removing it just so the IDEs can find it and we don't forget it. Probably not needed anywhere else because removing the definition will break the rest anyway.

protocol/contracts/leaser/src/migrate.rs
```diff
@@ -36,12 +36,12 @@ where
...
     I: Iterator<Item = MaybeCustomer<LI>>,
     LI: ExactSizeIterator<Item = Addr>,
-    MsgFactory: Fn() -> MigrateMsg,
+    MsgFactory: Fn(MigrationKind) -> MigrateMsg,
 {
     let mut msgs = MigrateBatch::new(lease_code, max_leases);
 
     let legacy_leases = msgs
-        .migrate_leases(leases, &migrate_msg)
+        .migrate_leases(leases, &|| migrate_msg(MigrationKind::LEGACY))
```
KirilMihaylov, line 44: Taking a reference to the closure is most likely unnecessary and might also be caught by clippy in the CI.

protocol/contracts/leaser/src/migrate.rs
```diff
@@ -51,7 +51,9 @@ where
     legacy_leases.and_then(|()| {
         customers
             .find_map(|maybe_customer| match maybe_customer {
-                Ok(customer) => msgs.migrate_or_be_next(customer, &migrate_msg),
+                Ok(customer) => {
+                    msgs.migrate_or_be_next(customer, &|| migrate_msg(MigrationKind::REGULAR))
```
KirilMihaylov, line 55: Taking a reference to the closure is most likely unnecessary and might also be caught by clippy in the CI.

protocol/contracts/leaser/src/migrate.rs
```diff
@@ -371,7 +377,7 @@ mod test {
         vec![Ok(cust1), Ok(cust2), Ok(cust3), Ok(cust4)].into_iter()
     }
 
-    fn migrate_msg() -> impl Fn() -> MigrateMsg {
-        || MigrateMsg {}
```
KirilMihaylov, line 375: Just as a note about the previous implementation, those are totally unnecessary gymnastics because passing the function pointer would've done that same without introducing indirection. I hope it will stay the way the new one is implemented.

KirilMihaylov (commented):
Comments do not concern the implementation but rather small details in the use of it.

protocol/contracts/leaser/src/migrate.rs
gmanev7: neither of them - the compiler complains if not provided, and the clippy is happy with it

protocol/contracts/leaser/src/migrate.rs
gmanev7: neither of them - the compiler complains if not provided, and the clippy is happy with it

protocol/contracts/leaser/src/migrate.rs
gmanev7: this is an old chunk of code and do net see any sense of commenting it now

PR #436, opened 2024-10-02
refactor(tools/topology): Refactor AMM pools and DEX definitions.

KirilMihaylov:


tools/topology/src/currency_resolution.rs
```diff
@@ -0,0 +1,259 @@
...
+    dex_symbol: symbol::Builder,
+    currency: &'currency currency::Native,
+    host_local: bool,
+}
+
+impl<'network, 'currency> DexResolvedCurrency<'network, 'currency>
+where
+    'currency: 'network,
+{
+    fn resolve_bank_side(
```
gmanev7, line 192: too long method, please refactor

tools/topology/src/currency_resolution.rs
```diff
@@ -0,0 +1,259 @@
...
+    {
+        // Shadow to shrink reference's lifetime.
+        let mut currency = currency;
+
+        let mut traversed_networks = vec![dex_network];
+
+        let mut dex_symbol = symbol::Builder::NEW;
+
+        Ok(loop {
+            match currency {
```
gmanev7, line 49: Can you de-loop the algo?

tools/topology/src/currency/mod.rs
```diff
@@ -7,7 +12,19 @@ mod native;
...
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize)]
 #[serde(transparent)]
-pub(crate) struct Id(String);
+pub struct Id(String);
+
+impl Hash for Id {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        str::hash(&self.0, state)
+    }
+}
```
gmanev7, line 21: curios how this implementation differs from the `impl Hash for String` that would be used if it was `derive(Hash)`-ed?

KirilMihaylov: It looks the same on first read of the expanded, but I'd prefer to keep it explicit but that's my own preference. My reasoning is that `Borrow{|Mut}` explicitly states the type much have the exact same `Eq`, `Ord` and `Hash` implementations.

KirilMihaylov: I agree about having `Eq` derived because it's just equality and field ordering doesn't matter. The point you make about `Ord` though is good because it too does rely on field ordering like `Hash`. If you wouldn't mind, I'd convert `Ord` to an explicit implementation.

My intentions to make it explicit were to propagate the intention that it's a special case, in this case a only-one-field structure. If you would prefer it being derived anyway, I won't be not opposed to the motion.

gmanev7: less code, better
I vote for derived versions

gmanev7 (approved):

PR #437, opened 2024-10-09
Batch methods modification

maneva3:


gmanev7 (commented):
Good job with the exception of `mod leaser::migrate`

PR #438, opened 2024-10-14
build: Reflect DEX selection changes

KirilMihaylov:


protocol/contracts/oracle/Cargo.toml
```diff
@@ -80,5 +61,6 @@ currency = { workspace = true, features = ["testing"] }
 finance = { workspace = true }
 platform = { workspace = true, features = ["testing"] }
 schema = { workspace = true }
+oracle = { workspace = true, features = ["contract"] }
```
gmanev7, line 64: Are you sure the oracle depends on itself?

gmanev7 (approved):

protocol/contracts/oracle/Cargo.toml
KirilMihaylov: These are the `dev-dependencies` so I'm pretty confident there was a reason to add it too. I think the tests required some `contract`-gated modules/definitions.

gmanev7 (approved):

PR #441, opened 2024-10-30
feat(protocol/currencies): Generate currencies based off provided network topology and protocol definition.

KirilMihaylov:


protocol/packages/currencies/src/payment/mod.rs
```diff
@@ -1,45 +1,51 @@
...
+use currency::{AnyVisitor, Matcher, MaybeAnyVisitResult, MemberOf};
+use sdk::schemars::JsonSchema;
 
-pub use self::only::PaymentOnlyGroup;
+use crate::{lease, lpn, native};
 
+pub use self::only::Group as OnlyGroup;
 #[cfg(feature = "testing")]
-pub use testing::*;
+pub use self::testing::*;
```
gmanev7, line 10: could you please enumerate the types explicitly?

protocol/packages/currencies/src/payment/mod.rs
```diff
@@ -1,45 +1,51 @@
...
-    fn maybe_visit_member<M, V>(_matcher: &M, _visitor: V) -> MaybeAnyVisitResult<Self::TopG, V>
+    #[cold]
+    #[inline]
+    fn maybe_visit_member<M, V>(_: &M, _: V) -> MaybeAnyVisitResult<Self::TopG, V>
     where
         M: Matcher,
         V: AnyVisitor<Self::TopG>,
     {
-        unreachable!()
+        unimplemented!()
```
gmanev7, line 47: disagree
the meaning of `unreachable` better match the real case, please revert

protocol/packages/currencies/src/payment/only/mod.rs
```diff
@@ -1,24 +1,32 @@
...
+#[path = "testing.rs"]
+pub(crate) mod impl_mod;
 
-#[derive(Clone, Copy, Debug, Ord, PartialEq, PartialOrd, Eq)]
-pub struct PaymentOnlyGroup {}
+#[derive(
+    Clone, Copy, Debug, Ord, PartialEq, PartialOrd, Eq, Serialize, Deserialize, JsonSchema,
+)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[schemars(crate = "sdk::schemars")]
```
gmanev7, line 21: [everywhere] what does it bring to the definition? We have been living without it.

gmanev7 (commented):
reviewed the code without currencies/build/

protocol/packages/currencies/build/convert_case.rs
```diff
@@ -0,0 +1,36 @@
...
+    .for_each(|substring| {
+        let mut chars = substring.chars();
+
+        if let Some(first_character) = chars.next() {
+            string.push(first_character.to_ascii_uppercase());
+
+            chars
+                .map(|ch| ch.to_ascii_lowercase())
+                .for_each(|ch| string.push(ch));
+        }
```
gmanev7, line 27: collecting chars from a composite iterator would avoid pushing chars into a String

protocol/packages/currencies/src/payment/only/mod.rs
KirilMihaylov: It removes the need for the `sdk::schemars::{self, JsonSchema}` which makes it cryptic why the module itself is imported via `self`. This makes it more explicit and readable.

protocol/packages/currencies/src/payment/mod.rs
KirilMihaylov: I changed it because the documentation of the `unreachable!` macro states that:

> This is useful any time that the compiler can’t determine that some code is unreachable. For example:
>
> * Match arms with guard conditions.
> * Loops that dynamically terminate.
> * Iterators that dynamically terminate.

I'd argue that our case does not fit any of these and it's more of an API design issue, thus `unimplemented!` fits better in my personal opinion.

protocol/packages/currencies/src/lease/mod.rs
```diff
@@ -1,30 +1,32 @@
...
-pub struct LeaseGroup {}
+#[schemars(crate = "sdk::schemars")]
+pub enum Group {}
 
-impl Group for LeaseGroup {
+impl currency::Group for Group {
     const DESCR: &'static str = "lease";
-    type TopG = PaymentGroup;
 
+    type TopG = payment::Group;
```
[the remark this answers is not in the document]
KirilMihaylov: I don't think there is a preferred way to do that, but it I think it pollutes the scope more and I followed the example of how we refer to functions from another module.

protocol/packages/currencies/build/currencies_tree.rs
```diff
@@ -0,0 +1,144 @@
...
+        );
+
+        let result = inverse_targets
+            .iter()
+            .try_fold(self.parents, |mut parents, target| {
+                if parents.entry(target).or_default().insert(ticker) {
+                    ControlFlow::Continue(parents)
+                } else {
+                    ControlFlow::Break(())
+                }
```
gmanev7, line 86: .then_some(parents) would be shorter

protocol/packages/currencies/build/currencies_tree.rs
```diff
@@ -0,0 +1,144 @@
...
+                .collect(),
+        );
+
+        let result = inverse_targets
+            .iter()
+            .try_fold(self.parents, |mut parents, target| {
+                if parents.entry(target).or_default().insert(ticker) {
+                    ControlFlow::Continue(parents)
+                } else {
+                    ControlFlow::Break(())
```
gmanev7, line 85: Is that an indication of an error? If yes,one must return some

protocol/packages/currencies/build/currencies_tree.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    fn process_targets(
+        mut self,
+        protocol: &Protocol,
+        host_currency_ticker: &str,
+        ticker: &'topology str,
+        targets: &'topology PairTargets,
+    ) -> ControlFlow<(), Self> {
```
gmanev7, line 64: Let's avoid misusing that type in scenarios it is not designed for. After all, here we do not control any flow.

protocol/packages/currencies/build/currencies_tree.rs
```diff
@@ -0,0 +1,144 @@
...
+            .filter(|&(ticker, _)| {
+                super::filter_selected_currencies(protocol, host_currency_ticker, ticker)
+            })
+            .try_fold(
+                const {
+                    Self {
+                        parents: const { BTreeMap::<_, BTreeSet<_>>::new() },
+                        children: const { BTreeMap::<_, BTreeSet<_>>::new() },
+                    }
+                },
```
gmanev7, line 44: Better move it into the type (CurrenciesTree)

protocol/packages/currencies/src/payment/mod.rs
KirilMihaylov: I would even go as far to suggest splitting the two traits if we even need to put a marker that this method is not implemented and/or shouldn't be executed.

gmanev7: The code is prose. The `unimplemened!` conveys the developer's intent to implement it at some point of time. In our case, that is not true.
The best solution would be to transform the trait(s) to avoid such cases, but since it requires additional work I insist on reverting to the main branch state.

protocol/packages/currencies/src/payment/only/mod.rs
gmanev7: this is quite opinionated 
the `use` solution is more straightforward - the compiler complains ... and one satisfies it. Nobody takes it as "cryptic" because when removes it - he receives the feedback quickly.

protocol/packages/currencies/build/convert_case.rs
```diff
@@ -0,0 +1,31 @@
...
+            })
+    })
+    .flat_map(|substring| {
+        let mut chars = substring.chars();
+
+        chars
+            .next()
+            .map(|first_character| first_character.to_ascii_uppercase())
+            .into_iter()
+            .chain(chars.map(|ch| ch.to_ascii_lowercase()))
```
gmanev7, line 23: why introduce a lambda? would not `char::to_ascii_lowercase` work for you?

protocol/packages/currencies/src/payment/mod.rs
KirilMihaylov: I disagree with the first statement because this is from the documentation of `todo!`:
> The difference between `unimplemented!` and `todo!` is that while `todo!` conveys an intent of implementing the functionality later and the message is “not yet implemented”, `unimplemented!` makes no such claims. Its message is “not implemented”.

For the second, if you agree, could you write it down as an issue so we can pick it up at a later date?

protocol/packages/currencies/src/payment/only/mod.rs
KirilMihaylov: Well, I did take it as cryptic, but I agree it is opinionated. I would rather have it explicit as all things, but I'm willing to compromise on this.

protocol/packages/currencies/build/convert_case.rs
KirilMihaylov: I would have if I could, as you know I usually am the one to try simplify those. It takes a reference to `self`, instead of by value.

protocol/packages/currencies/src/payment/mod.rs
gmanev7: After digging more into their descriptions, I have to admit that I have been using `unreachable` in cases it was not intended to be.  Since there are more occurrences in the code, I'd suggest changing that in a dedicated commit/PR. 
https://github.com/nolus-protocol/nolus-money-market/issues/445

KirilMihaylov: Okay. That sounds good to me! Reverting the change here now.

protocol/packages/currencies/src/lease/mod.rs
```diff
@@ -1,30 +1,32 @@
...
+
+    #[inline]
```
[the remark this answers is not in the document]
KirilMihaylov: The PR in it's current state is in the middle, thus why I chose it as a base.
image

protocol/packages/currencies/build/currencies_tree.rs
```diff
@@ -0,0 +1,130 @@
...
+
+pub(crate) struct CurrenciesTree<'parents_of, 'parent, 'children_of, 'child> {
+    parents: BTreeMap<&'parents_of str, BTreeSet<&'parent str>>,
+    children: BTreeMap<&'children_of str, BTreeSet<&'child str>>,
+}
+
+impl<'topology> CurrenciesTree<'topology, 'topology, 'topology, 'topology> {
+    const EMPTY: Self = Self {
+        parents: BTreeMap::<_, BTreeSet<_>>::new(),
+        children: BTreeMap::<_, BTreeSet<_>>::new(),
```
gmanev7, line 23: I think `Parents` and `Children` is what you wanted to use here, instead of `BtreeSet`?

protocol/packages/currencies/build/main.rs
```diff
@@ -0,0 +1,148 @@
...
+                File::open(TOPOLOGY_JSON).context("Failed to open \"topology.json\"!")?,
+            )
+            .context("Failed to parse topology JSON!")?,
+            serde_json::from_reader(
+                File::open(PROTOCOL_JSON).context("Failed to open \"protocol.json\"!")?,
+            )
+            .context("Failed to parse protocol JSON!")?,
+        )
+    } else {
+        Ok(())
```
gmanev7, line 76: I am curious why we say OK if the descriptors are missing.

protocol/packages/currencies/build/main.rs
```diff
@@ -0,0 +1,148 @@
...
+    );
+
+    let files_exist = IntoIterator::into_iter([PROTOCOL_JSON, TOPOLOGY_JSON])
+        .map(Path::new)
+        .map(Path::try_exists)
+        .try_fold(true, |all_exist, result| {
+            result
+                .map(|exists| all_exist && exists)
+                .context("Failed to check whether JSON descriptor file exists!")
+        })?;
```
gmanev7, line 45: please extract this chunk into a function to reduce the main fn length

protocol/packages/currencies/build/main.rs
```diff
@@ -0,0 +1,148 @@
...
+            )
+        } else {
+            bail!(
+                "`{BUILD_REPORT:?}` environment variable set but topology \
+                and/or protocol descriptors don't exist!",
+            );
+        }
+    } else {
+        Either::Right(io::stderr())
+    };
```
gmanev7, line 60: please extract this chunk into a function to reduce the main fn length

protocol/packages/currencies/build/main.rs
```diff
@@ -0,0 +1,148 @@
...
+        println!("cargo::rerun-if-changed={path}");
+    }
+
+    println!("cargo::rerun-if-env-changed={BUILD_REPORT}");
+
+    let output_directory: &Path = &PathBuf::from(
+        env::var_os("OUT_DIR").context("Cargo did not set `OUT_DIR` environment variable!")?,
+    );
+
+    let files_exist = IntoIterator::into_iter([PROTOCOL_JSON, TOPOLOGY_JSON])
```
gmanev7, line 38: please rename the variable to become more descriptive, for example, `definition_found`

protocol/packages/currencies/build/main.rs
```diff
@@ -0,0 +1,148 @@
...
+        output_directory,
+        &protocol,
+        &host_currency,
+        &dex_currencies,
+        &CurrenciesTree::new(&topology, &protocol, host_currency.ticker())?,
+    )
+}
+
+#[inline]
+fn filter_selected_currencies(
```
gmanev7, line 139: name it better - for example `protocol_currency`
and make it a member of `Protocol`

protocol/packages/currencies/build/main.rs
```diff
@@ -0,0 +1,148 @@
...
+        .map(Path::new)
+        .map(Path::try_exists)
+        .try_fold(true, |all_exist, result| {
+            result
+                .map(|exists| all_exist && exists)
+                .context("Failed to check whether JSON descriptor file exists!")
+        })?;
+
+    let build_report = if let Some(report_file) = env::var_os(BUILD_REPORT) {
+        if files_exist {
```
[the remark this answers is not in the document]
KirilMihaylov: That is a really good point. Thank you for pointing it out!

protocol/packages/currencies/build/main.rs
KirilMihaylov: This case is valid as we have the dummy currencies gated behind `testing`, which do not require any additional sources.

protocol/packages/currencies/build/currencies_tree.rs
KirilMihaylov: Sadly no, I thought the same thing when I extracted it, but `Parents` and `Children` actually hold references. I should be able to rework them a bit actually and use them here.
Also, I should be able to elide the generics here, as the types are well-know.

gmanev7: without claiming that I understand the entire mechanics here, I would suggest reviewing if the ownership of the tickets could be brought here and allow the usage of Parents and Children.

protocol/packages/currencies/build/main.rs
gmanev7: may we check the presence of that gate, and Err() otherwise?

protocol/packages/currencies/build/main.rs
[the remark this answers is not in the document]
KirilMihaylov: Looking a bit more into the context here, I actually do check that because my assumption is that if the environment variable is set, then the one setting it _expects_ that there should be a build step. That is why I error when that is not the case.

protocol/packages/currencies/build/main.rs
```diff
@@ -35,45 +36,50 @@ fn main() -> Result<()> {
         env::var_os("OUT_DIR").context("Cargo did not set `OUT_DIR` environment variable!")?,
     );
 
-    let files_exist = IntoIterator::into_iter([PROTOCOL_JSON, TOPOLOGY_JSON])
+    if env::var_os("CARGO_FEATURE_TESTING").is_some() {
+        Ok(())
+    } else if check_for_definitions()? {
+        let build_report = build_report_writer()?;
```
gmanev7, line 42: I believe the `let output_directory = ...` scope could be reduced to this block only

protocol/packages/currencies/build/main.rs
```diff
@@ -91,58 +97,26 @@ where
         dex_currencies,
     } = topology.currency_definitions(&protocol.dex_network)?;
 
-    if *protocol.lpn_ticker == *host_currency.ticker() {
+    if protocol.lpn_ticker == CurrencyDefinition::ticker(host_currency.borrow()) {
```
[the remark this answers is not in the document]
KirilMihaylov: I added it just now to prevent it being mixed up with the other currencies.

PR #444, opened 2024-11-14
refactor(protocol/lease): Remove unnecessary `unwrap`s by revisiting API.

KirilMihaylov:


gmanev7:
Thanks for the PR!
I have an even better and simpler solution that will be checked in soon

PR #447, opened 2024-11-15
ci: Revamp workflows to run with the configurations in mind and remove unnecessary support GitHub Actions.

KirilMihaylov:


.github/test-data/build-configuration/protocol.json
gmanev7, line 1: do we still need protocol and topology definitions for the tests?

KirilMihaylov: Yes, we do in the case when `testing` is not turned on, which is a non-negligible portion from what I remember. I don't expect us changing it as long as we don't change the topology, which too I hope is unlikely.

KirilMihaylov: Not to mention that this also does indirectly test the currency generation too.

gmanev7: expressed offline my concerns over usage of these dummy currencies in packages that depend on "currencies", like leaser, oracle, lpp, to mention a few

gmanev7: validated that they are copied in lint, test and check stages

PR #453, opened 2024-11-15
feat: Currency generation & major CI refactor

KirilMihaylov:


protocol/packages/marketprice/src/feed/observations.rs
```diff
@@ -5,19 +5,16 @@ use crate::error::PriceFeedsError;
 
 use super::Observation;
 
+pub type Result<T, E = PriceFeedsError> = std::result::Result<T, E>;
```
gmanev7, line 8: let's avoid default values of generic parameters

protocol/contracts/lease/src/position/steady.rs
```diff
@@ -70,8 +70,8 @@ where
 #[cfg(test)]
 mod tests {
     use cosmwasm_std::Timestamp;
+    use currencies::{testing::PaymentC3, Lpn};
     use currencies::{LeaseGroup, Lpns};
```
gmanev7, line 74: could you merge the two imports?

gmanev7 (approved):

PR #462, opened 2024-12-06
ci: Add a switch for checking the dependencies' versions.

KirilMihaylov:


.github/workflows/smart_contracts.yaml
```diff
@@ -157,7 +164,10 @@ jobs:
     steps:
       - uses: "actions/checkout@v4"
       - if: |-
-          github.ref_type == 'tag'
+          github.ref_type == 'tag' || (
+            github.event_name == 'workflow_dispatch' &&
+              inputs.check-dependencies-version
+          )
```
gmanev7, line 170: may we reuse `check-dependencies-versions`?

KirilMihaylov: I should be able to. There were some exceptions to where the `env` context can be used, but it should work here.

gmanev7 (approved):

PR #470, opened 2025-01-08
refactor: Remove `Either<L, R>` implementations and rewire to use `either` crate instead.

KirilMihaylov:


gmanev7 (commented):
a general question:
Why do we build a home-grown solution instead of using https://crates.io/crates/either?

tools/cargo-each/Cargo.toml
```diff
@@ -13,6 +13,8 @@ combinations = [
 ]
 
 [dependencies]
+either = { workspace = true }
+
 anyhow = { workspace = true }
```
gmanev7, line 18: alphabetical order?

protocol/Cargo.toml
```diff
@@ -52,6 +52,7 @@ topology = { path = "../tools/topology" }
 # General
 anyhow = "1"
 base64 = { version = "0.22", default-features = false }
+either = { version = "1.13.0", default-features = false }
```
gmanev7, line 55: as we have discussed multiple times we depend on the first not-zero version

tools/Cargo.toml
```diff
@@ -12,11 +12,14 @@ license = "Apache-2.0"
...
 [workspace.dependencies]
-topology.path = "topology"
+# Own packages
+topology = { path = "topology" }
 
+# General packages
 anyhow = "1"
 cargo_metadata = "0.18.1"
 clap = { version = "4.5", features = ["derive", "unicode"] }
+either = { version = "1.13", default-features = false }
```
gmanev7, line 22: as we have discussed multiple times we depend on the first not-zero version

gmanev7 (approved):

Issue #466, opened 2024-12-13
Add support for protocol releases


gmanev7:
The smart contracts migration ensures the wasm packages are part of the same release.

The wasm package build process has undergone an evolution of decoupling protocol descriptors from the base software packages. As a result, it is easy to develop multiple wasm packages originating from the same base software package, and hence carrying the same release tag. This would be a source of deployment mistakes that may or may not be caught during migration. For example, there is a chance the Oracle migration to spot currency mismatches.

This ticket includes
1. Pairing the protocol contracts' package release with the protocol release so they both get embedded into the software.
2. Adding a request/query to any of the contracts for its release, namely, the package release for the platform contracts, and the release defined in 1. for the protocol contracts.
3. Cease persisting the release in the contract storage.
4. Providing the previous release to the new contract as a parameter of the migration message. The new contract's migration should validate if it's compatible with the previous release. 

Compatibility checks include:
1. a check that the ProtocolPackageRelease::protocol of each migrated protocol package matches the target protocol release passed as a transaction attribute
2. a check that the SoftwarePackageRelease::code is not downgraded

On migration, the Admin contract should ensure all contracts' SoftwarePackageRelease::id matches the target version passed as a transaction attribute. 

A sketch of the abstractions:
```
SoftwarePackageRelease {

    /// release ID, for example, as outputted from `git describe`, or the git commit hash
    id: &str, // for ex. "v0.7.6-4-gbec4b91a" provided through an env variable

    code: SoftwarePackage {
                name: &str, // the Cargo package name
                version: SemVer, // the Cargo package version
                storage: VersionSegment
              },
},

```
```
PlatformPackageRelease {
    software: SoftwarePackageRelease,
}
```
```
ProtocolPackageRelease {
    software: SoftwarePackageRelease,
    protocol: ProtocolRelease {

              /// release ID, for example, as outputted from `git describe`, or the git commit hash
              id: &str, // for ex. "v0.2.1" provided through an env variable

              protocol: ProtocolID {
                   name: &str, // "osmosis-osmosis-osmo" provided through an env variable
                   network: &str, // "vitosha-7" provided through an env variable
               },
           } 
}

```

Before sending a migration message, the contract owner, the Leaser for all the Leases within a protocol or the Admin for all the rest, obtains the current release of the package and includes that release when building the migration message for that same package. Upon receiving the migration message, the contract performs all compatibility checks and optionally migrates the storage. Finally, the contract returns its package release.
[...]

PR #472, opened 2025-01-08
feat(versioning): intro Release

gmanev7:
- this is the foundation of the target SoftwarePackageRelease type described in #466
- going to evolve it further

### DIFF
--- a/platform/packages/versioning/src/lib.rs
+++ b/platform/packages/versioning/src/lib.rs
@@ -1,13 +1,16 @@
-use serde::{Deserialize, Serialize};
-
-#[cfg(feature = "schema")]
-use sdk::schemars::{self, JsonSchema};
-use sdk::{
-    cosmwasm_std::{StdError, StdResult, Storage},
-    cw_storage_plus::Item,
+use std::{
+    cmp::Ordering,
+    fmt::{Display, Formatter, Result as FmtResult},
 };
 
-pub use self::release::Release;
+use serde::{Deserialize, Serialize};
+
+use sdk::cosmwasm_std::{StdError, StdResult, Storage};
+#[cfg(feature = "schema")]
+use sdk::schemars::{self, JsonSchema};
+
+//TODO issue#466 hide `Id`
+pub use self::release::{Id, Release};
 
 mod release;
 
@@ -64,16 +67,65 @@ impl SemVer {
     }
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+impl Display for SemVer {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        f.write_fmt(format_args!("{}.{}.{}", self.major, self.minor, self.patch))
+    }
+}
+
+// #[derive(Debug, Copy, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
 #[cfg_attr(feature = "schema", derive(JsonSchema))]
+/// A reference type representing a software package
+// TODO rename to SoftwarePackage
 pub struct Version {
     storage: VersionSegment,
+    /// the reference identification attribute
     software: SemVer,
 }
 
 impl Version {
     pub const fn new(storage: VersionSegment, software: SemVer) -> Self {
         Self { storage, software }
+    }
+
+    pub const fn same_storage(&self, other: &Self) -> bool {
+        self.check_storage(other.storage)
+    }
+
+    pub const fn next_storage(&self, other: &Self) -> bool {
+        other.check_storage(self.storage.wrapping_add(1))
+    }
+
+    const fn check_storage(&self, expected: VersionSegment) -> bool {
+        self.storage == expected
+    }
+}
+
+impl Display for Version {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        f.write_fmt(format_args!(
+            "version: {}, storage: {}",
+            self.software, self.storage
+        ))
+    }
+}
+
+impl Eq for Version {}
+
+impl PartialEq for Version {
+    fn eq(&self, other: &Self) -> bool {
+        let res = self.software == other.software;
+        if res {
+            debug_assert_eq!(self.storage, other.storage);
+        }
+        res
+    }
+}
+
+impl PartialOrd for Version {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.software.partial_cmp(&other.software)
     }
 }
 
@@ -87,13 +139,13 @@ macro_rules! package_version {
     }};
 }
 
-const VERSION_STORAGE_KEY: Item<Version> = Item::new("contract_version");
-
-pub fn initialize(storage: &mut dyn Storage, version: Version) -> StdResult<()> {
-    VERSION_STORAGE_KEY.save(storage, &version)
+pub fn initialize(_storage: &mut dyn Storage, _version: Version) -> StdResult<()> {
+    // no op
+    // TBD remove from the stack upward
+    Ok(())
 }
 
-pub fn update_software<ContractError, MapErrorFunctor>(
+pub fn update_legacy_software<ContractError, MapErrorFunctor>(
     storage: &mut dyn Storage,
     new: Version,
     map_error: MapErrorFunctor,
@@ -101,20 +153,20 @@ pub fn update_software<ContractError, MapErrorFunctor>(
 where
     MapErrorFunctor: FnOnce(StdError) -> ContractError,
 {
-    load_version(storage)
-        .and_then(|current| Release::from_env().allow_software_update(&current, &new))
-        .and_then(|()| save_version(storage, &new))
-        .map(|()| Release::from_env())
+    Release::pull_prev(storage)
+        .and_then(|prev_release| {
+            let this_release = Release::this(new);
+            prev_release.update_software(this_release)
+        })
         .map_err(map_error)
 }
 
 pub struct FullUpdateOutput<MigrateStorageOutput> {
-    pub release_label: Release,
+    pub to: Release,
     pub storage_migration_output: MigrateStorageOutput,
 }
 
-pub fn update_software_and_storage<
-    const FROM_STORAGE_VERSION: VersionSegment,
+pub fn update_legacy_software_and_storage<
     ContractError,
     MigrateStorageFunctor,
     StorageMigrationOutput,
@@ -130,26 +182,18 @@ where
         FnOnce(&mut dyn Storage) -> Result<StorageMigrationOutput, ContractError>,
     MapErrorFunctor: FnOnce(StdError) -> ContractError,
 {
-    load_version(storage)
-        .and_then(|current| {
-            Release::from_env()
-                .allow_software_and_storage_update::<FROM_STORAGE_VERSION>(&current, &new)
+    Release::pull_prev(storage)
+        .and_then(|prev_release| {
+            let this_release = Release::this(new);
+            prev_release.update_software_and_storage(this_release)
         })
-        .and_then(|()| save_version(storage, &new))
         .map_err(map_error)
-        .and_then(|()| migrate_storage(storage))
-        .map(|storage_migration_output| FullUpdateOutput {
-            release_label: Release::from_env(),
-            storage_migration_output,
+        .and_then(|new_release| {
+            migrate_storage(storage).map(|storage_migration_output| FullUpdateOutput {
+                to: new_release,
+                storage_migration_output,
+            })
         })
-}
-
-fn load_version(storage: &mut dyn Storage) -> Result<Version, StdError> {
-    VERSION_STORAGE_KEY.load(storage)
-}
-
-fn save_version(storage: &mut dyn Storage, new: &Version) -> Result<(), StdError> {
-    VERSION_STORAGE_KEY.save(storage, new)
 }
 
 #[cfg(test)]
--- a/platform/packages/versioning/src/lib.rs
+++ b/platform/packages/versioning/src/lib.rs
@@ -9,8 +9,7 @@ use sdk::cosmwasm_std::{StdError, StdResult, Storage};
 #[cfg(feature = "schema")]
 use sdk::schemars::{self, JsonSchema};
 
-//TODO issue#466 hide `Id`
-pub use self::release::{Id, Release};
+pub use self::release::{PackageRelease, ReleaseId};
 
 mod release;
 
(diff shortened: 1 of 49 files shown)


platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
 
+    fn void() -> Self {
+        Self(Self::VOID.into())
+    }
+}
+
+impl Release {
     pub fn void() -> Self {
-        Self::instance(Self::VOID_RELEASE)
+        Self::instance(Id::void(), Version::new(0, SemVer::parse("0.0.0")))
```
KirilMihaylov, line 59: We shouldn't be parsing static data, known at compile time. Could you please add a private method that gives out this and only this specific version?
If we really don't want to do that, it should be wrapped inside a `const { ... }` block to force compile-time evaluation.

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
+        Self(Self::DEV.into())
+    }
 
+    fn void() -> Self {
+        Self(Self::VOID.into())
+    }
+}
+
+impl Release {
     pub fn void() -> Self {
```
KirilMihaylov, line 58: This method should be marked as `const` as it is a pure and total/non-divergent function, not even relying on heap operations.
```suggestion
    pub const fn void() -> Self {
```

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
+    pub(crate) fn pull_prev(storage: &mut dyn Storage) -> Result<Self, StdError> {
+        const VERSION_STORAGE_KEY: Item<Version> = Item::new("contract_version");
+
+        VERSION_STORAGE_KEY
+            .load(storage)
+            .inspect(|_| VERSION_STORAGE_KEY.remove(storage))
+            .map(|code| Self::instance(Id::prev(), code))
+    }
+
+    fn instance(id: Id, code: Version) -> Self {
```
KirilMihaylov, line 75: Same reasoning as for `Self::void`.
```suggestion
    const fn instance(id: Id, code: Version) -> Self {
```

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
+
+impl Release {
     pub fn void() -> Self {
-        Self::instance(Self::VOID_RELEASE)
+        Self::instance(Id::void(), Version::new(0, SemVer::parse("0.0.0")))
     }
 
-    pub(crate) fn from_env() -> Self {
-        Self::instance(Self::RELEASE_LABEL)
+    pub(crate) fn this(code: Version) -> Self {
```
KirilMihaylov, line 62: Same reasoning as for `Self::void`.
```suggestion
    pub(crate) const fn this(code: Version) -> Self {
```

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
-            current.storage,
-            FROM_STORAGE_VERSION,
-            "the expected origin storage version",
-        )?;
-
-        if current.storage.wrapping_add(1) == new.storage {
-            self.allow_software_update_int(current, new)
+    fn allow_software_update_int(&self, new: Self) -> Result<Self, StdError> {
+        let current = self.code;
+        if current < new.code || (self.id == Id::dev() && current == new.code) {
```
KirilMihaylov, line 91: Personally I do not feel like the introduction of a local variable is redundant as it does not improve readability nor does it make it more concise.

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
+        Self { id, code }
     }
 
-    pub fn allow_software_update(&self, current: &Version, new: &Version) -> Result<(), StdError> {
-        check_storage_match(
-            current.storage,
-            new.storage,
-            "the new software storage version",
-        )?;
+    pub fn update_software(self, to: Self) -> Result<Self, StdError> {
```
KirilMihaylov, line 79: Shouldn't this method only return `Id` as if I am not missing something, it should be the only thing the contract this gets returned to, `admin` or `leaser`, most likely don't care about the software nor storage version.

platform/packages/versioning/src/release.rs
KirilMihaylov, line 91: Shouldn't we allow the same the migration if the release ID and software version are the same? That would occur when only the currencies are changed.

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
-    ) -> Result<(), StdError> {
-        check_storage_match(
-            current.storage,
-            FROM_STORAGE_VERSION,
-            "the expected origin storage version",
-        )?;
-
-        if current.storage.wrapping_add(1) == new.storage {
-            self.allow_software_update_int(current, new)
+    fn allow_software_update_int(&self, new: Self) -> Result<Self, StdError> {
```
KirilMihaylov, line 89: It would probably be for the better to use the full word, `internal`, as `int` can also imply `integer`. While several characters more verbose, we are far from reaching identified length limits.

KirilMihaylov, line 89: `allow_` implies a state change, e.g. like a flag, rather that validation, which is what the implementation does.

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
+        }
+    }
+
+    fn check_storage_adjacent(&self, next: Version) -> Result<(), StdError> {
+        if self.code.next_storage(&next) {
+            Ok(())
+        } else {
+            Err(StdError::generic_err(format!(
+                "The new version \"{next}\" is not adjacent to the current one \"{current}\"!",
+                current = self.code
```
KirilMihaylov, line 116: [Minor] I would argue this would have mattered more if there was more than one value, but I am not willing to die on that hill, so it's up to your opinion on the topic.

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
-        )?;
+    pub fn update_software(self, to: Self) -> Result<Self, StdError> {
+        self.check_storage_match(to.code)
+            .and_then(|()| self.allow_software_update_int(to))
+    }
 
-        self.allow_software_update_int(current, new)
+    pub fn update_software_and_storage(self, to: Self) -> Result<Self, StdError> {
+        self.check_storage_adjacent(to.code)
+            .and_then(|()| self.allow_software_update_int(to))
```
KirilMihaylov, line 86: This repetition can be avoided via a higher-order function that will in turn take `check_storage_match` or `check_storage_adjacent` as an argument, e.g. `fn check<F: FnOnce(&Self, Version)>(&self, to: Self, f: F) -> -> Result<..., StdError> { f(self, to.code).and_then(|()| ...) }`. That would also remove the need for `allow_software_update_int` to exist on it's own because it can be easily inlined.

KirilMihaylov (changes requested):
Partial review over `platform/packages/versioning`.

KirilMihaylov:
I saw that you have replaced it in the CI and `README.md`, but I have a feeling the environment variable did also exist in the build script and possibly in the container image file.
Could you please make sure it is updated there too, as those two are used actively in the CI?

platform/packages/versioning/src/lib.rs
```diff
@@ -64,17 +67,66 @@ impl SemVer {
...
+}
+
+impl Eq for Version {}
+
+impl PartialEq for Version {
+    fn eq(&self, other: &Self) -> bool {
+        let res = self.software == other.software;
+        if res {
+            debug_assert_eq!(self.storage, other.storage);
+        }
```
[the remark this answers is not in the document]
gmanev7: this entity is of an Entity type, with identity carried out by the other attribute. If an instance is considered "equal-to" another then both should represent the same Entity and hence all their attributes should match. Put the other way around, the storage version of a given package should always be the same.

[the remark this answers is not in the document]
gmanev7: a randomly picked page [on Entities and Values](https://blog.jannikwempe.com/domain-driven-design-entities-value-objects)

platform/packages/versioning/src/lib.rs
```diff
@@ -64,17 +67,66 @@ impl SemVer {
...
+        res
+    }
+}
```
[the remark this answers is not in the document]
gmanev7: disagree
The `Eq` trait adds the reflexive property, a == a. 
I do not make the connection with your reasoning. Please elaborate on it.

platform/packages/versioning/src/release.rs
gmanev7: not exactly that, you miss the point that the contract owner should make sure that the new contracts pertain to the protocol, ProtocolRelease, they operate on. Please check out the first compatibility check in the issue #466.  The code is not yet there, soon the contracts will return an instance of ProtocolPackageRelease or PlatformPackageRelease depending on their type.

platform/packages/versioning/src/lib.rs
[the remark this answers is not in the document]
KirilMihaylov: I guess I was mistaking it for another marker trait. I apologize.

platform/packages/versioning/src/release.rs
KirilMihaylov: The contract owner still wouldn't need to report back anything _**but**_ the release ID and the protocol ID, when applicable. Anything more than that would be a waste as it wouldn't matter to the contract owner. The owner wouldn't check the software version.

platform/packages/versioning/src/lib.rs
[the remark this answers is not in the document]
KirilMihaylov: I understand your reasoning, considering the latter part.

I also would argue that the fact that even with the invariant is that on the same software the storage version should also be the same, does not make it an entity.
I feel like this just does not fit the DD model well as the storage version is not an attribute separate from the software version, the invariant we rely on.
Granted, I haven't dug deep into DDD, so value objects might be able to define invariants too, but even following the article you sent, this does _**not**_ fit the entity definition.

gmanev7:
> I saw that you have replaced it in the CI and `README.md`, but I have a feeling the environment variable did also exist in the build script and possibly in the container image file. Could you please make sure it is updated there too, as those two are used actively in the CI?

yeah, `grep -ri ` did not show any

platform/packages/versioning/src/release.rs
KirilMihaylov: I see now. I must have missed that detail on when I wrote it.
I would suggest refactoring it to use `Cow<'static, str>` as it would act like a `String` during deserialization, yet not come with the additional overhead of memory allocations. Doing so will also solve this problem here as we know this data statically.

platform/packages/versioning/src/lib.rs
```diff
@@ -87,34 +139,35 @@ macro_rules! package_version {
...
     MapErrorFunctor: FnOnce(StdError) -> ContractError,
 {
-    load_version(storage)
-        .and_then(|current| Release::from_env().allow_software_update(&current, &new))
-        .and_then(|()| save_version(storage, &new))
-        .map(|()| Release::from_env())
+    PackageRelease::pull_prev(storage)
+        .and_then(|prev_release| {
+            let this_release = PackageRelease::this(new);
+            prev_release.update_software(this_release)
```
KirilMihaylov, line 159: Since [last comment](https://github.com/nolus-protocol/nolus-money-market/pull/472#discussion_r1907412173) got marked as outdated.

platform/packages/versioning/src/release.rs
KirilMihaylov: If that is the case, would you not agree that both should be bound to local variables, not just the self one?

platform/packages/versioning/src/lib.rs
```diff
@@ -64,17 +67,66 @@ impl SemVer {
...
+        }
+        res
+    }
+}
+
+impl PartialOrd for Version {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.software.partial_cmp(&other.software)
+    }
 }
```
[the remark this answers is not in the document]
KirilMihaylov: I would propose implementing `Ord` instead of `PartialOrd` and just forwarding `partial_cmp` to `cmp` and wrap statically wrap it in `Option::Some`.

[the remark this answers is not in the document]
gmanev7: given the fact that `Ord: PartialOrd<Self>` I would stay with the minimal traits

platform/packages/versioning/src/lib.rs
```diff
@@ -87,34 +139,34 @@ macro_rules! package_version {
...
     MapErrorFunctor: FnOnce(StdError) -> ContractError,
 {
-    load_version(storage)
-        .and_then(|current| Release::from_env().allow_software_update(&current, &new))
-        .and_then(|()| save_version(storage, &new))
-        .map(|()| Release::from_env())
+    Release::pull_prev(storage)
+        .and_then(|prev_release| {
+            let this_release = Release::this(new);
+            prev_release.update_software(this_release)
```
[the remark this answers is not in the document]
gmanev7: this call `prev_release.update_software(this_release)` more readable, thanks!

platform/packages/versioning/src/release.rs
gmanev7: I do agree

platform/packages/versioning/src/lib.rs
gmanev7: the call `prev_release.update_software(this_release)` is more readable, thanks!

platform/packages/versioning/src/release.rs
```diff
@@ -1,279 +1,268 @@
...
 
-    const DEV_RELEASE: &'static str = "dev-release";
+    //TODO delete once deliver a version with contracts that take the current release as input
+    const PREV: &'static str = "v0.7.6";
+
+    const DEV: &'static str = "dev-release";
+
+    const VOID: &'static str = "void-release";
+
+    fn this() -> Self {
```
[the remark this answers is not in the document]
gmanev7: I do not know why this issue is not listed in the `Conversation` Github PR tab - not because it's resolved, nor is outdated

Anyway, I apologize.
Renamed

platform/packages/versioning/src/release.rs
gmanev7: done

platform/packages/versioning/src/release.rs
gmanev7: done

gmanev7:
given the fact that all comments were closed I close this PR, the merge into `main ` was done manually

PR #476, opened 2025-01-15
Simplify Fraction and improve Rational.

maneva3:


platform/packages/finance/src/ratio.rs
```diff
@@ -32,30 +56,72 @@ where
...
+                    .and_then(|scaled_left_nom| {
+                        rhs.nominator
+                            .checked_mul(self.denominator)
+                            .and_then(|scaled_right_nom| {
+                                scaled_left_nom
+                                    .checked_add(scaled_right_nom)
+                                    .map(|nominator| Rational::new(nominator, common_denominator))
+                            })
+                    })
+            })
```
gmanev7, line 125: the good news is that this is already implemented in Price and battle-tested, no need to do it again

platform/packages/finance/src/ratio.rs
```diff
@@ -32,30 +56,72 @@ where
...
-        self.nominator.into()
-    }
+    type Output = Self;
 
-    fn total(&self) -> U {
-        self.denominator.into()
+    fn checked_add(self, rhs: Self) -> Option<Self::Output> {
+        self.denominator
+            .checked_mul(rhs.denominator)
+            .and_then(|common_denominator| {
```
gmanev7, line 113: would overflow in many cases when it should not, for example: with U=u16, (1,u16::MAX) + (1,u16::MAX)

gmanev7 (commented):
a partial review

platform/packages/finance/src/percent/mod.rs
```diff
@@ -78,33 +101,50 @@ impl Zero for Percent {
...
+        Self::new(percent.units(), Percent::HUNDRED.units())
     }
 }
 
-impl Ratio<Units> for Percent {
-    fn parts(&self) -> Units {
-        self.units()
+impl From<Percent> for Rational<Units> {
+    fn from(percent: Percent) -> Self {
+        Self::new(percent.units(), Percent::HUNDRED.units())
```
gmanev7, line 112: I believe we have discussed another conversion Percent -> Ratio

platform/packages/finance/src/percent/mod.rs
```diff
@@ -78,33 +101,50 @@ impl Zero for Percent {
...
     }
+}
 
-    fn total(&self) -> Units {
-        Percent::HUNDRED.units()
+impl TryFrom<Rational<Units>> for Percent {
+    type Error = Error;
```
gmanev7, line 117: not sure we need this conversion

platform/packages/finance/src/percent/mod.rs
```diff
@@ -78,33 +101,50 @@ impl Zero for Percent {
...
+impl CheckedAdd for Percent {
+    type Output = Self;
 
-    fn total(&self) -> Units {
-        Ratio::<Percent>::total(self).units()
+    fn checked_add(self, rhs: Self) -> Option<Self::Output> {
+        let lhs: Rational<Units> = self.into();
+        let rhs: Rational<Units> = rhs.into();
+
+        lhs.checked_add(rhs).and_then(|res| res.try_into().ok())
```
gmanev7, line 147: it'd be trivial to implement it through `Units::checked_add` instead

platform/packages/finance/src/percent/mod.rs
```diff
@@ -170,6 +210,22 @@ impl<'a> Sub<&'a Percent> for Percent {
     }
 }
 
+impl Div for Percent {
+    type Output = Self;
+
+    fn div(self, rhs: Self) -> Self::Output {
+        Percent(self.0 / rhs.0)
```
gmanev7, line 217: let's retain the control in our hands by checking the preconditions ourselves, add a debug_assert!(this != Zero)

platform/packages/finance/src/percent/mod.rs
```diff
@@ -170,6 +210,22 @@ impl<'a> Sub<&'a Percent> for Percent {
...
+    }
+}
+
+impl Rem for Percent {
+    type Output = Self;
+
+    fn rem(self, rhs: Self) -> Self::Output {
+        Percent(self.0 % rhs.0)
```
gmanev7, line 225: same as above

platform/packages/finance/src/price/mod.rs
```diff
@@ -318,7 +314,7 @@ where
         // Please note that Price(amount, amount_quote) is like Ratio(amount_quote / amount).
 
         Self::Output::new(self.amount, rhs.amount_quote)
-            .lossy_mul(&Rational::new(self.amount_quote, rhs.amount))
+            .lossy_mul(&Ratio::new(self.amount_quote.into(), rhs.amount.into()))
```
gmanev7, line 317: no guarantees that the price is < 1 !!!

PR #477, opened 2025-01-22
Remove json schema generation

Lockwarr:


protocol/contracts/oracle/Cargo.toml
```diff
@@ -24,43 +24,55 @@ contract = [
     "stub_swap",
     "sdk/contract",
     "dep:cosmwasm-std",
-    "dep:marketprice",
     "dep:versioning",
+    "marketprice/testing",
+    "platform/testing",
```
gmanev7, line 29: as we discussed this is not acceptable

protocol/contracts/oracle/Cargo.toml
```diff
@@ -24,43 +24,55 @@ contract = [
...
+]
+stub_alarms = [
+    "dep:oracle-platform",
+    "dep:platform",
+    "dep:serde",
+    "dep:thiserror",
+    "dep:sdk",
+    "currencies/testing",
+    "finance/testing",
+    "currency/testing",
```
gmanev7, line 39: same here

protocol/contracts/oracle/Cargo.toml
```diff
@@ -24,43 +24,55 @@ contract = [
...
+    "dep:currency",
+]
+stub_swap = [
+    "dep:oracle-platform",
+    "dep:currency",
+    "dep:tree",
+    "dep:sdk",
+    "dep:serde",
+    "dep:thiserror",
+    "currencies/testing",
```
gmanev7, line 54: and here

gmanev7 (commented):
the 'testing' feature is acceptable as a dependency only in the `[dev-dependencies]` and the `testing` feature of the current contract, it any

protocol/contracts/oracle/src/lib.rs
```diff
@@ -11,7 +11,7 @@ pub mod result;
 #[cfg(feature = "contract")]
 pub mod state;
 pub mod stub;
-#[cfg(any(feature = "testing", test))]
+#[cfg(feature = "stub_swap")]
 pub mod test_tree;
```
KirilMihaylov, line 15: Should stay gated.

protocol/packages/dex/Cargo.toml
```diff
@@ -23,7 +23,7 @@ finance = { workspace = true }
 oracle = { workspace = true, features = ["stub_swap"] }
 timealarms = { workspace = true, optional = true, features = ["stub"] }
 platform = { workspace = true }
-sdk = { workspace = true, features = ["schema"] }
+sdk = { workspace = true}
```
KirilMihaylov, line 26:
```suggestion
sdk = { workspace = true }
```

KirilMihaylov (changes requested):

protocol/contracts/oracle/src/tests/mod.rs
```diff
@@ -101,6 +101,7 @@ pub(crate) fn dummy_feed_prices_msg(
     }
 }
 
+#[cfg(test)]
```
gmanev7, line 104: I believe this is redundant, and the same on line 31 because the whole module is defined behind `cfg(test)`

protocol/packages/dex/Cargo.toml
```diff
@@ -12,6 +12,9 @@ combinations = [
     { tags = ["ci", "@agnostic"], include-rest = true },
 ]
 
+[package.metadata.cargo-udeps]
+ignore = { development = ["currencies"] }
+
```
gmanev7, line 17: a similar question to the one for the `oracle`:
Why do we need it as a dev dependency and ignore it at the same time?

protocol/packages/swap/Cargo.toml
```diff
@@ -13,6 +13,9 @@ combinations = [
 ]
 feature-groups = { dex = { members = ["dex-astroport_main", "dex-astroport_test", "dex-osmosis", "dex-test_impl"], at-least-one = true, mutually-exclusive = true } }
 
+[package.metadata.cargo-udeps]
+ignore = { development = ["currencies"] }
```
gmanev7, line 17: a similar question to the above one

gmanev7 (commented):
most of the comments are around the dev dependency ingore

protocol/packages/dex/Cargo.toml
Lockwarr: it was needed due to oracle/src/test_tree.rs being gated behind the `testing` feature. I couldn't find direct usage of the currencies dependency inside the `dex` or `swap` packages, however it was required for the tests.

I changed the feature gate for test_tree from `testing` to `stub_swap` and removed this ignore

protocol/packages/swap/Cargo.toml
Lockwarr: it was needed due to oracle/src/test_tree.rs being gated behind the `testing` feature. I couldn't find direct usage of the currencies dependency inside the `dex` or `swap` packages, however it was required for the tests.

I changed the feature gate for test_tree from `testing` to `stub_swap` and removed this ignore

protocol/contracts/oracle/Cargo.toml
```diff
@@ -24,43 +27,62 @@ contract = [
...
 currencies = { workspace = true, features = ["testing"] }
 currency = { workspace = true, features = ["testing"] }
 finance = { workspace = true }
 platform = { workspace = true, features = ["testing"] }
-schema = { workspace = true }
-oracle = { workspace = true, features = ["contract"] }
 marketprice = { workspace = true, features = ["testing"] }
-tree = { workspace = true, features = ["schema", "testing"] }
+tree = { workspace = true, features = ["testing"] }
+oracle = { workspace = true, features = ["contract"] }
```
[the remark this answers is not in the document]
Lockwarr: yes, we need it for imports used in tests which are gated behind features like stub_alarms or other

protocol/contracts/oracle/Cargo.toml
```diff
@@ -14,6 +14,9 @@ combinations = [
 ]
 feature-groups = { contract-and-testing = { members = ["contract", "testing"], at-least-one = true, mutually-exclusive = false }, stubs ={ members = ["stub_alarms", "stub_price", "stub_swap"], at-least-one = false, mutually-exclusive = false} }
 
+[package.metadata.cargo-udeps]
+ignore = { development = ["oracle"] }
+
```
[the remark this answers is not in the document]
Lockwarr: i've added a comment, let me know what you think

protocol/contracts/oracle/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub mod result;
 #[cfg(feature = "contract")]
 pub mod state;
 pub mod stub;
-#[cfg(any(feature = "testing", test))]
+#[cfg(any(feature = "stub_swap", test))]
```
KirilMihaylov, line 11: I don't think this is a good idea. This will include a testing module in production, if I am not mistaken. AFAIR, the stubs are used by the rest, exposing the API in a sane way.

KirilMihaylov (changes requested):

protocol/contracts/oracle/src/lib.rs
Lockwarr: what would be best in your opinion? I've discussed with @gmanev7 that it may be better to have cfg `stub_swap` or `stub_swap, test`. 

If I leave it with `testing, test` as it was, then the dev-dependency checks for dex and swap packages start to complain about `unused dependencies` and I have to also ignore `currencies` in the respective package, as the test_tree.rs is checked when running the dex/swap packages' checks

KirilMihaylov: If it is required but only for testing, then `#[cfg(any(all(feature = "stub_swap", feature = "testing"), test))]` should suffice, yes?

Lockwarr: I also tried this, but I still have to ignore `currencies` in the dex and swap packages, should I proceed with the ignoring? If yes, might as well remove the `stub_swap `feature and leave only `testing` like it was before

PR #479, opened 2025-01-23
refactor(protocol/oracle): Refactor alarms iterator implementation.

KirilMihaylov:


protocol/contracts/oracle/src/contract/alarms/iter.rs
```diff
@@ -96,24 +91,37 @@ where
...
+            result = alarms_iter.next();
+
+            if result.is_some() {
+                break;
+            }
+
+            #[cfg(debug_assertions)]
+            {
+                self.alarm_iter = None;
+            }
```
gmanev7, line 107: frankly, I cannot find a reason for having this assignment, we do not trust the compiler that the next assignment will be always completed???

protocol/contracts/oracle/src/contract/alarms/iter.rs
```diff
@@ -96,24 +91,37 @@ where
...
+                Some(Ok(iter)) => Some(iter),
+                Some(Err(error)) => {
+                    result = Some(Err(error));
+
                     None
-                } else {
-                    self.next_alarm()
                 }
+                None => None,
             };
```
gmanev7, line 117: It seems to me, that the same processing of `self.next_alarms()` is done in the `fn new`.  Could you try to reuse it?

protocol/contracts/oracle/src/contract/alarms/iter.rs
```diff
@@ -96,24 +91,37 @@ where
...
         }
+
+        debug_assert!(matches!(
+            (&result, &self.alarm_iter),
+            (Some(Ok(_)), Some(_)) | (Some(Err(_)), None) | (None, None)
```
gmanev7, line 122: I'd be grateful if you merged the disjunctions using `Option::is_some` or similar functions. The final result would be more readable and, more importantly, would not skip any value combination.

PR #480, opened 2025-01-23
fix(tools/cargo-each): Correct `IterOrElseIter` implementation to not poll primary iterator after exhaustion.

KirilMihaylov:


tools/cargo-each/src/iter_or_else_iter.rs
```diff
@@ -1,40 +1,90 @@
...
+        }
 
-            Some(value)
+        #[expect(if_let_rescope)]
+        if let Some(ref mut alt_iter) = self.alt_iter {
+            alt_iter.next()
         } else {
-            self.alt_iter.as_mut().and_then(Iterator::next)
+            None
         }
```
gmanev7, line 48: IMHO this code should be refactored. A couple of observations:
- how do we select which iterator to use? Is not that based on the both iterators passed as input?
- if yes, I argue that this struct is unnecessary and complicated
- I'd suggest a function that returns an `impl Iterator` based on the content of the first input iterator - either `alt_iter` if `iter.next().is_none()`, or str::iter::chain(iter_head, iter)

KirilMihaylov: To answer the questions in order:
* Yes, it is based solely on the inputs. The design I am striking for is as-pure-as-possible.
* My argument here is that I am following the standard library's design decision to make iterators lazily evaluated, thus the operations until the first poll should be pure/not causing side effects.
* I could always use the `Peekable` adaptor and decide based on it and stuff the iterators in `Either`, but as already mentioned, this would result in impurity.

I would love to hear it, if you have any suggestions on how to achieve that same result with a simpler solution.

gmanev7: I have it.

The complexity comes from the extensive interior mutability. Instead, keep the two iterators straight, not in Option<I>, and add a new member variable, the actual Iterator, to iterate over. Initialize it on the first Iter::next() call.

gmanev7: hm, actually you may need `Option<I>`-s ... the idea still remains the same

KirilMihaylov: The two types can be different tho, and are from the perspective of the type system. That's why I do these gymnastics too. One way I can do it is with two boolean fields, but that effectively diminishes the advantages otherwise I get from the type system.

KirilMihaylov: If I was able to internally use something like `Either`, with a third variant marking the undecided state, I would have. I did try that approach first, but that in turn required me to use `Option` and `Option::take` in order to move the appropriate iterator into the now decided state. I know how to do that with unsafe Rust, but that would be an overkill if nothing else.

gmanev7: The trade-off you make to avoid first-element eager evaluation is too expensive - poor readability.
Re-think again the procs and cons, take a look at this code 
https://doc.rust-lang.org/src/core/iter/adapters/chain.rs.html#83, and decide wisely.

PR #490, opened 2025-01-29
Release query

gmanev7:


platform/packages/versioning/src/release/query.rs
```diff
@@ -0,0 +1,52 @@
...
+/// A common versioning API of each protocol package
+#[derive(Serialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub enum ProtocolPackage {
+    /// Query the protocol package for its release.
+    ///
+    /// The result is [versioning::ProtocolPackageRelease]
+    #[serde(rename = "protocol_package_release")]
+    Release {},
+}
```
KirilMihaylov, line 28: These two could be combined into one `enum`, named `Release` with the variants being `Platform` and `Protocol`.  
Also, as far as I can see at least, they don't do any good being public. If they were perhaps used in tests to verify that the query is actually of the same form as the contract's message, then I would agree they should be public.

protocol/contracts/lease/src/api/query.rs
```diff
@@ -19,12 +19,18 @@ pub use opened::ClosePolicy;
...
     ///
+    /// Return a [StateResponse]
+    ///
     /// The value is meaningfull only if the lease is in Opened state.
-    #[serde(default, rename = "due_projection_secs")]
-    pub due_projection: Seconds,
+    State {
+        #[serde(default, rename = "due_projection_secs")]
+        due_projection: Seconds,
+    },
```
KirilMihaylov, line 31: Just want to make sure, do we know that who the clients of this API are, e.g. the web app? My concern is that we may break some of our systems, so we should check up with them.

KirilMihaylov (changes requested):
Just one comment about minor refactoring in the `platform/versioning/release/query` module and one concern about changing the API.

Otherwise, LGTM.

protocol/contracts/lease/src/api/query.rs
gmanev7: I did that this morning - all known clients go through the `nolusjs` ... so that is why I resorted to this breaking API change

platform/packages/versioning/src/release/query.rs
gmanev7: Indeed, they could, but they would represent the abstractions poorly. What if we need another message call for either or both of them?
>  they don't do any good being public
Thanks for catching this. Actually, I intended to add such tests ... and I'll do before merging it.

protocol/contracts/lease/src/api/query.rs
KirilMihaylov: I see now. I did not realize we're talking about this same change.

platform/packages/versioning/src/release/query.rs
KirilMihaylov: > Indeed, they could, but they would represent the abstractions poorly. What if we need another message call for either or both of them?

The messages are only used for serialization, so I don't think they were abstracting anything to begin with, the functions were. The messages weren't even used as parameters to the functions, so effectively they weren't part of the abstraction itself.

gmanev7: They **are** abstractions, though the cosmwasm model makes them impossible to use on the contracts side.

PR #492, opened 2025-01-30
feat(platform/versioning): Add unified migration message wrapper type.

KirilMihaylov:


protocol/contracts/leaser/src/migrate.rs
```diff
@@ -341,12 +342,15 @@ mod test {
...
-            ),
+    fn migrate_msg() -> ProtocolMigrationMessage<MigrateMsg> {
+        ProtocolMigrationMessage {
+            to_release: const {
+                ProtocolPackageReleaseId::sample(
+                    ReleaseId::new_test("v0.7.6"),
+                    ReleaseId::new_test("v0.0.5"),
+                )
+            },
+            message: MigrateMsg {},
```
gmanev7, line 353: Is there any reason this is not const as you did in the leaser.rs tests?

tests/src/common/test_case/mod.rs
```diff
@@ -1,3 +1,10 @@
+use super::{
+    lease::{
+        InitConfig, Instantiator as LeaseInstantiator, InstantiatorAddresses,
+        InstantiatorConfig as LeaseInstantiatorConfig,
+    },
+    mock_app, CwContractWrapper, ADMIN,
+};
```
gmanev7, line 7: this block should stay the last in the list of all imports

gmanev7 (commented):
good job!
Just a few minor items

platform/contracts/admin/src/msg.rs
```diff
@@ -22,13 +22,10 @@ pub struct InstantiateMsg {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, JsonSchema)]
-#[serde(rename_all = "snake_case", deny_unknown_fields)]
```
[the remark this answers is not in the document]
KirilMihaylov: No, it shouldn't needed in this case as it delegates to the inner type via `#[serde(transparent)]`.

protocol/contracts/leaser/src/migrate.rs
KirilMihaylov: No particular one. Just didn't process it properly in my brain. That's why the scope of the `const` block is narrower. I shall correct this now.

tests/src/common/test_case/mod.rs
KirilMihaylov: My IDE sometimes has those hick-ups. I must have missed correcting it.

PR #493, opened 2025-01-31
refactor(platform/admin): Make trait methods and message queueing functions fallible.

KirilMihaylov:


platform/contracts/admin/src/endpoints.rs
```diff
@@ -295,8 +295,9 @@ fn deregister_protocol(
                 release: ReleaseId::VOID,
             }
             .store(storage)
```
gmanev7, line 297: this could and should be removed with another PR

gmanev7 (approved):

PR #494, opened 2025-01-31
feat(tools/json-value): Introduce an equivalent of `serde_json::Value` that does not include floating-point numbers.

KirilMihaylov:


tools/json-value/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum JsonValue {
+    Null,
+    Bool(bool),
+    I64(i64),
+    U64(u64),
+    String(String),
+    Array(Vec<JsonValue>),
+    Object(Vec<(String, JsonValue)>),
```
KirilMihaylov, line 16: I specifically did not use a map type, as it would introduce changes on overlapping keys and also might reorder the keys. Whether such behavior is acceptable or not is out of the scope of the library, as such it is delegated to the type that is later deserialized.

tools/json-value/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl Serialize for JsonValue {
```
gmanev7, line 23: What is the motivation of providing our ser/de implementations?

KirilMihaylov: I was taking the example of the original `serde_json::Value`.

tools/json-value/src/lib.rs
gmanev7: nice

tools/json-value/src/lib.rs
gmanev7: - 200 to 300 LoC would justify a "non-hacky" solution, though, be it an extra type used in the `Object` variant or deserialized into.

gmanev7: no custom ser-de, no thorough testing either

KirilMihaylov: Actually I would still need manual deserialization for the type that I'll be putting into `Object` anyway. There won't really be a way to avoid it, just minimize it.

About the tests, I'd like to emphasize on the nature of this code and that once written, there won't be any need to ever touch it again unless for some reason CosmWasm and we both decide that we'll be accepting floating-point operations. If that ever happens, we could quite easily delete this crate and opt for using the original `serde_json`.  
To reiterate, there won't be any maintenance required, ever, unless a bug is found, which with this thorough testing I'd like to believe is highly unlikely.

gmanev7 (approved):

PR #495, opened 2025-01-31
refactor(platform): collect all balances with separate queries per cu…

gmanev7:
…rrency

- refactor(currency): intro `fn Group::currencies`


platform/packages/currency/src/test/group.rs
```diff
@@ -25,6 +25,19 @@ impl Group for SuperGroup {
...
+    fn currencies() -> impl Iterator<Item = CurrencyDTO<Self>> {
+        [
+            SuperGroupTestC1::definition().dto(),
+            SuperGroupTestC2::definition().dto(),
+            SuperGroupTestC3::definition().dto(),
+            SuperGroupTestC4::definition().dto(),
+            SuperGroupTestC5::definition().dto(),
+        ]
+        .into_iter()
+        .map(ToOwned::to_owned)
```
KirilMihaylov, line 37: Please refer to previous comment.
```suggestion
        .copied()
```

platform/packages/currency/src/test/group.rs
```diff
@@ -164,6 +177,15 @@ impl Group for SubGroup {
     const DESCR: &'static str = "sub_group";
     type TopG = SuperGroup;
 
+    fn currencies() -> impl Iterator<Item = CurrencyDTO<Self>> {
+        [
+            SubGroupTestC6::definition().dto(),
+            SubGroupTestC10::definition().dto(),
+        ]
+        .into_iter()
+        .map(ToOwned::to_owned)
```
KirilMihaylov, line 186: Please refer to previous comment.
```suggestion
        .copied()
```

platform/packages/platform/src/bank.rs
```diff
@@ -122,16 +123,24 @@ impl BankAccountView for BankView<'_> {
...
+                    |err| Some(Err(err)),
+                    |ref cw_balance| {
+                        if cw_balance.amount == Uint128::zero() {
+                            None
+                        } else {
+                            Some(coin_legacy::from_cosmwasm_any::<G, _>(
+                                cw_balance,
+                                cmd.clone(),
+                            ))
+                        }
```
KirilMihaylov, line 138: The convenience method `bool::then` can be used for the purposes on conditionally constructing an option.
```suggestion
                        (cw_balance.amount != const { Uint128::zero() }).then(|| {
                            coin_legacy::from_cosmwasm_any::<G, _>(cw_balance, cmd.clone())
                        })
```

platform/packages/platform/src/coin_legacy.rs
```diff
@@ -45,19 +45,23 @@ where
...
-    coin: CosmWasmCoin,
+/// Transform CW coin to Nolus coin and then return the [WithCoin] result
+///
+/// If seeking for the corresponding Nolus coin is not successfull a [V::Error] is returned.
+/// Otherwise, an `v.on(Nolus_coin)` is returned
+pub(crate) fn from_cosmwasm_any<VisitedG, V>(
+    coin: &CosmWasmCoin,
     v: V,
-) -> Option<WithCoinResult<VisitedG, V>>
+) -> StdResult<WithCoinResult<VisitedG, V>, Error>
```
KirilMihaylov, line 55:
```suggestion
) -> Result<WithCoinResult<VisitedG, V>>
```

platform/packages/currency/src/platform.rs
```diff
@@ -90,6 +90,13 @@ impl Group for PlatformGroup {
     const DESCR: &'static str = "platform currencies";
     type TopG = Self;
 
+    fn currencies() -> impl Iterator<Item = CurrencyDTO<Self>> {
+        [Nls::definition().dto(), Stable::definition().dto()]
```
KirilMihaylov, line 94: We should investigate whether there is anything actually requiring (1) a reference to the `CurrencyDTO`, as it already holds a reference inside itself causing double indirection, (2) a non-`'static` enforced type reference to it, for the purposes of sanity checking, (3) do we actually need `CurrencyDef::definition`, as the types to my knowledge aren't type erased which we also actively discourage anyway.

Since this change list already requires changes to the `currencies`'s build process and output, I think this is a great time to clean up this, in my eyes unnecessary, tech debt.

KirilMihaylov (changes requested):
Some minor issues when it comes to the proposed changes.

Since these changes require changes to the `currencies`'s build process and output, I feel compelled to point out that there might be areas for improvement that we could address while we'll already be making changes to it.

platform/packages/currency/src/platform.rs
```diff
@@ -90,6 +90,13 @@ impl Group for PlatformGroup {
...
+            .into_iter()
+            .map(ToOwned::to_owned)
```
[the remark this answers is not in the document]
gmanev7: `Copied` is our guy

platform/packages/platform/src/bank.rs
gmanev7: thanks, did not recall to it

platform/packages/currency/src/platform.rs
gmanev7: I've been looking for a while into optimizing these three types `CurrencyDef`, `CurrencyDTO`, and `Definition`.

KirilMihaylov (approved):

platform/packages/currency/src/group/find.rs
```diff
@@ -0,0 +1,23 @@
...
+            Ok(ref _result) => {
+                break;
+            }
+            Err(f) => {
+                result = next.find_map(f);
+            }
+        }
+        may_next = next.next();
+    }
+    result
```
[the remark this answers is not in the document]
gmanev7: The suggested change breaks the 'single-return-of-function' practice. In addition, the code is less readable than before.

platform/packages/currency/src/group/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    where
+        FindMap: FindMapT<TargetG = Self>;
+}
+
+pub type MaybeAnyVisitResult<VisitedG, V> = Result<<V as AnyVisitor<VisitedG>>::Outcome, V>;
+
+pub trait FilterMapT {
+    type VisitedG: Group;
+
+    type Outcome;
```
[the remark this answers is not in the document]
gmanev7: Recently, all such 'currency-resolving' traits have had `type Output` and `type Error`. To eliminate potential confusion, I have chosen `Outcome`. We can rename it once we finish merging the two types.

platform/packages/currency/src/pairs/find.rs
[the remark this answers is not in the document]
gmanev7: done only the 1st part

PR #497, opened 2025-02-06
refactor(marketprice): use a higher abstraction for the primary key

gmanev7:


KirilMihaylov (approved):
LGTM.

KirilMihaylov:
As I was reminded by this PR that we use `HashMap`, I wanted to point out that we could checkout for example on the test-nets performance with an alternative hasher.  
The default one is chosen a bit more conservatively, to protect unsuspecting users from possible DoS/HashDoS attacks. The documentation of the standard library mentions that one can use an alternative hasher.  
Since we don't allow arbitrary data from outside, we could easily take the liberty of changing it.  
The `ahash` crate, claiming to be the fastest \[1\], seems to already be used within out code base judging by the lockfiles. The other option is to try out `rustc`'s hasher, `FxHash`, provided in the `rustc-hash` crate.

\[1\] [Comparison of performance of hash functions on strings](https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md#Speed).

gmanev7:
> As I was reminded by this PR that we use `HashMap`, I wanted to point out that we could checkout for example on the test-nets performance with an alternative hasher. The default one is chosen a bit more conservatively, to protect unsuspecting users from possible DoS/HashDoS attacks. The documentation of the standard library mentions that one can use an alternative hasher. Since we don't allow arbitrary data from outside, we could easily take the liberty of changing it. The `ahash` crate, claiming to be the fastest [1], seems to already be used within out code base judging by the lockfiles. The other option is to try out `rustc`'s hasher, `FxHash`, provided in the `rustc-hash` crate.
> 
> [1] [Comparison of performance of hash functions on strings](https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md#Speed).

Thanks for providing the info. Given the amount of data—in our case, the number of observations multiplied by the CosmosSDK address length—we'll have to assess carefully whether a faster algorithm is needed.

PR #502, opened 2025-02-11
refactor: Omit reply data on migration endpoints.

KirilMihaylov:


protocol/contracts/void/src/lib.rs
```diff
@@ -29,7 +28,7 @@ pub fn migrate(
     _env: Env,
     EmptyMsg {}: EmptyMsg,
 ) -> Result<CwResponse, platform::error::Error> {
-    response::response(ReleaseId::VOID)
```
gmanev7, line 32: please check if `ReleaseId::VOID` is still used, and delete if not
I suppose there might be places it is still used but they should be reviewed again

gmanev7: for example, in `versioning::update_software_and_storage` it should be removed

KirilMihaylov: You are correct about the standalone function, but the `ReleaseId::VOID` is still used, namely for when voiding protocols. I think I will open a separate PR to remove it to not block these.

PR #505, opened 2025-02-11
refactor(platform/versioning): Minor adjustments and cleanup.

KirilMihaylov:


platform/packages/versioning/src/protocol/mod.rs
```diff
@@ -26,10 +26,13 @@ impl Release {
     }
 
     fn check_release_match(&self, target: &Id) -> Result<(), Error> {
-        if &self.id == target {
+        if self.id == *target {
```
gmanev7, line 29: I am curios why this is favorable to you.

platform/packages/versioning/src/release/mod.rs
```diff
@@ -42,15 +42,13 @@ pub struct ProtocolPackageReleaseId {
...
+    pub const VOID: Self = Self {
+        software: Id::VOID,
+        protocol: Id::VOID,
+    };
 
-    #[cfg(feature = "testing")]
-    pub const fn sample(software: Id, protocol: Id) -> Self {
+    #[inline]
+    pub const fn new(software: Id, protocol: Id) -> Self {
         Self { software, protocol }
```
gmanev7, line 52: we do not want to create instances internally, they should only be created via deserialization

platform/packages/versioning/src/software/mod.rs
```diff
@@ -92,10 +92,13 @@ impl PackageRelease {
     }
 
     fn check_release_match(&self, target: &Id) -> Result<(), Error> {
-        if &self.id == target {
+        if self.id == *target {
```
gmanev7, line 95: again, a question on the rationale here

platform/packages/versioning/src/protocol/mod.rs
KirilMihaylov: It skips the middle man and looks somewhat clearer to me, but I understand that's somewhat of a personal preference, so I'll be fine with reverting.

platform/packages/versioning/src/release/mod.rs
KirilMihaylov: The whole thing is that I can't rely on deserialization as the admin itself needs to be able to construct them for each protocol. This is the foundation for the next PR which does that.

gmanev7: ah, ok, I missed this scenario

PR #506, opened 2025-02-12
refactor(platform/admin): Build migration messages packed with release ID per each contract and protocol.

KirilMihaylov:


platform/contracts/admin/src/contracts/mod.rs
```diff
@@ -106,38 +114,67 @@ where
...
+    pub migrate_message: JsonValue,
+    pub post_migrate_execute: Option<ExecuteSpec>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[repr(transparent)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct ExecuteSpec {
+    pub message: JsonValue,
 }
```
gmanev7, line 126: why do we need to wrap it? The migration message is not

platform/contracts/admin/src/endpoints.rs
```diff
@@ -67,7 +67,9 @@ pub fn migrate(
         .and_then(|()| {
             //TODO remove the check!!!
             check_release_label(ReleaseId::VOID, to_release.clone())
```
gmanev7, line 69: I believe this check is stale, please remove

protocol/Cargo.toml
```diff
@@ -47,6 +47,7 @@ tree = { path = "../platform/packages/tree", default-features = false }
 versioning = { path = "../platform/packages/versioning", default-features = false }
 
 # Tools Packages
+json-value = { path = "../tools/json-value" }
```
gmanev7, line 50: + default-features = false

platform/contracts/admin/src/contracts/impl_mod.rs
```diff
@@ -10,19 +16,20 @@ use crate::{
...
-    ContractsTemplate, Granularity, HigherOrderOption, HigherOrderPlatformContracts,
+    ContractsTemplate, ExecuteSpec, Granularity, HigherOrderOption, HigherOrderPlatformContracts,
     HigherOrderProtocolContracts, HigherOrderType, MigrationSpec, PlatformContractAddresses,
     PlatformExecute, PlatformMigration, Protocol, ProtocolContractAddresses, ProtocolExecute,
     ProtocolMigration, Protocols,
 };
 
 pub(crate) fn migrate(
     storage: &mut dyn Storage,
+    to_software_release: ReleaseId,
```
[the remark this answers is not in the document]
KirilMihaylov: It is common for both platform and protocol though. If it weren't, I would have agreed but as it stands that would imply that the protocol is just an extension over the platform, not it's own separate entity.

platform/contracts/admin/src/contracts/mod.rs
KirilMihaylov: This is needed in order to distinguish the two variants:
* No execute should be dispatched,
* Execute should be dispatched with `null` as the message.
Because of how `Option<T>` is serialized, namely `null` or the serialization of `T`, there is no way to distinguish them which might lead to errors down the road.

KirilMihaylov: The migration message actually is already wrapped the same way, it was just covered up by the necessity of storing the code id and post-migration execute message.

platform/contracts/admin/src/endpoints.rs
KirilMihaylov: As I mentioned in one of the previous PRs, I have taken it into account. [Changed here.](https://github.com/nolus-protocol/nolus-money-market/compare/KirilMihaylov/refactor-platform-admin-build-messages...KirilMihaylov/refactor-platform-admin-migrate-via-sudo-migrate#diff-a0c59fd77c574dd7635e67dedd4ffcf59d9a943da09efd1cc6ec9548ce81b3afL68-L73)

PR #510, opened 2025-02-27
refactor(oracle): rename price alarms events

Lockwarr:


protocol/contracts/oracle/src/contract/mod.rs
```diff
@@ -173,7 +173,7 @@ pub fn sudo(
 // TODO: compare gas usage of this solution vs reply on error
 #[entry_point]
 pub fn reply(deps: DepsMut<'_>, _env: Env, msg: Reply) -> Result<CwResponse, PriceCurrencies> {
-    const EVENT_TYPE: &str = "market-alarm";
+    const EVENT_TYPE: &str = "results-pricealarm";
```
gmanev7, line 176: [this and the other type] let's have the same word to start with, as you have suggested, "pricealarm", and then an optional extra word. This way we would go from the more general to more specific. As a result, we would get "wasm-pricealarm", and "wasm-pricealarm-delivery" events.

Unfortunately, cosmwasm or cosmwasm-sdk merges all events of the same type and we observe a giant event that includes multiple repeatable keys.

gmanev7 (commented):
just a suggestion on the naming, otherwise LGTM

gmanev7 (approved):

PR #511, opened 2025-03-03
chore(container/check): Add necessary environment variables.

KirilMihaylov:
Depends on #512 / Edition migration.

### DIFF
--- a/check.Containerfile
+++ b/check.Containerfile
@@ -64,6 +64,14 @@ USER 1000:1000
 
 ENTRYPOINT ["/check.sh"]
 
+ENV SOFTWARE_RELEASE_ID="dev"
+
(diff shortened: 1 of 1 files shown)


check.Containerfile
```diff
@@ -64,6 +64,14 @@ USER 1000:1000
 
 ENTRYPOINT ["/check.sh"]
 
+ENV SOFTWARE_RELEASE_ID="dev"
```
Lockwarr, line 67: why did we choose 'dev'? wouldn't 'ci' be a more generic option

KirilMihaylov: It's not used anywhere really. This one is just for convenience when you'd want to run it locally.

gmanev7 (approved):

PR #512, opened 2025-03-03
refactor: Migrate Rust edition.

KirilMihaylov:


platform/packages/lpp/src/stub.rs
```diff
@@ -12,19 +12,19 @@ use crate::{
...
-    querier: QuerierWrapper<'a>,
-    env: &'a Env,
+    querier: QuerierWrapper<'querier>,
+    env: &'env Env,
 }
 
-impl<'a> Stub<'a> {
-    pub(crate) fn new(lpp: Addr, querier: QuerierWrapper<'a>, env: &'a Env) -> Self {
+impl<'querier, 'env> Stub<'querier, 'env> {
+    pub(crate) fn new(lpp: Addr, querier: QuerierWrapper<'querier>, env: &'env Env) -> Self {
```
gmanev7, line 22: could you share the rules you follow when decide to split vs. merge lifetimes? In an another comment I am going to point out merging.

protocol/contracts/oracle/src/contract/oracle/feed/mod.rs
```diff
@@ -47,15 +47,14 @@ where
     BaseG: Group + MemberOf<PriceG>,
     Observations: ObservationsReadRepo<Group = PriceG>,
 {
-    pub fn all_prices_iter<'self_, 'iterator, I>(
-        &'self_ self,
+    pub fn all_prices_iter<I>(
+        &self,
```
gmanev7, line 51: here you have decided to move off the named lifetimes ... does it provide more readability?

scripts/check/lint.sh
```diff
@@ -32,6 +32,7 @@ append_lint_flags() {
   # Allowed Lints
   "${@:?}" \
     -- \
+    --forbid "deprecated-safe" \
```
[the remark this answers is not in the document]
KirilMihaylov: You're probably thinking of `deprecated-safe-2024`. Source: https://doc.rust-lang.org/stable/rustc/lints/groups.html

platform/packages/lpp/src/stub.rs
KirilMihaylov: Those two lifetimes are independent of each other in the sense that they do not come from a common origin, and in contrast to how it is currently done in `currencies`' build script, they don't need to be unified into an intersection, `'r, 't, 'intersection where 'r: 'intersection, 't: 'intersection`.

In theory, all unrelated parameters should use their own lifetime and then be constrained with trait bounds.

protocol/contracts/oracle/src/contract/oracle/feed/mod.rs
KirilMihaylov: Here my reasoning is that there is only one lifetime parameter that can be elided and due to the nature of return position `impl Trait` (RPIT) in the 2024 edition it is no longer even required to be specified how long the iterator, `I`, will live because it is implicitly captured.

gmanev7: right, however the previous change was also yours and the rule there was to explicitly capture because when we make changes, e.g. add inner or outer template parameter, that would be captured, too

KirilMihaylov: Yes, you are right. I did make that point but in this case it does take a reference to `self` which would cover all outer-scoped generics already and I feel like when it has only one parameter that has any generics, type or lifetime, it would be acceptable to keep it simpler.  
I'm starting to come around to be on your side that there are things that should be developer's responsibility when it comes to this specific cases. :smile:

gmanev7 (approved):

PR #517, opened 2025-03-11
refactor(platform/admin): Query old package versions.

KirilMihaylov:


platform/packages/versioning/src/release/mod.rs
```diff
@@ -13,6 +13,10 @@ pub trait UpdatablePackage
 where
     Self: Sized,
 {
+    type VersionQuery: Serialize + 'static;
```
gmanev7, line 17: let's follow the convention to put all bounds in the `where` statement

platform/contracts/admin/src/contracts/impl_mod.rs
```diff
@@ -47,17 +47,20 @@ pub(crate) fn execute(
...
     to_release: Package::ReleaseId,
     migration: MigrationSpec,
 ) -> Result<()>
 where
-    Package: UpdatablePackage + Serialize,
+    Package: UpdatablePackage + Serialize + DeserializeOwned,
     Package::ReleaseId: Serialize,
 {
+    let migrate_from: Package =
+        querier.query_wasm_smart(address.clone(), &Package::VERSION_QUERY)?;
```
gmanev7, line 62: please merge this into the next expression, right before its use

platform/contracts/admin/src/contracts/impl_mod.rs
```diff
@@ -47,17 +47,20 @@ pub(crate) fn execute(
...
+
```
gmanev7, line 63: [lines 74-86]: extract a function `migrate_contract` next to `execute_contract`

gmanev7, line 63: [lines 64-87] Though the order of "post-migration-to-migration" is irrelevant to the correctness here, the reader is left with the impression that the post-migration is done *before* the actual migration.

KirilMihaylov: I specifically have ordered them in such a manner not because I wanted to confuse the reader but rather because it will introduce an unnecessary cloning.  
After all, we're not Haskell and sadly don't have infinite memory. ;)

Jokes aside, I think it's better for the reader to pay attention there are two different queues than to do what I would consider junior dev mistake.

platform/packages/versioning/src/release/mod.rs
KirilMihaylov: I specifically didn't do that. This causes `rustc` to spit out this error: 
```
overflow evaluating the requirement `<ProtocolPackageRelease as UpdatablePackage>::VersionQuery == _`
```

gmanev7: can you provide the full error? I have done it locally without compiler issues.

platform/contracts/admin/src/contracts/impl_mod.rs
gmanev7: It is not true that there would be extra cloning. The switch went smoothly.

platform/packages/versioning/src/release/mod.rs
KirilMihaylov: That was literally all I got. This was reported to me by `rust-analyzer`.  
What I did was:
```rust
    type VersionQuery
    where
        Self::VersionQuery: Serialize + 'static;
```

platform/contracts/admin/src/contracts/impl_mod.rs
KirilMihaylov: [This line here](https://github.com/nolus-protocol/nolus-money-market/pull/517/commits/8b2fcf3f96d2d55a4c123e0a5fa0c048a9f778d3#diff-2f4a5eb54d3d02943e26d103b4f1133cf26b27ee82733dcd73d80d64df833eb3R68) is what I meant by an unnecessary clone. It **will** clone it and then *conditionally* either use it or just drop it.

gmanev7: right, and here is where we have to decide whether to sacrifice readability for this extra 20 bytes clone

KirilMihaylov: Okay then, I just wanted to make sure we're on the same page that it will introduce waste.

gmanev7 (approved):

PR #518, opened 2025-03-11
feat(swap): parameterize the min output amount in swaps

gmanev7:


protocol/contracts/oracle/src/api/swap.rs
```diff
@@ -21,15 +21,13 @@ where
 {
     /// Provides a path in the swap tree between two arbitrary currencies
     ///
```
Lockwarr, line 23: why was the comment changed? should the enum be modified to return the desired Vec

Lockwarr (approved):

protocol/contracts/oracle/src/api/swap.rs
gmanev7: This is not a complete remote method call definition. This is only a definition of query messages. The way to communicate the expected result messages is the doc.

gmanev7: The doc has been changed because the typedef SwapPath was removed with this PR.

protocol/packages/swap/src/osmosis/testing.rs
```diff
@@ -27,12 +27,11 @@ impl ExactAmountInSkel for Impl {
             testing::pattern_match_else(type_name::<RequestMsg>())
         };
 
-        assert_eq!({ token_out_min_amount }, "1");
-
         let token_in = testing::parse_dex_token(&token_in.amount, &token_in.denom);
 
         SwapRequest {
             token_in,
+            min_token_out: Amount::from_str(&token_out_min_amount).expect("valid amount integer"),
```
KirilMihaylov, line 34: [nit] Calling `.parse()` should suffice too and would simplify the expression.

protocol/contracts/oracle/src/api/contract.rs
```diff
@@ -135,7 +135,7 @@ where
 
     /// Provides a path in the swap tree between two arbitrary currencies
     ///
-    /// Returns `oracle::api::swap::SwapPath`
+    /// Returns [Vec<crate::api::SwapTarget>]
```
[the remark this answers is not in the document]
gmanev7: thanks

protocol/packages/swap/src/osmosis/testing.rs
gmanev7: thanks, one `use` less

Issue #70, opened 2023-04-04
[Lease] Validate time and market price alarms' sender in all states


gmanev7:
Currently, it is done for the Active state. 

For supporting DEX-related operations, we have extended the Lease State Machine with operations triggered by time alarms. There we missed adding alarms' sender validation.

PR #519, opened 2025-03-12
Denislav/validate alarms sender

Lockwarr:


protocol/contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -75,12 +75,7 @@ impl Active {
         env: &Env,
         info: MessageInfo,
     ) -> ContractResult<Response> {
-        // TODO ref. the TODO in try_on_time_alarm
-        if !self.lease.lease.oracle.owned_by(&info.sender) {
-            return Err(ContractError::Unauthorized(
-                access_control::error::Error::Unauthorized {},
-            ));
-        }
+        access_control::check(&self.lease.lease.oracle, &info.sender)?;
```
KirilMihaylov, line 78: I might be mistaken here but shouldn't this one be done on entering `crate::contract::state::State`'s method for handling alarms? At least that's how interpreted (4.) from the issue, #70.

cc @gmanev7

gmanev7: That struct defines all Lease states but lacks their details, such as the address of the time alarm contract. The DEX-related states are defined in the dex package and rely on the `dex::SwapTask` to define the swap task being performed.
Hopefully, one of the task attributes is `TimeAlarmRef`.

KirilMihaylov: If the solution would be to add it to each state, that is really error-prone. What if we implement the same trait, or perhaps a different one as the states aren't access resources themselves, on the states that would do the check and call that trait's method on the top-level?

platform/contracts/timealarms/src/stub.rs
```diff
@@ -66,6 +67,12 @@ impl TimeAlarmsRef {
     }
 }
 
+impl AccessPermission for TimeAlarmsRef {
+    fn permit_access(&self, caller: &Addr) -> bool {
+        self.owned_by(caller)
+    }
+}
+
```
gmanev7, line 75: The time alarm per se does not provide any permission, so this implеmentation is inappropriate.

A permission should be specific for the accessed resource.

platform/packages/access-control/src/lib.rs
```diff
@@ -11,14 +11,28 @@ use self::error::{Error, Result};
...
 pub mod error;
 
-pub fn check(permitted_to: &Addr, accessed_by: &Addr) -> Result {
-    if permitted_to == accessed_by {
+pub trait AccessPermission {
+    fn permit_access(&self, caller: &Addr) -> bool;
+}
+
+/// Checks if access is permitted for the given caller.
+pub fn check<R>(resource: &R, caller: &Addr) -> Result
```
gmanev7, line 19: please rename `resource` and the type parameter

platform/packages/access-control/src/lib.rs
```diff
@@ -11,14 +11,28 @@ use self::error::{Error, Result};
...
         Err(Error::Unauthorized {})
     }
 }
 
+impl AccessPermission for Addr {
+    fn permit_access(&self, caller: &Addr) -> bool {
+        self == caller
+    }
+}
+
```
gmanev7, line 35: similarly to my previous comment about TimeAlarmRef, this is inappropriate

platform/packages/oracle/src/stub/mod.rs
```diff
@@ -143,3 +144,13 @@ where
...
+
+impl<QuoteC, QuoteG> AccessPermission for OracleRef<QuoteC, QuoteG>
+where
+    QuoteC: Currency + MemberOf<QuoteG>,
+    QuoteG: Group,
+{
+    fn permit_access(&self, caller: &Addr) -> bool {
+        self.addr == caller
+    }
+}
```
gmanev7, line 156: same as above

protocol/contracts/lease/src/contract/state/opened/active.rs
gmanev7: In this particular case, the application permission one has to be granted is `TimeAlarmDelivery`. So we need a struct that holds what is necessary for the permission to check if it is granted to a caller.

protocol/contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -91,13 +86,7 @@ impl Active {
...
         info: MessageInfo,
     ) -> ContractResult<Response> {
-        // TODO define a trait 'RestrictedResource' with 'fn owner(&Addr) -> bool'
-        // and move this check to the 'access_control' package
-        if !self.lease.lease.time_alarms.owned_by(&info.sender) {
-            return Err(ContractError::Unauthorized(
-                access_control::error::Error::Unauthorized {},
-            ));
-        }
+        access_control::check(&self.lease.lease.time_alarms, &info.sender)?;
```
gmanev7, line 89: In this particular case, the application permission one has to be granted is `TimeAlarmDelivery`. So we need a struct that holds what is necessary for the permission to check if it is granted to a caller.

protocol/packages/dex/src/impl_/transfer_in_finish.rs
```diff
@@ -156,6 +156,11 @@ where
     }
 
     fn on_time_alarm(self, querier: QuerierWrapper<'_>, env: Env) -> HandlerResult<Self> {
+       let res = access_control::check(self.spec.time_alarm(), &env.contract.address);
```
gmanev7, line 159: checking the permission here would be applied on the `heal` operation which is not desired

protocol/packages/dex/src/impl_/transfer_in_finish.rs
```diff
@@ -156,6 +156,11 @@ where
...
+        if res.is_err() {
+            // TODO ?
+        }
+
```
gmanev7, line 163: we have `impl From<Error> for HandlerResult<H>`, so you may want to just use `?`

protocol/packages/dex/src/impl_/transfer_in_finish.rs
Lockwarr: should I change the heal func to just directly call try_complete()? Or are you suggesting to move the check somewhere else

platform/contracts/timealarms/src/stub.rs
Lockwarr: What do you refer to by the accessed resource then? I thought the accessed resource could be timeAlarmsRef, oracleRef or even Addr

protocol/packages/dex/src/impl_/transfer_in_finish.rs
Lockwarr: I tried this again in my latest commit, but just `?` is not accepted. Should I handle the result manually here - if yes - what is the expected outcome on error?

protocol/packages/dex/src/impl_/transfer_in_finish.rs
gmanev7: calling `try_complete` would be fine

protocol/packages/dex/src/impl_/transfer_in_finish.rs
gmanev7: > manually

not sure what you refer to

The only thing to do is to convert from the `access_control::Error` to `dex::Error`. Obviously, we need another variant. 
Finally, we would end up with a code chunk like this: `access_control::check(...).map_or_else(|err|DexError::Unauthorized(err).into(), |()| self.try_complete(querier, env))`

Lockwarr: thanks, I didn't realize it's that straight forward

protocol/contracts/lease/src/contract/endpoins.rs
```diff
@@ -146,11 +147,11 @@ fn process_execute(
         ExecuteMsg::TimeAlarm {} => state.on_time_alarm(querier, env, info),
         ExecuteMsg::PriceAlarm() => state.on_price_alarm(querier, env, info),
         ExecuteMsg::DexCallback() => {
-            access_control::check(&info.sender, &env.contract.address)?;
+            access_control::check(&GrantedAddress::new(&info.sender), &env.contract.address)?;
```
gmanev7, line 150: similarly to the TimeAlarmsRef's case, here we need a `DexResponseSafeDelivery` permission defined close to its usage

so again, the abstractions we need are *permissions*

platform/packages/access-control/src/lib.rs
```diff
@@ -11,14 +11,38 @@ use self::error::{Error, Result};
...
+        Self { addr }
+    }
+}
+
+impl AccessPermission for GrantedAddress<'_> {
+    fn is_granted_to(&self, caller: &Addr) -> bool {
+        self.addr == caller
+    }
+}
+
```
gmanev7, line 45: not needed, see below

platform/packages/access-control/src/lib.rs
```diff
@@ -11,14 +11,38 @@ use self::error::{Error, Result};
...
 pub struct SingleUserAccess<'storage, S>
```
gmanev7, line 46: I have reviewed the usage of this struct as it interferes with the permissions model we've introduced with this PR. All occurrences may be substituted with permissions. In some places, we need to enhance the stored data to contain the necessary pieces of information, for example, `lease_code_admin` to be included in Lpp config.

Although it's relative to the current task,  it deserves a separate PR.

platform/packages/oracle/src/stub/mod.rs
```diff
@@ -143,3 +144,31 @@ where
         }
     }
 }
+
+pub struct GrantedOracle<'a, QuoteC, QuoteG>
```
gmanev7, line 148: similarly to the time_alarms_ref, we need a `PriceAlarmDelivery` permission structure defined close to its usage

protocol/contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -162,27 +165,30 @@ impl Handler for Active {
...
-                            time_alarms,
-                            &oracle_ref,
-                            reserve,
-                        ),
-                        querier,
-                    )
-                    .map(|(lease, batch)| Response::from(batch, Self::new(lease)))
-            })
+        access_control::check(
+            &GrantedAddress::new(&self.lease.lease.customer),
```
gmanev7, line 169: GrantedAddress -> `ChangeClosePolicyPermission`

protocol/contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -192,9 +198,12 @@ impl Handler for Active {
         env: Env,
         info: MessageInfo,
     ) -> ContractResult<Response> {
-        access_control::check(&self.lease.lease.customer, &info.sender)
-            .map_err(Into::into)
-            .and_then(|()| customer_close::start(spec, self.lease, &env, querier))
+        access_control::check(
+            &GrantedAddress::new(&self.lease.lease.customer),
```
gmanev7, line 202: GrantedAddress -> `ClosePositionPermission`

protocol/contracts/lease/src/contract/state/opened/close/anomaly.rs
```diff
@@ -55,7 +55,7 @@ impl Handler for SlippageAnomaly {
     ) -> ContractResult<Response> {
         self.lease
             .leases
-            .check_assess(info.sender, querier)
+            .check_access(info.sender, querier)
```
gmanev7, line 58: -> `AnomalyResolution` that `impl Permission`

protocol/contracts/lease/src/contract/state/paid/mod.rs
```diff
@@ -39,7 +40,10 @@ impl Handler for Active {
         env: Env,
         info: MessageInfo,
     ) -> ContractResult<Response> {
-        access_control::check(&self.lease.lease.customer, &info.sender)?;
+        access_control::check(
+            &GrantedAddress::new(&self.lease.lease.customer),
```
gmanev7, line 44: GrantedAddress -> `CloseLease` that `impl Permission`

protocol/contracts/lease/src/contract/state/opened/close/anomaly.rs
Lockwarr: Not sure how exactly to refactor here as we need to execute the AccessCheck query in order to get the lease_admin and do the check
Maybe with the refactoring of SingleUserAccess, when we include lease_admin in the config, we can also refactor here

platform/packages/access-control/src/lib.rs
```diff
@@ -11,14 +11,38 @@ use self::error::{Error, Result};
...
+    P: AccessPermission,
+{
+    if permission.is_granted_to(caller) {
         Ok(())
     } else {
         Err(Error::Unauthorized {})
     }
 }
 
+pub struct SingleUserPermission<'a> {
```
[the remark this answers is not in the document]
Lockwarr: Regarding 1. - my plan was for the SingleUserPermission to replace SingleUserAccess which is also in this module, but it will be done in the following PR.

I agree with 2.

protocol/packages/dex/src/impl_/transfer_in_finish.rs
```diff
@@ -230,11 +232,18 @@ where
...
-        self.on_time_alarm(querier, env)
+        self.try_complete(querier, env)
     }
 
     fn on_time_alarm(self, querier: QuerierWrapper<'_>, env: Env) -> HandlerResult<Self> {
-        self.try_complete(querier, env)
+        access_control::check(
+            &TimeAlarmDelivery::new(self.spec.time_alarm()),
+            &env.contract.address,
+        )
```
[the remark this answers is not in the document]
Lockwarr: Is the suggestion to change AccessPermission's parameter to MessageInfo so it's used everywhere? It makes sense

However, here in the TransferInFinish, I don't see any way to get the caller

platform/packages/access-control/src/lib.rs
```diff
@@ -11,14 +11,28 @@ use self::error::{Error, Result};
 mod contract_owner;
 pub mod error;
 
-pub fn check(permitted_to: &Addr, accessed_by: &Addr) -> Result {
-    if permitted_to == accessed_by {
+pub trait AccessPermission {
+    fn permit_access(&self, caller: &Addr) -> bool;
```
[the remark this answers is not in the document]
gmanev7: the suggested variant has a prefix `is` that is discouraged for use except in cases to validate the target is a specific instance. Here we apply the common recommendation - without "is" unless it adds value

platform/contracts/timealarms/src/stub.rs
gmanev7: the accessed resource is for example the Lease or the Profit

protocol/contracts/lease/src/contract/state/opened/close/anomaly.rs
gmanev7: We are not going to include `lease_admin` in this contract's storage.
What I asked with this thread was to "wrap" the check in a type that `impl Permission` and take benefit of the new `fn access_control::check`.

protocol/contracts/lease/src/contract/state/opened/active.rs
gmanev7: I'll suggest resolving this in the next PR. See more in a comment at this PR below

protocol/contracts/leaser/src/contract.rs
```diff
@@ -164,10 +181,10 @@ pub fn execute(
             }),
         ExecuteMsg::ChangeLeaseAdmin { new } => Leaser::new(deps.as_ref())
             .config()
-            .and_then(|ref config| {
-                ChangeLeaseAdminPermission::from(config).check_access(&info.sender)
+            .and_then(|config| {
+                access_control::check(&ChangeLeaseAdminPermission::new(&config), &info)?;
+                validate(&new, deps.api)
             })
```
KirilMihaylov, line 187:
```suggestion
            .and_then(|config| {
                access_control::check(&ChangeLeaseAdminPermission::new(&config), &info)
            })
            .and_then(|()| validate(&new, deps.api))
```

protocol/contracts/leaser/src/contract.rs
```diff
@@ -197,17 +214,21 @@ pub fn sudo(deps: DepsMut<'_>, _env: Env, msg: SudoMsg) -> ContractResult<Respon
...
-                if granted {
-                    AccessGranted::Yes
-                } else {
-                    AccessGranted::No
-                }
-            })
-            .and_then(serialize_to_json),
+        QueryMsg::CheckAnomalyResolutionPermission { by: caller } => {
+            let msg_info = MessageInfo {
+                sender: caller.clone(),
```
KirilMihaylov, line 219: Redundant `Clone::clone` call.

protocol/contracts/leaser/src/contract.rs
```diff
@@ -197,17 +214,21 @@ pub fn sudo(deps: DepsMut<'_>, _env: Env, msg: SudoMsg) -> ContractResult<Respon
...
+                funds: vec![],
+            };
+
+            Leaser::new(deps)
+                .config()
+                .and_then(|config| {
+                    access_control::check(&AnomalyResolutionPermission::new(&config), &msg_info)
+                        .map(|_| AccessGranted::Yes)
+                        .or_else(|_| Ok(AccessGranted::No))
+                })
```
KirilMihaylov, line 229:
```suggestion
                .map(|config| {
                    if access_control::check(&AnomalyResolutionPermission::new(&config), &msg_info).is_ok() {
                        AccessGranted::Yes
                    } else {
                        AccessGranted::No
                    }
                })
```

protocol/packages/dex/src/impl_/transfer_in_finish.rs
```diff
@@ -1,6 +1,7 @@
 use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::marker::PhantomData;
```
KirilMihaylov, line 2: [Code style]
```suggestion
use std::{
    fmt::{Display, Formatter, Result as FmtResult},
    marker::PhantomData,
};
```

protocol/packages/dex/src/impl_/transfer_in_finish.rs
```diff
@@ -1,6 +1,7 @@
...
 
+use crate::Error as DexError;
```
KirilMihaylov, line 4: [Code style] As per my last comment, imports should be grouped and ordered.

KirilMihaylov (changes requested):

protocol/contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -162,41 +169,50 @@ impl Handler for Active {
...
-                })
+        access_control::check(
+            &ChangeClosePolicyPermission::new(&self.lease.lease.customer),
+            &info,
+        )
+        .map_err(Into::into)
+        .and_then(|()| {
+            let profit = self.lease.lease.loan.profit().clone();
+            let time_alarms = self.lease.lease.time_alarms.clone();
+            let oracle_ref = self.lease.lease.oracle.clone();
```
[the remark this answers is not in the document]
Lockwarr: for me it's not obvious that `oracle_ref` is a reference, the new() func in `contracts/lease/src/lease/dto.rs` doesn't use `&` for the oracle.
I will leave this as it is, let me know if I'm mistaken @KirilMihaylov

protocol/contracts/lease/src/contract/state/opened/active.rs
```diff
@@ -162,41 +169,50 @@ impl Handler for Active {
...
+            let reserve = self.lease.lease.reserve.clone();
+            self.lease.update(
+                ChangeClosePolicy::new(
+                    change,
+                    &env.block.time,
```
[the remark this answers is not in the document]
gmanev7: This project passes small Rust structs by-ref.
I have done a thorough research on the Rust idioms and have not found any strong opinion pro- or against.
Finally, I have decided to set `by-ref` as a code policy. Rationale: If not inlined or optimized away, the compiler would need to send the same size for references, 4 bytes on WASM, despite the size of the structure they point to.
Basically, `Timestamp` is a `u64`, i.e., 8 bytes, so passing by ref seems a natural choice.

>

gmanev7:
Switching from `sender: &Addr` to `info: &MessageInfo` eliminated a class of errors but introduced unnecessary dependencies on a `cosmwasm-std` type deep in our design.
So I would suggest defining our `struct Sender<'info>(&'info MessageInfo)`  and `impl AsRef<Addr>` to facilitate checks.

platform/packages/access-control/src/lib.rs
```diff
@@ -38,11 +48,13 @@ where
         }
     }
 
-    pub fn check(&self, user: &Addr) -> Result {
+    pub fn check(&self, info: &MessageInfo) -> Result {
         self.storage_item
             .load(self.storage.deref())
             .map_err(Into::into)
-            .and_then(|granted_to| check(&granted_to, user))
+            .and_then(|granted_to| {
```
gmanev7, line 55: bind `granded_to` by ref

protocol/contracts/leaser/src/contract.rs
```diff
@@ -42,6 +39,26 @@ const CURRENT_RELEASE: ProtocolPackageRelease = ProtocolPackageRelease::current(
...
+impl AccessPermission for LeaseAdminOnly<'_> {
+    fn granted_to(&self, info: &MessageInfo) -> bool {
+        self.lease_config.lease_admin == info.sender
+    }
+}
+
+type LeasesConfigurationPermission<'a> = LeaseAdminOnly<'a>;
+type ChangeLeaseAdminPermission<'a> = LeaseAdminOnly<'a>;
+type AnomalyResolutionPermission<'a> = LeaseAdminOnly<'a>;
+
```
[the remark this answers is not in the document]
Lockwarr: Maybe the authz module that you remember was the one in `lease/src/api`. 
The agreed module naming is permissions, I will organize those struct and types to be also used from a separate module `permissions` here in the leaser

protocol/contracts/leaser/src/contract.rs
```diff
@@ -189,9 +197,13 @@ pub fn query(deps: Deps<'_>, _env: Env, msg: QueryMsg) -> ContractResult<Binary>
...
         QueryMsg::CheckAnomalyResolutionPermission { by: caller } => Leaser::new(deps)
             .config()
-            .map(|ref config| AnomalyResolutionPermission::from(config).granted_to(&caller))
-            .map(|granted| {
-                if granted {
+            .map(|ref config| {
+                if access_control::check(
+                    &AnomalyResolutionPermission::new(config),
+                    &Sender::from_addr(&caller),
+                )
```
gmanev7, line 204: It is not appropriate to call `check` since we do not do AuthZ check - this is just a query. Even if we wanted we could not do that because queries do not authenticate callers.

gmanev7: The previous code was accurate. Please revert.

platform/packages/access-control/src/lib.rs
```diff
@@ -1,18 +1,42 @@
...
 use self::error::{Error, Result};
+use self::permissions::SingleUserPermission;
 
 mod contract_owner;
 pub mod error;
+pub mod permissions;
 
-pub fn check(permitted_to: &Addr, accessed_by: &Addr) -> Result {
-    if permitted_to == accessed_by {
+pub struct Sender<'a> {
```
gmanev7, line 16: nice - a couple of remarks:
1. Again, please do not clutter the common namespace - this struct deserves its module.
2. The [issue]( https://github.com/nolus-protocol/nolus-money-market/pull/519#issuecomment-3044602172 ) was not resolved fully. Its second part is not implemented - the `AsRef` impl. Why did we need it? To work with abstractions, not specific types. Where?  The `fn access_control::check` and `fn AccessPermission::granted_to`. Another improvement would be a `trait Sender where Self: AsRef<Addr>{}` to make sure plain `&Addr`-s cannot be used to call the `check` function.
4. Would recommend renaming `fn Sender::new` to `fn Sender::of_execute_msg`.

gmanev7 (commented):
A few new, or unfinished previous,issues left. Hurry up, there are a a couple of other big change lists competing to merge onto main.

protocol/contracts/profit/src/contract.rs
```diff
@@ -104,13 +106,13 @@ pub fn execute(
                 deps.storage.deref(),
                 crate::access_control::TIMEALARMS_NAMESPACE,
             )
-            .check(&info.sender)?;
+            .check(&info)?;
 
-            try_handle_execute_message(deps, env, State::on_time_alarm)
+            try_handle_execute_message(deps, env, info, State::on_time_alarm)
                 .map(response::response_only_messages)
```
gmanev7, line 112: it is more complex to change the handler spec, the `try_handle_execute_message` fn, and 3 invocations, than to provide a closure binding `info` inside

protocol/contracts/profit/src/contract.rs
```diff
@@ -188,10 +203,11 @@ fn try_handle_neutron_msg(
 fn try_handle_execute_message<F, R, E>(
     deps: DepsMut<'_>,
     env: Env,
+    info: MessageInfo,
```
gmanev7, line 206: please revert, see above

gmanev7 (commented):
With this set of comments, I have completed the review round

PR #525, opened 2025-03-18
refactor(protocol/leaser): House-keeping and minor improvements.

KirilMihaylov:


protocol/contracts/leaser/src/migrate.rs
```diff
@@ -5,7 +5,7 @@ use versioning::{ProtocolMigrationMessage, ProtocolPackageRelease};
 
 use crate::{lease::Release as LeaseReleaseTrait, msg::MaxLeases, result::ContractResult};
 
-pub struct Customer<Leases> {
+pub(crate) struct Customer<Leases> {
```
gmanev7: [the entire commit] I did not do that intentionally because it adds a lot of noise.
I reason here in this way: the module exports the main function `migrate_leases`, so the related types are. The parent module is responsible for deciding whether they should be visible upward, not the module itself.

gmanev7: and hence, purposefully, the definition of the module is private within the root crate

protocol/contracts/leaser/src/state/leases.rs
```diff
@@ -88,6 +88,22 @@ impl Leases {
     }
 }
 
+pub(crate) trait CustomerWithLeasesIterator
```
gmanev7, line 91: - rename to `CustomerLeases`
- move up as the first definition in this module
- avoid the backward dependencies to `mod migrate` by bringing `Customer` and MaybeCustomer` here

protocol/contracts/leaser/src/leaser.rs
```diff
@@ -171,6 +168,10 @@ where
     })
 }
 
+fn finalize_batch_migration(result: MigrationResult) -> MessageResponse {
```
gmanev7, line 171: rename to `build_response`

protocol/contracts/leaser/src/customer.rs
gmanev7, line 1: a step in the right direction ... but as I have already expressed my concern, this keeps a dependency in the wrong direction.
Assuming that the `mod state` is the core of the leaser domain model then this new module should be part of it. The goal is to make the domain model self-sufficient (except for the types borrowed for use in the lease contract API).

PR #529, opened 2025-04-08
Introduce BoundPercent and improve safety in arithmetic operations

maneva3:
Done in this PR:
    Introduced a new BoundPercent type.

    Separated and generalized logic by spreading Percent and Percent100 usage.

    Removed the Add and Sub trait implementations for BoundPercent and replaced their 
    usage with safer alternatives using checked_ methods to avoid panics and ensure overflow handling.

Still TODO (in this PR):
    Reorganize some of the modules to improve readability and overall usability.

    Properly propagate potential overflows in the protocol module to ensure consistent error handling.

Planned for Future PRs:

    Define a new trait Scalar to enable more flexible and unified comparison logic for Rational types.

    Refactor Ratio and Rational to be combined into a more coherent abstraction, similar to how BoundPercent encapsulates Percent.

    Eliminate duplication in arithmetic operations shared between Percent and Rational.


platform/packages/finance/src/ratio.rs
```diff
@@ -1,25 +1,58 @@
...
+        Self { parts, total }
+    }
+
+    pub(crate) fn parts(&self) -> U {
+        self.parts
+    }
+
+    pub(crate) fn total(&self) -> U {
+        self.total
+    }
```
gmanev7, line 35: please double check if we rfeally need such accessors

platform/packages/finance/src/ratio.rs
```diff
@@ -30,30 +63,160 @@ where
...
     }
+
+    pub fn nominator(&self) -> U {
+        self.nominator
+    }
+
+    pub fn denominator(&self) -> U {
+        self.denominator
+    }
+
```
gmanev7, line 74: please double check if we rfeally need such accessors

platform/packages/finance/src/ratio.rs
```diff
@@ -1,25 +1,58 @@
...
+    fn of<A>(&self, whole: A) -> A
+    where
+        A: Fractionable<U>,
+    {
+        whole.safe_mul(self)
+    }
 }
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
-#[serde(rename_all = "snake_case")]
```
gmanev7, line 14: let keep this eventhough the properties may not need such a renaming

platform/packages/finance/src/ratio.rs
```diff
@@ -1,25 +1,58 @@
...
+#[derive(Clone, Copy, Debug, Deserialize, Eq, Serialize)]
```
gmanev7, line 47: we cannot rely on the derived `Eq`, because it compares the objects literally, which would conclude that 1/2 != 2/4. We should guarantee a canonical representation to benefit from the derived `Eq`. By the way, why do we need `Eq`?

platform/packages/finance/src/ratio.rs
```diff
@@ -1,25 +1,58 @@
...
 use serde::{Deserialize, Serialize};
 
 use crate::{fraction::Fraction, fractionable::Fractionable, zero::Zero};
 
-// TODO review whether it may gets simpler if extend Fraction
-pub trait Ratio<U> {
-    fn parts(&self) -> U;
-    fn total(&self) -> U;
+#[cfg_attr(any(test, feature = "testing"), derive(Debug))]
+pub struct Ratio<U> {
```
gmanev7, line 14: please add at least a line long rustdoc

platform/packages/finance/src/ratio.rs
```diff
@@ -1,25 +1,58 @@
...
+    parts: U,
+    total: U,
+}
```
gmanev7, line 17: The `Ratio` instances seem to be a subset of the `Rational` instances. Why don't we compose the latter into the definition of the former? That would express such a relation to a high extent.

gmanev7 (commented):
just a first few insights, more to follow

platform/packages/finance/src/coin/mod.rs
```diff
@@ -22,6 +30,64 @@ mod dto;
...
+{
+    type Output = BoundPercent<B>;
+
+    fn checked_mul(self, rhs: BoundPercent<B>) -> Option<Self::Output> {
+        checked_mul_and_convert(self, rhs, |result| {
+            result.try_into().ok().map(BoundPercent::from_permille)
+        })
+    }
+}
+
```
gmanev7, line 80: Please double-check that we need these implementations.
They are meaningless from a business perspective and dangerous from a technical perspective.

platform/packages/finance/src/duration.rs
```diff
@@ -84,27 +94,38 @@ impl Duration {
...
     pub fn checked_mul(&self, rhs: u16) -> Option<Self> {
         self.nanos().checked_mul(rhs.into()).map(Self::from_nanos)
     }
 
     #[track_caller]
-    pub fn annualized_slice_of<T>(&self, annual_amount: T) -> T
+    pub fn annualized_slice_of<T>(&self, annual_amount: T) -> Option<T>
     where
-        T: TimeSliceable,
+        Self: Copy + Div + Rem<Output = Self> + Zero,
```
gmanev7, line 108: Have you tried to remove these boundaries? It seems to me that the compiler already knows what does `Self` implement

platform/packages/finance/src/duration.rs
```diff
@@ -117,6 +138,12 @@ impl From<Duration> for Uint128 {
     }
 }
 
+impl From<Duration> for Uint256 {
+    fn from(d: Duration) -> Self {
+        Uint256::from(d.nanos())
```
gmanev7, line 143: Please leverage the previous conversion.
Do not write this from scratch!

platform/packages/finance/src/interest.rs
```diff
@@ -1,17 +1,24 @@
...
+    ratio::{CheckedAdd, CheckedMul},
     zero::Zero,
 };
 
 /// Computes how much interest is accrued
-pub fn interest<U, F, P>(rate: F, principal: P, period: Duration) -> P
+pub fn interest<U, F, P>(rate: F, principal: P, period: Duration) -> Option<P>
 where
     F: Fraction<U>,
-    P: Fractionable<U> + TimeSliceable,
```
gmanev7, line 14: please remove the `TimeSliceable` definition if it is no more in use

platform/packages/finance/src/interest.rs
```diff
@@ -20,40 +27,53 @@ where
 /// Computes how much time this payment covers, return.0, and the change, return.1
 ///
 /// The actual payment is equal to the payment minus the returned change.
-pub fn pay<U, F, P>(rate: F, principal: P, payment: P, period: Duration) -> (Duration, P)
+pub fn pay<P>(rate: Percent100, principal: P, payment: P, period: Duration) -> Option<(Duration, P)>
```
gmanev7, line 30: Please make sure the `rate` argument uses the same type as in `fn interest`, either generic or concrete

platform/packages/finance/src/interest.rs
```diff
@@ -1,17 +1,24 @@
...
+    fmt::Debug,
+    ops::{Div, Rem, Sub},
+};
 
 use crate::{
     duration::Duration,
     fraction::Fraction,
-    fractionable::{Fractionable, TimeSliceable},
+    fractionable::Fractionable,
+    percent::{Percent100, Units},
```
gmanev7, line 11: A suggestion: provide a more specific name to a type that is taken out of its context.
Here, `Units` is imported, and the reader would have to dig it out what particular units it represent

PR #541, opened 2025-07-05
refactor(ci): Refactor CI to make workflows reproducible and rely on binary caches.

KirilMihaylov:
As a result of these changes, the current breakage will be avoided as the need to query versions is subverted.
Due to the delegation of the responsibilities to maintain the environment to Nix, this allows for the use of binary caches provided by the Nix/-OS project, and ease of caching both the provided binaries and what is built from local packages.

The notable changes to the workflows are as follows:
* Parallelisation on top of separation by workspaces and DEX type is removed. What is left is separation by profiles as those steps cannot share binary outputs.
  
  _Observation:_ While it does add more compilation time to individual steps, this change allows running workflows on different branches/tags in parallel.
* Building WebAssembly binaries and running all the checks on them is run in parallel with the rest of the "checks" jobs as they don't have a dependency between each other. The job for creation of a draft release is left to depend on all "checks" jobs being successful.
  
  _Observation:_ This change compensates for the added time by the other change.

_Overall observation:_ Workflow times are on par, if not slightly lower, compared to successful completion times before breakage occurred. The existence of fresh Nix caches stored in GitHub Actions' cache system sheds around a minute.

_Additional notes:_ This PR _should not_ impact the current state of the shell scripts within the `scripts` directory. While the new recommended way, and thus source of truth because of it's reproducibility, is to use the Nix derivations, they are not expected to misbehave.


tools/cargo-each/src/subcommands/run.rs
```diff
@@ -171,8 +174,51 @@ fn execute_command_builder(
...
             let build_base = move || match command_type {
-                CommandType::CargoSubcommand { ref cargo_path } => Command::new(cargo_path)
-                    .pipe_mut(|command| _ = command.arg(subcommand.as_os_str())),
+                CommandType::CargoSubcommand { ref rust_path } => {
+                    let mut command = Command::new({
+                        static CARGO_PATH: OnceLock<Cow<'static, OsStr>> = OnceLock::new();
+
+                        CARGO_PATH.get_or_init(|| {
+                            if let Some(rust_path) = rust_path {
+                                Cow::Owned(rust_path.join("cargo").into_os_string())
```
gmanev7, line 183: I've found `join("cargo"` two times in this PR. Please avoid such repetitions

gmanev7:
In general, let's try to evaluate the slowdown of the build if we do not use cached binaries. Have you measured it?

PR #547, opened 2025-07-21
Multiplication overflow preparation

bmanev18:
Pending test refactoring.


platform/packages/finance/src/coin/mod.rs
```diff
@@ -86,6 +163,10 @@ impl<C> Coin<C> {
         })
     }
 
+    pub fn to_rational<OtherC>(self, denominator: Coin<OtherC>) -> Rational<Amount> {
+        Rational::new(self.amount, denominator.amount)
+    }
+
```
gmanev7, line 169: I am cautious about adding extra dependencies if we may live without them. When they enter the public API I am even more resistant. Please re-think this function

PR #548, opened 2025-07-31
Introduce Percent100 as bounded Percent

maneva3:


platform/packages/finance/src/percent/bound.rs
```diff
@@ -1,57 +1,240 @@
...
+        Self::HUNDRED.units()
+    }
+}
+
+impl<const UPPER: Units> FractionUnit for BoundPercent<UPPER> where
+    BoundPercent<UPPER>: Copy + Debug + Ord + Zero
+{
+}
+
+impl<const UPPER_BOUND: Units> From<BoundPercent<UPPER_BOUND>> for Amount {
```
gmanev7, line 139: Since it is not obvious, could you document the need for this conversion?

platform/packages/finance/src/percent/bound.rs
```diff
@@ -1,57 +1,240 @@
...
+    fn from(percent: BoundPercent<UPPER_BOUND>) -> Self {
+        Amount::from(percent.units())
+    }
+}
+
+impl<const UPPER_BOUND: Units> From<BoundPercent<UPPER_BOUND>> for SimpleFraction<Amount> {
```
gmanev7, line 145: Since it is not obvious, could you document the need for this conversion?

platform/packages/finance/src/percent/bound.rs
```diff
@@ -1,57 +1,244 @@
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult, Write};
+
+#[cfg(any(test, feature = "testing"))]
+use std::ops::{Add, Sub};
+
+use sdk::cosmwasm_std::Uint256;
```
gmanev7, line 6: `sdk` is not a 3rd party crate, it's Nolus Money Market one, so it should reside in the next set of `use`-s

platform/packages/finance/src/percent/bound.rs
```diff
@@ -1,57 +1,244 @@
...
+    pub const ZERO: Self = Self::try_from_primitive(0).expect("0% is a valid instance");
+    pub const HUNDRED: Self = Self::try_from_primitive(100).expect("100% is a valid instance");
+    pub(crate) const PRECISION: Self =
+        Self::try_from_permille(1).expect("0.1% is a valid instance");
+
+    const UNITS_TO_PERCENT_RATIO: Units = 10;
+
+    #[cfg(any(test, feature = "testing"))]
+    pub const fn from_percent(percent: u16) -> Self {
+        Self::try_from_primitive(percent as u32)
```
gmanev7, line 35: `as` is forbidden for use in our project since it may change the value without notice. We have discussed the shortcomings of `as`. Although the conversion is lossless here, we do not want to become accustomed to it in the code.

platform/packages/finance/src/percent/bound.rs
```diff
@@ -1,57 +1,232 @@
...
-    type Error = Error;
+#[cfg(any(test, feature = "testing"))]
+impl<const UPPER_BOUND: Units> Add for BoundPercent<UPPER_BOUND> {
+    type Output = Self;
 
-    fn try_from(value: Percent) -> Result<Self> {
-        Self::try_from_percent(value)
+    #[track_caller]
+    fn add(self, rhs: Self) -> Self {
+        Self::checked_add(self, rhs).expect("attempt to add with overflow")
```
gmanev7, line 171:
```suggestion
        self.checked_add(rhs).expect("attempt to add with overflow")
```

platform/packages/finance/src/percent/bound.rs
```diff
@@ -1,57 +1,232 @@
...
-impl<const UPPER_BOUND: Units> From<BoundPercent<UPPER_BOUND>> for Percent {
-    fn from(value: BoundPercent<UPPER_BOUND>) -> Self {
-        value.percent()
+#[cfg(any(test, feature = "testing"))]
+impl<const UPPER_BOUND: Units> Sub for BoundPercent<UPPER_BOUND> {
+    type Output = Self;
+
+    #[track_caller]
+    fn sub(self, rhs: Self) -> Self {
+        Self::checked_sub(self, rhs).expect("attempt to subtract with overflow")
```
gmanev7, line 181:
```suggestion
        self.checked_sub(rhs).expect("attempt to subtract with overflow")
```

platform/packages/finance/src/fractionable/coin.rs
```diff
@@ -26,32 +26,31 @@ mod test {
...
-            Fractionable::<u32>::safe_mul(
-                Coin::<SuperGroupTestC1>::new(2),
-                &Rational::new(u32::MAX, 1u32)
-            )
+            Coin::<SuperGroupTestC1>::new(Amount::from(u32::MAX - 1)),
+            SimpleFraction::new(u32::MAX - 1, u32::MAX)
+                .of(Coin::<SuperGroupTestC1>::new(Amount::from(u32::MAX)))
+                .unwrap()
         );
     }
```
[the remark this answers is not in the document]
maneva3: The tests were placed in the module covering `of()` because `safe_mul` was only used there. After reverting to call `safe_mu`l directly instead of through `of()`, the tests were moved back, closer to the module where `safe_mu`l is implemented.

platform/packages/finance/src/duration.rs
```diff
@@ -305,17 +339,54 @@ mod tests {
...
     #[test]
-    fn checked_mul() {
-        assert_eq!(Some(D::from_secs(10)), D::from_secs(5).checked_mul(2));
-        assert_eq!(Some(D::from_secs(0)), D::from_secs(5).checked_mul(0));
+    fn annualized_slice_of() {
+        let annual_amount = test_coin(1000);
+        assert_eq!(annual_amount, D::YEAR.annualized_slice_of(annual_amount));
+        assert_eq!(
+            annual_amount
+                .checked_mul((D::HOUR.0 / D::YEAR.0).into())
```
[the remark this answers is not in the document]
maneva3: Already fixed.

platform/packages/finance/src/fractionable/price.rs
```diff
@@ -73,19 +73,19 @@ mod test {
...
         use crate::fractionable::price::test::{c, q};
-        use crate::{percent::Percent, price};
+        use crate::{percent::Percent100, price};
 
         #[test]
         fn greater_than_one() {
             let price = price::total_of(c(1)).is(q(1000));
-            let permille = Percent::from_permille(1);
+            let permille = Percent100::from_permille(1);
             assert_eq!(permille.of(price), price::total_of(c(1)).is(q(1)));
```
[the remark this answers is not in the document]
maneva3: You’re right, but since this matches the main branch and this refactor isn’t part of this PR, I'd prefer to keep it as-is to minimize the diff. All `safe_mul` usages will be replaced with `of()` in the next PR.

platform/packages/finance/src/duration.rs
```diff
@@ -299,17 +331,58 @@ mod tests {
...
+        // SimpleFraction::of() calls its checked_mul method instead of safe_mul
+        _ = (D::YEAR + D::HOUR).annualized_slice_of(test_coin(Amount::MAX));
+    }
+
+    #[test]
+    fn into_slice_per_ratio() {
         assert_eq!(
-            None,
-            D::from_nanos(Units::MAX / Units::from(u16::MAX) + 1).checked_mul(u16::MAX)
+            D::from_nanos(D::YEAR.0 / 4),
```
gmanev7, line 360: Please don't access internal data. Here, you exploit the sub-module visibility.
Instead, adjust values to avoid it, for example, D::from_days(365/5) and D::YEAR.into_slice_per_ratio(test_coin(1), test_coin(5)).

platform/packages/finance/src/fractionable/percent.rs
```diff
@@ -35,29 +36,42 @@ impl<C> Fractionable<Coin<C>> for Percent {
...
+            assert_eq!(
+                Percent::from_permille(Units::MAX),
+                Percent::from_percent(100).safe_mul(&Percent::from_permille(Units::MAX))
+            );
 
             let p_units: Units = 410;
-            let p64: <u32 as HigherRank<u8>>::Type = p_units.into();
-            let p64_res: <u32 as HigherRank<u8>>::Type = p64 * u64::from(Units::MAX) / 1000;
+            let p64: u64 = p_units.into();
+            let p64_res = p64 * u64::from(Units::MAX) / 1000;
```
gmanev7, line 72: Please do not degrade the tests. Why did you specify a concrete type instead of its `HigherRank::Type`?

platform/packages/finance/src/fraction.rs
```diff
@@ -1,7 +1,16 @@
-use crate::fractionable::Fractionable;
+use std::fmt::Debug;
 
+use crate::{fractionable::Fractionable, zero::Zero};
+
+/// A fraction is <= 1 (100%) that applied to a `whole` returns a part of it.
```
gmanev7, line 5: /// A part of a `whole`
///
/// Never greater than the `whole`

platform/packages/finance/src/fraction.rs
```diff
@@ -1,7 +1,16 @@
...
 pub trait Fraction<U> {
     fn of<A>(&self, whole: A) -> A
     where
         A: Fractionable<U>;
 }
+
+pub trait Unit
+where
+    Self: Copy + Debug + Ord + Zero,
```
gmanev7, line 14: I believe `weak ordering` is enough, which means using `PartialOrd` instead of `Ord`.

protocol/contracts/lpp/src/borrow.rs
```diff
@@ -43,42 +43,51 @@ impl InterestRate {
...
 
-    pub fn calculate<Lpn>(&self, total_liability: Coin<Lpn>, balance: Coin<Lpn>) -> Percent {
-        let utilization_max = Percent::from_ratio(
+    pub fn calculate<Lpn>(&self, total_liability: Coin<Lpn>, balance: Coin<Lpn>) -> Percent100 {
+        let utilization_max = Percent::from_fraction(
             self.utilization_optimal.units(),
-            (Percent::HUNDRED - self.utilization_optimal).units(),
-        );
+            Percent100::HUNDRED
+                .checked_sub(self.utilization_optimal)
```
gmanev7, line 62: Suggest using `Percent100::complement`

protocol/contracts/lpp/src/borrow.rs
```diff
@@ -43,42 +43,51 @@ impl InterestRate {
...
-
-        self.base_interest_rate + Fraction::<Units>::of(&config, utilization)
+        Rational::<PercentUnits>::of(
+            &SimpleFraction::new(self.addon_optimal_interest_rate, self.utilization_optimal),
+            utilization,
+        )
+        .and_then(|utilization_config| {
+            utilization_config
+                .checked_add(self.base_interest_rate.into())
+                .map(|res| Percent100::try_from(res).expect("The borrow rate must not exceed 100%"))
```
gmanev7, line 83: The error from `checked_add` goes up and panics with "The utilization_config must be a valid Percent" which is incorrect. Rather, have a `map` on `utilization_config` and expect the result of `checked_add *and* `Percent100::fry_from`

protocol/contracts/lpp/src/borrow.rs
```diff
@@ -198,22 +183,25 @@ mod tests {
...
+            );
+
+            res.expect("Rates must not exceed 100%!")
         }
 
-        fn ratio(n: Units, d: Units) -> Percent {
-            Percent::from_ratio(n, d)
+        fn ratio(n: Units, d: Units) -> Percent100 {
+            Percent100::from_fraction(n, d).expect(
+                "TODO replace with convertion from Ratio to Percent100 when Ratio become a struct",
```
gmanev7, line 197: Ratio *is* a struct, please resolve

protocol/contracts/lpp/src/lpp.rs
```diff
@@ -752,27 +746,32 @@ mod test {
 
         use super::{super::LiquidityPool, TheCurrency};
 
-        const FIFTY_PERCENT_MIN_UTILIZATION: fn() -> BoundToHundredPercent =
-            || Percent::from_permille(500).try_into().unwrap();
+        const FIFTY_PERCENT_MIN_UTILIZATION: fn() -> Percent100 = || Percent100::from_permille(500);
```
gmanev7, line 749: no need to be `fn()` anymore, make it `Percent100` only

protocol/packages/marketprice/src/feed/sample.rs
```diff
@@ -101,21 +101,21 @@ where
...
 
-    fn end_of_period(&mut self) {
+    fn end_of_period(&mut self) -> Option<()> {
         let prices_number = self.sample_prices.len();
         if prices_number > 0 {
             let mut values = self.sample_prices.values();
-            let first = values
-                .next()
-                .expect("should have been checked that there is at least one member");
+            let first = values.next()?;
```
gmanev7, line 108: Do not agree to translate the code in this way. It has already been checked that there are values.

protocol/packages/marketprice/src/feed/sample.rs
```diff
@@ -101,21 +101,21 @@ where
...
-            let avg = Fraction::<usize>::of(&part, sum);
-            self.last_sample = Sample { price: Some(avg) }
+            let denominator = u32::try_from(prices_number).ok()?;
+            let part = SimpleFraction::new(1, denominator);
+            let avg = Rational::<u32>::of(&part, sum);
+            self.last_sample = Sample { price: avg }
         }
         self.sample_prices.clear();
         self.sample_start = self.sample_end();
+        Some(())
```
gmanev7, line 118: The result is not checked, so why do we generate Option?!?

protocol/contracts/lpp/src/borrow.rs
```diff
@@ -43,42 +43,51 @@ impl InterestRate {
...
 
-    pub fn addon_optimal_interest_rate(&self) -> Percent {
+    pub const fn addon_optimal_interest_rate(&self) -> Percent100 {
         self.addon_optimal_interest_rate
     }
 
-    pub fn calculate<Lpn>(&self, total_liability: Coin<Lpn>, balance: Coin<Lpn>) -> Percent {
-        let utilization_max = Percent::from_ratio(
+    pub fn calculate<Lpn>(&self, total_liability: Coin<Lpn>, balance: Coin<Lpn>) -> Percent100 {
+        let utilization_max = Percent::from_fraction(
```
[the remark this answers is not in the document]
maneva3: The method cannot be const because it depends on non-const methods.

protocol/packages/marketprice/src/feed/sample.rs
```diff
@@ -103,17 +103,25 @@ where
...
-            let sum = values.fold(*first, |acc, current| acc + *current);
-            let part = Rational::new(1, prices_number);
-            let avg = Fraction::<usize>::of(&part, sum);
-            self.last_sample = Sample { price: Some(avg) }
+            let max_items = usize::try_from(u128::MAX).unwrap_or(usize::MAX);
+            let sum = values
+                .take(max_items - 1)
+                .fold(*first, |acc, current| acc + *current);
+
+            let denominator = u128::try_from(prices_number).unwrap_or(u128::MAX);
```
gmanev7, line 118: - The variable name is meaningless
- This creates a risk of diverging from the real number of elements that are summed. Please unify.

platform/packages/finance/src/fractionable/price.rs
```diff
@@ -17,6 +17,20 @@ where
...
+{
+    fn safe_mul<F>(self, fraction: &F) -> Self
+    where
+        F: RatioLegacy<Amount>,
+    {
+        self.lossy_mul(fraction)
+    }
+}
+
+// TODO remove when Fractionable is refactored
```
gmanev7, line 33: It is ready to be removed now - the `impl Fractionable<usize> for Price`

maneva3: Yes, you’re right. I’m aware of it, but since it’s not directly part of this PR, I preferred to keep the removal for the next PR where I’ll refactor the `Fractionable`

platform/packages/finance/src/coin/mod.rs
```diff
@@ -25,6 +25,8 @@ pub type Amount = u128;
 #[cfg(feature = "testing")]
 pub type NonZeroAmount = NonZeroU128;
 
+impl FractionUnit for Amount {}
+
```
gmanev7, line 29: a bit of misleading

What we need is to implement it for `u128`

tests/src/lpp_tests.rs
```diff
@@ -50,30 +51,37 @@ type LeaseCurrency = LeaseC1;
...
+    addon_rate: Percent100,
+    optimal_rate: Percent100,
+) -> Percent100 {
+    // TODO migrate to using SimpleFraction once it starts implementing Ord
+    Percent::from_fraction(loan, balance)
+    .map(|utilization_factor_max| {
+            // TODO migrate to using SimpleFraction once it starts implementing Ord
+            let utilization_factor = Percent::from_fraction(
+                    optimal_rate.units(),
+                    (Percent100::HUNDRED - optimal_rate).units(),
```
gmanev7, line 64: complement?

gmanev7 (approved):
Good job!

PR #549, opened 2025-08-05
Remove From<Amount> for Coin

maneva3:


platform/contracts/treasury/src/state/reward_scale/mod.rs
```diff
@@ -188,7 +188,7 @@ mod tests {
         .unwrap();
 
         assert_eq!(
-            res.get_apr::<SuperGroupTestC1>(0.into()),
+            res.get_apr::<SuperGroupTestC1>(Coin::new(0)),
```
gmanev7, line 191: use the local `fn coin` to avoid the furbofish operator, or the `ZERO` const

gmanev7 (commented):
went through 20 out of 61 files
Let's do another round when the issues get addressed.

gmanev7 (approved):
Good job!
Thank you for finishing such a high-volume code update!

PR #565, opened 2025-08-26
Remove HigherRank::Intermediate

bmanev18:
With the decreased dependency on cosm_wasm's Uint types, abstractions like HigherRank::Intermediate are no longer required.


gmanev7 (approved):
Good job!

PR #567, opened 2025-08-28
Define the new Ratio

maneva3:


platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
-use crate::{fraction::Fraction, fractionable::Fractionable, zero::Zero};
+use crate::{
+    error::{Error, Result as FinanceResult},
+    fraction::Fraction,
+    fractionable::Fractionable,
+    zero::Zero,
+};
+
+// /// A part of something that is divisible
+#[derive(Clone, Deserialize, Serialize)]
```
gmanev7, line 13: What about `Copy`? I believe we would always benefit from having it.

platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
+#[cfg_attr(any(test, feature = "testing"), derive(Debug, PartialEq))]
+#[serde(
+    try_from = "Rational<U>",
+    into = "Rational<U>",
+    rename_all = "snake_case"
```
gmanev7, line 18: irrelevant

platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
+)]
+pub struct Ratio<U>(Rational<U>)
+where
+    U: Clone + Debug + PartialOrd + Zero;
```
gmanev7, line 22: Why do we impose these restrictions? Fewer bounds lead to broader applicability.

platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
+
+impl<U> Ratio<U>
+where
+    U: Clone + Debug + PartialOrd + Zero,
```
gmanev7, line 26: here too

platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
+{
+    pub(crate) fn new(parts: U, total: U) -> FinanceResult<Self> {
```
gmanev7, line 28: This is done in a "defensive programming" style - i.e., always treat the input as coming from an untrusted source and validate. I would recommend keeping the `new` function with a precondition that is only validated with `debug_assert`, and having the `try_from` to enforce the invariant.

platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
+        Error::broken_invariant_if::<Self>(parts > total, "Parts must not exceed total")
+            .map(|()| Self(Rational::new(parts, total)))
+    }
+}
+
+impl<U> TryFrom<Rational<U>> for Ratio<U>
+where
+    U: Clone + Debug + PartialOrd + Zero,
```
gmanev7, line 36: here too

platform/packages/finance/src/ratio.rs
```diff
@@ -2,9 +2,55 @@ use std::fmt::Debug;
...
+    type Error = Error;
+
+    fn try_from(rational: Rational<U>) -> Result<Self, Self::Error> {
+        Self::new(rational.nominator, rational.denominator)
+    }
+}
+
+impl<U> From<Ratio<U>> for Rational<U>
+where
+    U: Clone + Debug + PartialOrd + Zero,
```
gmanev7, line 47: here too

platform/packages/finance/src/ratio.rs
```diff
@@ -58,3 +104,97 @@ where
...
+        );
+
+        let serialized_2 =
+            cosmwasm_std::to_json_vec(&Rational::new(coin(Amount::MAX), coin(Amount::MAX)))
+                .unwrap();
+        assert_eq!(
+            Ratio::new(coin(Amount::MAX), coin(Amount::MAX)).unwrap(),
+            cosmwasm_std::from_json(&serialized_2).unwrap()
+        );
+    }
```
gmanev7, line 171: This is a test of the `Rational -> bytes -> Ratio`. Do we claim that we support it? Do we need it?
Isn't it rather, `Ratio -> bytes -> Ratio` that we need?

platform/packages/finance/src/ratio.rs
```diff
@@ -58,3 +104,97 @@ where
...
+
+    const fn coin(amount: Amount) -> Coin<SuperGroupTestC1> {
+        Coin::new(amount)
+    }
+
+    fn assert_err<U, E>(res: Result<Ratio<U>, E>, msg: &str)
+    where
+        U: Clone + Debug + PartialOrd + Zero,
```
gmanev7, line 189: here too

platform/packages/finance/src/ratio.rs
```diff
@@ -2,16 +2,73 @@ use std::fmt::Debug;
...
+impl<U> TryFrom<Rational<U>> for Ratio<U>
+where
+    U: Debug + PartialOrd + Zero,
+{
+    type Error = Error;
+
+    fn try_from(rational: Rational<U>) -> Result<Self, Self::Error> {
+        let res = Self::new(rational.nominator, rational.denominator);
+        res.invariant_held()?;
+        Ok(res)
```
gmanev7, line 52:
```suggestion
        res.invariant_held().map(|()| res)
```

platform/packages/finance/src/ratio.rs
```diff
@@ -58,3 +115,72 @@ where
...
+        let ratio_2 = Ratio::new(coin(Amount::MAX), coin(Amount::MAX));
+        let serialized = cosmwasm_std::to_json_vec(&ratio_2).unwrap();
+        assert_eq!(ratio_2, cosmwasm_std::from_json(&serialized).unwrap());
+    }
+
+    #[test]
+    #[should_panic = "Parts must not exceed total"]
+    fn serialize_deserialize_invalid() {
+        let serialized = cosmwasm_std::to_json_vec(&Rational::new(coin(5), coin(4))).unwrap();
+        let _ = cosmwasm_std::from_json::<Ratio<Coin<SuperGroupTestC1>>>(&serialized);
```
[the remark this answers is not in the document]
gmanev7: look for aspirations in the code searching for "debug_asserts"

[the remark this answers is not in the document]
gmanev7: You have applied this recommendation to `Ratio::new`, whereas we primarily need it for deserialization. You may want to leave the new test in `cfg(debug_assertions)`.

PR #570, opened 2025-09-05
 Define SimpleFraction::checked_mul() using MaxPrimitive and TryFromMaxPrimitive

maneva3:


platform/packages/finance/src/fractionable/mod.rs
```diff
@@ -3,14 +3,39 @@ use std::{
...
+    Other: ToDoublePrimitive,
+{
+    type Max: CheckedMul<Output = Self::Max> + Div<Output = Self::Max>;
+
+    // Having two identical methods so the trait becomes symmetric
+    fn into_max(self) -> Self::Max;
+    fn into_max_from(other: Other) -> Self::Max;
+}
+
+pub trait TryFromMaxPrimitive<Max>
```
gmanev7, line 32: Two concerns here:
- not consistent with the reversed trait `ToDoublePrimitive` ... so it is better to be `TryFromDoublePrimitive`
- not related to the other two traits, increasing the risk of misusing it. What about `where Self: ToDoublePrimitive<Double = Max>`

platform/packages/finance/src/ratio.rs
```diff
@@ -102,6 +104,24 @@ where
             denominator,
         }
     }
+
+    pub fn checked_mul<M>(self, rhs: M) -> Option<M>
+    where
+        U: MaxPrimitive<M>,
+        M: ToDoublePrimitive + TryFromMaxPrimitive<<U as MaxPrimitive<M>>::Max>,
```
gmanev7, line 111: The `ToDoublePrimitive` bound is not needed

platform/packages/finance/src/fractionable/mod.rs
```diff
@@ -3,14 +3,39 @@ use std::{
...
 
+pub trait ToDoublePrimitive {
+    type Double;
+}
+
+/// Defines a common `Max` type based on `Self::Double` and `Other::Double` types.
+pub trait MaxPrimitive<Other>
+where
+    Self: ToDoublePrimitive,
+    Other: ToDoublePrimitive,
```
gmanev7, line 23: The `ToDoublePrimitive::Double` type is not used ... which is a sign that the design has a flow.

platform/packages/finance/src/ratio.rs
```diff
@@ -102,6 +102,25 @@ where
             denominator,
         }
     }
+
+    pub fn checked_mul<M>(&self, rhs: M) -> Option<M>
+    where
+        U: ToDoublePrimitive,
```
gmanev7, line 108: it seems unnecessary

platform/packages/finance/src/fractionable/mod.rs
```diff
@@ -3,15 +3,47 @@ use std::{
...
+pub trait ToDoublePrimitive {
+    type Double;
+
+    fn to_double(self) -> Self::Double;
+}
+
+/// Defines a common `Max` type, chosen as one of the `Double` types from either `Self` or `Other`
+pub trait MaxDoublePrimitive<Other>
+where
+    Self: ToDoublePrimitive + Sized,
```
gmanev7, line 24: I am not sure if we need `Sized` here; instead, I suppose it is needed on `type Max`

gmanev7 (approved):

PR #572, opened 2025-09-09
Normalizing Rational

bmanev18:
On Rational::new, the values for nominator and denominator will be scaled down by their greatest common divisor (if it exists)


gmanev7:
@bmanev18,  Please rebase and rethink on:
- the best module for the `Scalar` definition, and
- how `Unit` and `Scalar` relate to each other, I'd expect a valuable input from @maneva3 here

PR #574, opened 2025-09-16
SimpleFraction normalization on ::new

bmanev18:
Instead of rebasing and resolving conflicts in #572, this branch has been started from the latest main version. Because of the low quantity of changes, this approach has been more time-efficient.

#572 will be deleted since it contains nearly the same logic

### DIFF
--- a/platform/packages/finance/src/coin/mod.rs
+++ b/platform/packages/finance/src/coin/mod.rs
@@ -170,7 +170,7 @@ impl<C> FractionUnit for Coin<C> {
     fn modulo(self, scale: Self::Times) -> Self::Times {
         debug_assert_ne!(scale, Self::Times::ZERO);
 
-        self.amount.modulo(scale)
+        self.amount.rem(scale)
     }
 
     fn to_primitive(self) -> Self::Times {
--- a/platform/packages/finance/src/duration.rs
+++ b/platform/packages/finance/src/duration.rs
@@ -158,7 +158,7 @@ impl FractionUnit for Duration {
     fn modulo(self, scale: Self::Times) -> Self::Times {
         debug_assert_ne!(scale, Self::Times::ZERO);
 
-        self.nanos().modulo(scale)
+        self.nanos().rem(scale)
     }
 
     fn to_primitive(self) -> Self::Times {


platform/packages/finance/src/fraction.rs
```diff
@@ -13,6 +13,19 @@ pub trait Fraction<U> {
...
-    Self: Copy + Debug + PartialOrd + Zero,
+    Self: Copy + Debug + PartialOrd + Sized + Zero,
 {
+    type Times: Copy + Debug + PartialEq + Zero;
+
+    fn gcd(self, other: Self) -> Self::Times;
+
+    /// Divides `self` by the given `scale`
+    ///
+    /// [scale] should should be non zero
```
gmanev7, line 24: [a couple of occurrences] nit: double "should"

platform/packages/finance/src/ratio.rs
```diff
@@ -130,6 +133,27 @@ where
     }
 }
 
+fn into_coprime<U>(a: U, b: U) -> (U, U)
```
gmanev7, line 136: Let's not duplicate code, if it's impossible to avoid it at all,  without a plan on how to get rid of it. The trivial step would be to add a TODO, though that would be our last resort. Rather, I'd recommend adjusting it for the Coin<C> use-case.

platform/packages/finance/src/ratio.rs
```diff
@@ -150,6 +153,28 @@ where
...
+        "LHS-value is not divisible by the GCD!"
+    );
+    debug_assert!(
+        b.modulo(gcd) == Zero::ZERO,
+        "RHS-value is not divisible by the GCD!"
+    );
+
+    (a.scale_down(gcd), b.scale_down(gcd))
+} */
+
```
gmanev7, line 177: Do we need this chunk of commented code?

platform/packages/finance/src/fraction.rs
```diff
@@ -10,9 +10,61 @@ pub trait Fraction<U> {
...
+    ///
+    /// [scale] should be non zero
+    fn scale_down(self, scale: Self::Times) -> Self;
+
+    /// Returns the remainder of `self` divided by `scale`
+    ///
+    /// [scale] should be non zero
+    fn modulo(self, scale: Self::Times) -> Self::Times;
+
+    fn primitive(self) -> Self::Times;
```
gmanev7, line 33: We have to follow the [Rust naming guideslines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) and name it `to_primitive`

platform/packages/finance/src/coin/mod.rs
```diff
@@ -152,7 +173,30 @@ impl<C> Default for Coin<C> {
...
+    where
+        U: FractionUnit<Times = Self::Times>,
+    {
+        Gcd::gcd(self.amount, other.primitive())
+    }
+
+    fn scale_down(self, scale: Self::Times) -> Self {
+        debug_assert_ne!(scale, Self::Times::ZERO);
+
+        Coin::new(self.amount.scale_down(scale))
```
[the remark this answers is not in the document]
gmanev7: not revolved for Coin and Duration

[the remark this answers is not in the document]
gmanev7: As I see that you have partially addressed this, I'd like to clarify for you. The comment is attached to this line, but it applies to the entire `impl`.

Issue #297, opened 2024-03-05
[Oracle][Price feeds] Tolerate unknown currencies


gmanev7:
Currently, on receiving a price feed, all currencies must be known. Otherwise, the transaction fails.

To manage easier feeders and the price tree, we have to tolerate unknown currencies. Prices with such a currency involved should be ignored and should emit a warning.

PR #582, opened 2025-10-15
Tolerate unknown currency feeds

Lockwarr:
Closes https://github.com/nolus-protocol/nolus-money-market/issues/297


Lockwarr:
The changes required in https://github.com/nolus-protocol/nolus-money-market/issues/297 are no longer needed by the protocol

PR #583, opened 2025-10-17
Refactor feeders count

bmanev18:
Ensure the count of feeders doesn't go over the limit of the target

### DIFF
--- a/protocol/packages/marketprice/src/feeders/count.rs
+++ b/protocol/packages/marketprice/src/feeders/count.rs
@@ -34,8 +34,12 @@ impl Count {
     }
 
     /// Checks if [self] can be safely incremented
-    pub fn can_increment(&self) -> Option<()> {
-        (self != &Self::MAX).then_some(())
+    pub fn check_increment(&self) -> Result<(), PriceFeedersError> {
+        if self != &Self::MAX {
+            Ok(())
+        } else {
+            Err(PriceFeedersError::MaxFeederCount {})
+        }
     }
 
     /// Converts [self] into a reciproral fraction
@@ -143,14 +147,14 @@ mod test {
 
     #[test]
     fn can_increment_some() {
-        assert_eq!(Some(()), Count::ZERO.can_increment());
-        assert_eq!(Some(()), Count::new_test(100).can_increment());
-        assert_eq!(Some(()), Count::new_test(65536).can_increment());
-        assert_eq!(Some(()), Count::new_test(Unit::MAX - 1).can_increment());
+        assert_eq!(Ok(()), Count::ZERO.check_increment());
+        assert_eq!(Ok(()), Count::new_test(100).check_increment());
+        assert_eq!(Ok(()), Count::new_test(65536).check_increment());
+        assert_eq!(Ok(()), Count::new_test(Unit::MAX - 1).check_increment());
     }
 
     #[test]
     fn can_increment_none() {
-        assert!(Count::MAX.can_increment().is_none());
+        assert!(Count::MAX.check_increment().is_err());
     }
 }
--- a/protocol/packages/marketprice/src/feeders/mod.rs
+++ b/protocol/packages/marketprice/src/feeders/mod.rs
@@ -60,9 +60,7 @@ impl PriceFeeders {
     ) -> Result<(), PriceFeedersError> {
         let mut db = self.feeders(storage)?;
 
-        count_of(&db)
-            .can_increment()
-            .ok_or(PriceFeedersError::MaxFeederCount {})?;
+        count_of(&db).check_increment()?;
 
         (!db.contains(&feeder))
             .then_some(())


protocol/packages/marketprice/src/feeders.rs
```diff
@@ -72,6 +72,42 @@ impl PriceFeeders {
...
+impl FeederCount {
+    pub const MAX: Self = Self(u32::MAX);
+
+    pub(super) fn new(count: u32) -> Self {
+        Self(count)
+    }
+
+    pub(super) fn count(&self) -> u32 {
+        self.0
+    }
```
gmanev7: 1. Please try `const fn`
2. It'd be easier to manage type visibility than fn-by-fn visibility, so I'd recommend having these fn-s `pub` and restricting the type's visibility.

protocol/packages/marketprice/src/config.rs
```diff
@@ -55,8 +58,8 @@ impl Config {
         }
     }
 
-    pub fn min_feeders(&self, total: usize) -> usize {
-        self.min_feeders.of(total)
+    pub fn min_feeders(&self, total: FeederCount) -> FeederCount {
+        FeederCount::new(self.min_feeders.of(total.count()))
```
gmanev7: lacks encapsulation
Your solution defines `new` and `count` and implements externally the calculation of `percent_of` and `less_than`. The latter two could fit perfectly within the type, defeating the need for the former two.

protocol/packages/marketprice/src/feeders.rs
```diff
@@ -72,6 +72,42 @@ impl PriceFeeders {
...
+
+impl From<usize> for FeederCount {
+    fn from(value: usize) -> Self {
+        Self::new(
+            value
+                .try_into()
+                .expect("The total amount of feeders is upper bounded by u32::MAX"),
+        )
+    }
+}
```
gmanev7: While correct, we should not build the error-free conversion into the type `FeederCount`. It is specific for the use case `register-up-to-FeederCount::MAX feeders`, so for better cohesion, we should allow here overflows, `impl TryFrom`. The `expect` should be moved up the stack at the place we count feeders.

protocol/contracts/oracle/src/contract/oracle/feeder.rs
```diff
@@ -38,6 +38,10 @@ impl Feeders {
     where
         PriceG: Group,
     {
+        if FeederCount::from(Self::total_registered(deps.storage)?).ge(&FeederCount::MAX) {
+            return Err(Error::MaxFeederCount {});
+        }
+
```
gmanev7: This check causes an extra load of all feeder addresses from the storage. It'd be better to guarantee that at the persistence level, i.e. in `PriceFeeders::register`.

protocol/packages/marketprice/src/feeders/count.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    fn try_from(value: usize) -> Result<Self, Self::Error> {
+        value
+            .try_into()
+            .map_err(|_| Self::Error::MaxFeederCount {})
+            .map(Self::new)
+    }
+}
+impl CommonDoublePrimitive<Percent100> for Count {
+    type CommonDouble = u64;
```
gmanev7: let's not repeat the knowledge of which the Count's double type is. Instead, refer to it, e.g. `<Count as ToDoublePrimitive>::Double`

protocol/packages/marketprice/src/feeders/count.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl Fractionable<Percent100> for Count {}
+
+impl IntoMax<u64> for Count {
+    fn into_max(self) -> u64 {
```
gmanev7: same as above

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -21,6 +24,9 @@ pub enum PriceFeedersError {
 
     #[error("Unauthorized")]
     Unauthorized {},
+
+    #[error("Maximum feeder count reached")]
+    MaxFeederCount {},
```
gmanev7: Where possible, errors should be accompanied by their cause. Please add the root cause of the error.

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -45,7 +51,11 @@ impl PriceFeeders {
     }
 
     pub fn register(&self, deps: DepsMut<'_>, feeder: Addr) -> Result<(), PriceFeedersError> {
-        let mut db = self.0.may_load(deps.storage)?.unwrap_or_default();
+        let mut db = self.feeders(deps.storage)?;
+
+        if FeederCount::try_from(db.len()).is_err() {
+            return Err(PriceFeedersError::MaxFeederCount {});
+        }
```
gmanev7: More Rust idiomatic would be to use `?` and `Result::map_err` than an `if-return` statement

protocol/packages/marketprice/src/feeders/mod.rs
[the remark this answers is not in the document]
bmanev18: bool methods, which return Result<> are part of a `nightly-only experimental API. `(bool_to_result [#142748](https://github.com/rust-lang/rust/issues/142748))

protocol/packages/marketprice/src/feeders/mod.rs
gmanev7: "Please add the root cause of the error." means please add as a member the error instance that caused this error, not describe the root cause in the message

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -78,6 +78,14 @@ impl PriceFeeders {
             })
             .map_err(Into::into)
     }
+
+    pub fn total_registered(
+        &self,
+        storage: &dyn Storage,
+    ) -> StdResult<Result<FeederCount, PriceFeedersError>> {
```
gmanev7: A few reasons the return type is inappropriate:
1. `Result` of `Result` does not make any sense.
2. [It is not respected for other functions in this type, right, that is an opportunity for improvement.] The result from functions is part of the API. As such, it should not expose types from the underlying APIs used by the implementation. The clients would be tied to that implementation and would be broken whenever that API changes or we switch to using another implementation.

A Result with this API errors would fit perfectly.

protocol/packages/marketprice/src/feed/mod.rs
```diff
@@ -113,13 +113,12 @@ where
...
         config: &Config,
-        total_feeders: usize,
+        total_feeders: FeederCount,
     ) -> bool
     where
         Observations: for<'item> Iterator<Item = &'items Observation<C, QuoteC>>,
     {
-        self.count_unique_feeders(items)
-            >= (config.min_feeders(total_feeders.try_into().expect("Feeder count exceeded")))
+        self.count_unique_feeders(items) >= (config.min_feeders(total_feeders))
```
gmanev7: [nit] an unnecessary pair of brackets

protocol/packages/marketprice/src/feed/sample.rs
```diff
@@ -105,27 +105,18 @@ where
...
-                .fold(*first, |acc, current| acc + *current);
+            let sum = values.fold(*first, |acc, current| acc + *current);
 
-            let part = Ratio::new(
-                1,
-                u128::try_from(prices_len)
-                    .expect("prices_number is already restricted to fit in u128::MAX"),
-            );
-            let avg = FractionLegacy::of(&part, sum);
+            let avg = sum.lossy_mul(&prices_count);
```
gmanev7: From a technical standpoint, this is clever but hard to maintain. The average is the division of the sum by the number, not a multiplication. So the flaws are that FeederCount is represented as RatioLegacy ot its reciprocal value, and that the RatioLegacy contract is breakable if = 0. Yeah, I know, we are in a code branch !=0, but that code is in FeederCount where 0 is allowed..

I'd suggest defining `FeedersCount::try_into_reciproral(self) -> Result<impl RatioLegacy>`, ideally, with a type parameter Self.

protocol/Cargo.toml
```diff
@@ -70,6 +66,7 @@ topology = { path = "../tools/topology" }
 # General
 anyhow = "1"
 base64 = { version = "0.22", default-features = false }
+gcd = { version = "2", default-features = false }
```
gmanev7: 'g' is after 'e', let's keep in order

protocol/packages/marketprice/src/feed/sample.rs
```diff
@@ -116,7 +119,11 @@ where
 
             let sum = values.fold(*first, |acc, current| acc + *current);
 
-            let avg = sum.lossy_mul(&prices_count);
+            let reciproral = prices_count.try_into_reciproral().expect("msg");
```
gmanev7: What should the message be?

protocol/packages/marketprice/src/feed/mod.rs
```diff
@@ -112,19 +113,24 @@ where
...
     where
         Observations: for<'item> Iterator<Item = &'items Observation<C, QuoteC>>,
     {
-        items.map(Observation::feeder).collect::<HashSet<_>>().len()
+        items
+            .map(Observation::feeder)
+            .collect::<HashSet<_>>()
+            .len()
+            .try_into()
+            .expect("Feeder count exceeded")
```
gmanev7: Please update the message to comply with the [official recommendations](https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style).

protocol/packages/marketprice/src/feed/sample.rs
```diff
@@ -102,34 +107,29 @@ where
     }
 
     fn end_of_period(&mut self) {
-        let prices_number = Self::prices_number(self.sample_prices.len());
+        let prices_len = self.sample_prices.len();
+        let prices_count =
+            FeederCount::try_from(prices_len).expect("More prices stored than allowed");
```
gmanev7: The same comment about the message as above

protocol/packages/marketprice/src/feeders/count.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub(crate) const fn new_test(count: u32) -> Self {
+        Self::new(count)
+    }
+
+    pub fn can_increment(&self) -> Option<()> {
+        (self != &Self::MAX).then_some(())
+    }
+
+    pub fn try_into_reciproral(self) -> Option<impl RatioLegacy<Self>> {
+        (self.0 != 0).then_some(SimpleFraction::new(Self::ONE, self))
```
gmanev7: The implementation is top!
However, please refrain from spilling constants through the code. Define 0 as a private const Zero ... that is then used in `impl Zero for Count`

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -7,6 +7,9 @@ use sdk::{
     cw_storage_plus::Item,
 };
 
+mod count;
+pub use count::Count as FeederCount;
```
gmanev7: Publishing with a name carrying the object should be done once the name is about to escape from the object area. Here, the current module is `feeders`, so `Count` does not require specialization. Please move it up the modules tree, where `feeders` is no longer in the context.

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -70,6 +78,22 @@ impl PriceFeeders {
...
     }
+
+    pub fn total_registered(
+        &self,
+        storage: &dyn Storage,
+    ) -> Result<FeederCount, PriceFeedersError> {
+        let result = self
+            .feeders(storage)
+            .map_err(PriceFeedersError::Std)
+            .and_then(|feeders| FeederCount::try_from(feeders.len()));
```
gmanev7: Why do we need to merge the error with others to distinguish it later? Just act on time.

protocol/packages/marketprice/src/tests.rs
```diff
@@ -37,32 +37,32 @@ const DISCOUNTING_FACTOR: Percent100 = Percent100::from_permille(750);
 
 #[test]
 fn register_feeder() {
-    let mut deps = cosmwasm_test::mock_dependencies();
+    let mut storage = cosmwasm_test::mock_dependencies().storage;
```
gmanev7: [here and below] You may want to use `MockStorage::default()`

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -109,7 +109,7 @@ mod tests {
 
     #[test]
     fn remove_empty() {
-        let mut storage = testing::mock_dependencies().storage;
+        let mut storage = testing::MockStorage::default();
```
gmanev7: Add `use testing` to avoid `testing::`

platform/packages/finance/src/price/mod.rs
```diff
@@ -101,7 +101,7 @@ where
     /// Price(amount, amount_quote) * Ratio(nominator / denominator) = Price(amount * denominator, amount_quote * nominator)
     /// where the pairs (amount, nominator) and (amount_quote, denominator) are transformed into co-prime numbers.
     /// Please note that Price(amount, amount_quote) is like Ratio(amount_quote / amount).
-    pub(crate) fn lossy_mul<R>(self, rhs: &R) -> Self
+    pub fn lossy_mul<R>(self, rhs: &R) -> Self
```
gmanev7: this too

protocol/packages/marketprice/src/feeders/count.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Checks if [self] can be safely incremented
+    pub fn can_increment(&self) -> Option<()> {
+        (self != &Self::MAX).then_some(())
+    }
+
+    /// Converts [self] into a reciproral fraction
+    ///
+    /// Returns [None] if the Count is zero
+    pub fn try_into_reciproral(self) -> Option<impl RatioLegacy<CountUnit>> {
```
gmanev7: The visibility of `CountUnit` is probably less than the `struct Count` and this function. Please adjust these three.

protocol/packages/marketprice/src/feeders/count.rs
```diff
@@ -111,3 +111,32 @@ impl FractionUnit for Count {
...
 }
+
+#[cfg(test)]
+mod test {
+
+    use super::Count;
+
+    #[test]
+    fn try_into_reciproral_nonzero() {
+        assert!(Count::new_test(4096).try_into_reciproral().is_some())
```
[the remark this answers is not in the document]
gmanev7: Alright, but the test should somehow verify the value besides being only `Some(...)`. I'd suggest examining the `RatioLegacy`, for example, `assert_eq!(count, ....try_into_reciproral().unwrap().total())`

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    pub fn register(
+        &self,
+        storage: &mut dyn Storage,
+        feeder: Addr,
+    ) -> Result<(), PriceFeedersError> {
+        let mut db = self.feeders(storage)?;
+
+        Count::try_from(db.len())?.can_increment();
```
gmanev7: not fixed yet ...apparently no test either

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    pub fn total_registered(&self, storage: &dyn Storage) -> Result<Count, PriceFeedersError> {
+        let result = self
+            .feeders(storage)
+            .map_err(PriceFeedersError::Std)
+            .and_then(|feeders| Count::try_from(feeders.len()));
+
+        if let Err(PriceFeedersError::MaxFeederCount(_)) = result {
+            panic!("More registred feeders than allowed!")
+        };
```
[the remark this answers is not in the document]
gmanev7: what about this?

protocol/packages/marketprice/src/feeders/mod.rs
```diff
@@ -93,7 +98,7 @@ impl PriceFeeders {
             .map_err(PriceFeedersError::Std)
             .and_then(|feeders| Count::try_from(feeders.len()));
 
-        if let Err(PriceFeedersError::MaxFeederCount(_)) = result {
+        if let Err(PriceFeedersError::FeederCountExceeded(_)) = result {
             panic!("More registred feeders than allowed!")
         };
```
gmanev7: I got the lint warning, but my point is not whether to panic or expect, but to do it *right* after we get the result of conversion.

PR #589, opened 2025-12-02
Housekeeping

maneva3:


platform/packages/finance/src/percent/bound.rs
```diff
@@ -133,6 +146,24 @@ impl<const UPPER_BOUND: Units> RatioLegacy<Units> for BoundPercent<UPPER_BOUND>
...
+    fn into_target(self) -> T;
+}
+
+impl<const UPPER_BOUND: u32> BoundedAs<U256> for BoundPercent<UPPER_BOUND> {
+    fn into_target(self) -> U256 {
+        self.units().into()
+    }
+}
+
+impl<const UPPER_BOUND: u32> BoundedAs<BoundPercent<UPPER_BOUND>> for BoundPercent<UPPER_BOUND> {
```
gmanev7: `BoundedAs<Self>`

platform/packages/finance/src/percent/bound.rs
```diff
@@ -133,6 +146,24 @@ impl<const UPPER_BOUND: Units> RatioLegacy<Units> for BoundPercent<UPPER_BOUND>
     }
 }
 
+/// Defines how a BoundPercent can be represented in another type.
+/// This replaces the generic `From` to allow only explicitly permitted conversions.
+pub(crate) trait BoundedAs<T> {
+    fn into_target(self) -> T;
+}
+
+impl<const UPPER_BOUND: u32> BoundedAs<U256> for BoundPercent<UPPER_BOUND> {
```
gmanev7: may we use `<Coin<C> as DoublePrimitive<Self>>::Type` instead of `U256`?

platform/packages/finance/src/ratio/multiplication.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Checked multiplication of two `SimpleFraction`-s
+/// Returns `None` if either the numerator or denominator multiplication overflows
+impl<U> CheckedMul for SimpleFraction<U>
+where
+    U: CheckedMul<U, Output = U> + FractionUnit,
+{
+    type Output = Self;
+
+    fn checked_mul(self, rhs: Self) -> Option<Self::Output> {
```
gmanev7: if we implement `Fractionable<U> for SimpleFraction<U>` we may not need this implementation that is a specific case of the `SimpleFraction<U>::checked_mul`

platform/packages/finance/src/price/base/fraction.rs
```diff
@@ -0,0 +1,30 @@
+use bnum::types::U256;
+
+use crate::{fraction::Unit as FractionUnit, zero::Zero};
+
+impl FractionUnit for U256 {
```
gmanev7: What is this needed for?

platform/packages/finance/src/ratio/multiplication.rs
```diff
@@ -0,0 +1,31 @@
...
+    fraction::Unit as FractionUnit,
+    fractionable::{Fractionable, IntoMax, TryFromMax, checked_mul::CheckedMul},
+    ratio::SimpleFraction,
+};
+
+impl<U> SimpleFraction<U>
+where
+    U: FractionUnit,
+{
+    pub fn checked_mul<M>(&self, rhs: M) -> Option<M>
```
gmanev7: a design question to be discussed further: would it benefit from putting it into the form of `impl CheckedMul<>` ?

gmanev7 (approved):

PR #595, opened 2026-02-02
Use Permilles and Percent distinguish hundred and max

maneva3:


platform/packages/finance/src/percent/bound.rs
```diff
@@ -5,52 +5,48 @@ use std::ops::{Add, Sub};
...
-        if permille <= UPPER_BOUND {
+    const fn try_from_permille(permille: Permilles) -> Option<Self> {
+        if permille.units() <= UPPER_BOUND {
             Some(Self(permille))
         } else {
             None
         }
     }
 
     // TODO revisit it's usage and remove
```
gmanev7: not a goal  for this PR, but is related to, hence please review if we can remove this accessor, and if not - what should we do to remove it

tests/src/lease/close_policy/trigger.rs
```diff
@@ -86,6 +86,6 @@ fn assert_events(resp: &AppResponse, lease: &Addr, exp_strategy_key: &str, exp_l
         &resp.events,
         &Event::new("wasm-ls-auto-close-position")
             .add_attribute("to", lease)
-            .add_attribute(exp_strategy_key, exp_ltv.units().to_string()),
+            .add_attribute(exp_strategy_key, exp_ltv.permilles().to_string()),
```
gmanev7: always, always be hesitant and conscious when making changes in tests - Such changes may "hide" bugs!
Before, the value of this event attribute was "304", for example, i.e. 30.4%. Now it become "30.4%" which breaks the clients who watch out for these events

platform/packages/finance/src/price/arithmetics.rs
```diff
@@ -66,8 +69,11 @@ where
     where
         QuoteQuoteC: 'static,
     {
+        // The second multiplicand could not be a Price.
+        // Constructing Price<QuoteC, QuoteC> would enforce the invariant that same-currency prices are identity (1/1),
+        // which is incorrect for cross-rate arithmetic intermediates and would panic.
         Price::new(self.amount, rhs.amount_quote)
-            .lossy_mul(SimpleFraction::new(self.amount_quote, rhs.amount).to_fraction())
+            .lossy_mul(SimpleFraction::new(self.amount_quote, rhs.amount).map(Coin::to_primitive))
```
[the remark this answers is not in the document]
maneva3: Unfortunately, no—but `.lossy_mul(SimpleFraction::new(self.amount_quote, rhs.amount))` works.

platform/packages/finance/src/ratio/mod.rs
```diff
@@ -107,19 +106,24 @@ where
         }
     }
 
-    pub(super) const fn nominator(&self) -> U {
-        self.nominator
+    pub(super) fn inv(&self) -> Self {
+        Self::new(self.denominator, self.nominator)
     }
```
gmanev7: This is not a correct operation and may break the invariant. A valid SimpleFraction(0, 10) would produce an invalid one.

PR #608, opened 2026-04-14
feat(tools/topology): Add IBC symbol overriding.

KirilMihaylov:


tools/topology/src/skippable.rs
```diff
@@ -0,0 +1,20 @@
+use serde::{Deserialize, Deserializer};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default)]
+pub(crate) enum Skippable<T> {
+    #[default]
+    Skipped,
+    Some(T),
+}
```
gmanev7, line 33: very strange variation of `Option<T>` ... no doc about the motivation behind it ... what problem does it solve/avoid? Why is the second option `Some(T)`, not `T' ?

KirilMihaylov: The motivation is to behave like `Option<T>` but disallow the use of `null` values. The current implementation allows for example the `icon` field to be set to null, which is undesired. This way it's either not set or set to a non-null value. This is closely aligned with normalization forms in RDBMS. The data itself doesn't allow null to be stored, even if it may appear due to `JOIN` statements.

gmanev7: I do not know what `null` is in Rust, what you refer to when saying `null`. If it is None, then that is perfectly alifned with `Option<T>`, If it is `""` (empty String), then your type behave the same as `Option<T>`, so I do not see any merits in adding such an abstraction.

gmanev7 (approved):

PR #631, opened 2026-05-20
feat(lease): RemoteLeaseCallback entry point

gmanev7:
## Summary

Implements §§3.7/3.7.1/3.7.3 of the Remote Lease Protocol ADR — the Lease contract now receives operation outcomes from its configured `remote_lease` controller via a new `ExecuteMsg::RemoteLeaseCallback` variant, authorised against a per-lease controller address. Success / error / timeout each route into the existing safe-delivery pipeline (`ResponseDelivery` + `DexCallback`), so the controller's `ibc_packet_ack` commits as long as the lease can persist its delivery state. Closes [ibc-solray#141](https://github.com/nolus-protocol/ibc-solray/issues/141).

## Changes

- `Config.remote_lease: Addr` field on the Lease and the Leaser; threaded through every site that creates a `Lease`, `BuyAsset`, `OpenIcaAccount`, `BuyLpn`, `SellAsset`, or `TransferIn`. Each in-swap state surfaces it via `SwapTask::remote_lease()` / `IcaConnectee::remote_lease()`.
- `ExecuteMsg::RemoteLeaseCallback(RemoteLeaseCallback)` variant on the Lease contract; routed via `state.on_remote_lease_callback(...)`.
- New auth hook `dex::Handler::authz_remote_lease_callback`. Default impl rejects every caller; `out_local::State` / `out_remote::State` umbrella enums fan it out across variants. The actual check is the shared `check_remote_lease_callback` helper in `packages/dex/src/response.rs`, using `SingleUserPermission` as the `SameContractOnly` analogue.
- Variant→method dispatch in `contracts/lease/src/contract/state/dex.rs` extracted into a pure `classify_callback` helper plus a `CallbackDispatch` enum, then routed to the existing `on_dex_response` / `on_dex_error` / `on_dex_timeout` pipeline.
- Storage version bumped (lease v9 → v10, leaser v6 → v7). The new field is unmigratable on purpose — no v9 instance is planned to be carried across. `migrate(...)` returns the new `ContractError::UnsupportedMigration` so an accidental upgrade fails loudly rather than silently leaving the field uninitialised.

## Verification

- `cargo each --tag ci --tag @agnostic -p lease -p dex -p leaser run -- clippy --all-targets --locked` — clean across each crate's feature combinations.
- `cargo each --tag ci --tag @agnostic -p lease -p dex -p leaser run -- test --all-targets --locked` — all tests pass (lease: 91, dex: 6, leaser: 29).
- Pre-existing `unnecessary_sort_by` lints in `contracts/oracle/src/contract/config.rs` exist on `origin/main` and are out of scope.

## Test coverage

- Auth gate (`packages/dex/src/response.rs`): no controller → `UnsupportedOperation`; matching sender → `Ok(())`; mismatched sender → `Unauthorized`.
- Variant→method dispatch (`contracts/lease/src/contract/state/dex.rs::classify_callback`): each of `OperationOk` / `OperationErr` / `OperationTimeout`.
- Wire shape pinned at the `ExecuteMsg::RemoteLeaseCallback` boundary for all three variants (`contracts/lease/src/api/mod.rs`).

## Out of scope

- Removal of `DexCallbackContinue` — ADR §3.7.1 says it's no longer required, but the issue explicitly states existing `DexCallback` / `SameContractOnly` machinery stays unchanged. Tracked under ibc-solray#143.

## Follow-ups

- ibc-solray#142 — switch lifecycle calls to `remote_lease` stubs.
- ibc-solray#143 — remove the now-unused ICA-open machinery (including `DexCallbackContinue`).
- ibc-solray#144 — integration tests for `remote_lease` stubs.


protocol/contracts/lease/src/contract/endpoins.rs
```diff
@@ -61,17 +61,14 @@ pub fn instantiate(
...
-        .map(|()| response::empty_response())
-        .map_err(ContractError::UpdateSoftware)
-        .inspect_err(platform_error::log(deps.api))
+    // v10 adds the `remote_lease` controller address to the on-chain Lease
+    // state. No existing pre-v10 instance is planned to be carried across the
+    // protocol cut-over — a fresh deployment is required. Reject any migrate
+    // call explicitly so an accidental upgrade attempt fails loudly rather
+    // than silently leaving the field uninitialised.
+    Err(ContractError::UnsupportedMigration).inspect_err(platform_error::log(deps.api))
 }
```
kostovster: **[SECURITY] Storage version bump paired with always-fail migrate** — Still present from review at `2781d32`:

Still present from review at `4b2e196`:

CONTRACT_STORAGE_VERSION is bumped 9→10 while migrate is rewritten to unconditionally return UnsupportedMigration. If the protocol upgrade pipeline invokes migrate on an existing v9 instance (the usual cw path when bumping the stored version), it will hard-fail every time. The 'fresh deployment' intent needs to be enforced operationally — otherwise revert the version bump or implement a real migration that initialises `remote_lease`.

protocol/contracts/lease/src/contract/state/dex.rs
```diff
@@ -128,3 +147,25 @@ where
...
+
+fn classify_callback(callback: RemoteLeaseCallback) -> ContractResult<CallbackDispatch> {
+    match callback {
+        RemoteLeaseCallback::OperationOk(response) => cosmwasm_std::to_json_binary(&response)
+            .map(CallbackDispatch::Response)
+            .map_err(Into::into),
+        RemoteLeaseCallback::OperationErr(message) => Ok(CallbackDispatch::Error(
+            ICAErrorResponse::from(message.as_str().to_owned()),
+        )),
+        RemoteLeaseCallback::OperationTimeout => Ok(CallbackDispatch::Timeout),
```
kostovster: **[SECURITY] Counterparty error string echoed verbatim into ICA error path** — Still present from review at `2781d32`:

Still present from review at `4b2e196`:

Still present from review at `a359fec`:

`ICAErrorResponse::from(message.as_str().to_owned())` re-emits text authored by the remote Solana controller into the inner dex error pipeline. Length is bounded by `RemoteErrorMessage`'s 512-byte cap, but content is fully attacker-controlled. Confirm this string never reaches event attributes, persisted storage, or any surface where untrusted payload becomes durable (project checklist on IBC error strings) — if it can, sanitise to a fixed prefix plus opaque digest instead of pass-through.

tests/src/lease/remote_lease_callback.rs
```diff
@@ -0,0 +1,213 @@
...
+    });
+    assert!(
+        event_observed,
+        "expected the safe-delivery retry to be scheduled (`wasm-next-delivery`), got events {:?}",
+        app_response.events,
+    );
+}
+
+fn drive_to_swap_pending() -> (
+    LeaseTestCase,
```
kostovster: **[STYLE][NAMING] CLAUDE.md Code Style 34/36: closure parameter naming** — Still present from review at `2781d32`:

Same bare-`_` closure parameter as in `expect_swap_retry`: `|_| {}` discards the non-unit `&AppResponse`. Rename to `|_response| {}`, or use `|()|` if the parameter is unit.

PR #640, opened 2026-06-11
feat(remote-lease): pair the transfer channel in the handshake version

kostovster:
## What

Two controller fixes that gate end-to-end deployment of the remote-lease channel, plus a docs sync.

**Handshake version pairing (closes nolus-protocol/ibc-solray#388).** The Solana responder validates and persists the lease↔transfer channel binding from the channel version (ibc-solray#322), so the bare `nls-remote-lease.v1` no longer completes the handshake. The controller now:

- learns the Solana-side ICS-20 transfer channel at instantiation (`InstantiateMsg.transfer_channel`, stored in `Config`, surfaced in `ConfigResponse`),
- rejects non-canonical ids at instantiation (`channel-` prefix, decimal ordinal, no sign or leading zeros, `u16` range — exactly what the responder enforces, failing locally instead of cross-chain),
- proposes `nls-remote-lease.v1+transfer=channel-<N>` in `MsgChannelOpenInit` and requires the same grammar on every handshake callback,
- verifies the counterparty's echoed version on `OpenAck` (`InvalidCounterpartyVersion`).

The suffix exists at the handshake layer only. `remote_lease::VERSION` and the per-packet `ProtocolVersion` pin stay bare — extending the shared constant would break packet deserialization on both sides (the double-duty hazard ADR-0002 §3.3 warns about).

**Wire-shape ack decoding (closes #637).** `ibc_packet_ack` used to decode the acknowledged payload into the typed `OperationResponse`, whose coin parsing validates tickers against the currency registry — a response naming an out-of-registry ticker failed *at the controller*, and a failing ack handler makes the relayer retry forever. The controller now decodes the wire shape only and `RemoteLeaseCallback::OperationOk` carries it verbatim; the lease's typed, registry-validating decode runs at the remote-swap leg, where a miss is absorbed (`undecodable-response`) so the ack commits. Wire-shape hardening (length caps, canonical amounts, `deny_unknown_fields`) still applies at the controller — wire garbage remains the one case it may surface as an ack-handling error.

**Docs.** `protocol/docs/remote-lease-callback-flow.md` now matches the shipped callback routing (direct remote-swap processing, leaser-query auth gate, safe-delivery segment scoped to the still-ICA legs, storage v11); `RUNBOOK.md` gains the local-vs-CI clippy divergence entry and the `PROTOCOL_*` env trio.

## Testing

- New handshake tests: bare version rejected, wrong transfer suffix rejected, counterparty echo mismatch rejected on `OpenAck`; instantiation rejects eight non-canonical channel-id shapes.
- New ack-path tests at both levels: the controller forwards an out-of-registry ticker without erring (`packet_ack_out_of_registry_ticker_dispatches_ok`); the lease absorbs it (`out_of_registry_ticker_absorbed_at_lease`).
- Pinned `.bin` StdAck fixture tests unchanged and green; the dispatched-callback wire-shape pin unchanged.
- Full gate green on the CI-pinned toolchain (1.94): build, fmt, clippy with and without test targets, full test run across the protocol and tests workspaces.
- Each commit builds standalone (verified in a separate worktree for the first).

## Reviews

Two independent reviews ran over `4a863c1f0..HEAD` before push; the scope of the diff was separately checked against the plan (no out-of-plan changes, no manifest/CI edits).

**Correctness checklist (12 generic items + the project rules from the contribution checklist): all PASS.** Three follow-ups surfaced and were resolved:

1. *Stored-`Config` schema change with a pass-through `migrate`* — an already-instantiated controller migrated onto this code would pass `update_software` and then fail every `Config::load`. Resolved in the last commit: `migrate` now probes `Config::load`, refusing the upgrade (old code keeps running) instead of bricking after it. No deployed controller instance exists on any network (verified against the archive node — only the five legacy OSMOSIS protocols are registered), so no migration path is owed.
2. *Proposed handshake version asserted only transitively* — the admin test now decodes the emitted `MsgChannelOpenInit` and pins `channel.version` to the suffixed grammar.
3. *Wire-contract spec edit initially left out of the docs commit* — `docs/remote-lease-wire-contract.md` now records the suffixed handshake grammar and the wire-shape ack decoding alongside the unchanged packet-level pin.

**Security-focused review: CLEAR (nothing above MEDIUM).** The MEDIUM was the same migration finding (resolved above). One LOW: `OpenConfirm` carries no counterparty version to verify — unreachable while `OpenTry` stays rejected; the invariant is now documented at the match. Verified non-findings worth recording: (a) moving registry validation out of the ack handler is a net availability improvement — at the merge-base an out-of-registry ticker permanently stranded the ack (unordered channel, no timeout proof possible once the ack is written); all wire-layer hardening (`deny_unknown_fields`, canonical amounts, length caps, version pin) is unchanged, and the fund-routing floor still comes from the leg's own persisted `min_out`, never from the counterparty. (b) No version-grammar injection surface: the suffix is composed from a compile-time constant plus a digits-only validated channel id, config is immutable post-instantiation, and all comparisons are whole-string equality. Coverage note: semgrep is not installed on this machine, so mechanical scanning of the parsing changes did not run; review of that category was manual plus the wire crate's canonicalization tests.

### DIFF
--- a/protocol/contracts/remote_lease/src/contract.rs
+++ b/protocol/contracts/remote_lease/src/contract.rs
@@ -47,6 +47,7 @@ pub fn instantiate(
 ) -> Result<CwResponse> {
     require_non_empty("connection_id", &new_controller.connection_id)
         .and_then(|()| require_non_empty("dex_label", &new_controller.dex_label))
+        .and_then(|()| require_canonical_transfer_channel(&new_controller.transfer_channel))
         .and_then(|()| {
             deps.api
                 .addr_validate(new_controller.protocol_admin.as_str())
@@ -72,6 +73,7 @@ pub fn instantiate(
             Config::new(
                 new_controller.connection_id,
                 new_controller.dex_label,
+                new_controller.transfer_channel,
                 lease_code,
             )
             .store(deps.storage)
@@ -166,6 +168,25 @@ fn require_non_empty(field: &'static str, value: &str) -> Result<()> {
     } else {
         Ok(())
     }
+}
+
+const TRANSFER_CHANNEL_NAME_PREFIX: &str = "channel-";
+
+/// Accept only the canonical decimal rendering of a `u16` ordinal — the
+/// counterparty's responder rejects leading zeros, signs, and ordinals beyond
+/// its 16-bit entity range, so a non-canonical id would fail the handshake
+/// cross-chain instead of failing here at instantiation.
+fn require_canonical_transfer_channel(channel_id: &str) -> Result<()> {
+    channel_id
+        .strip_prefix(TRANSFER_CHANNEL_NAME_PREFIX)
+        .and_then(|ordinal| {
+            ordinal
+                .parse::<u16>()
+                .ok()
+                .filter(|parsed| parsed.to_string() == ordinal)
+        })
+        .map(|_ordinal| ())
+        .ok_or_else(|| Error::NonCanonicalTransferChannel(channel_id.to_string()))
 }
 
 fn authorize_protocol_admin_only(store: &dyn Storage, call_message: &MessageInfo) -> Result<()> {
--- a/protocol/contracts/remote_lease/src/contract.rs
+++ b/protocol/contracts/remote_lease/src/contract.rs
@@ -1,4 +1,7 @@
-use std::ops::{Deref, DerefMut};
+use std::{
+    mem,
+    ops::{Deref, DerefMut},
+};
 
 use access_control::SingleUserAccess;
 use cosmwasm_std::Storage;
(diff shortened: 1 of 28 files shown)


protocol/contracts/remote_lease/src/contract.rs
```diff
@@ -168,6 +177,25 @@ fn require_non_empty(field: &'static str, value: &str) -> Result<()> {
     }
 }
 
+const TRANSFER_CHANNEL_NAME_PREFIX: &str = "channel-";
```
kostovster: **[STYLE] CLAUDE.md: Module item ordering** — The new module constant is declared after functions. Constants belong in the module constants section before function items.

PR #656, opened 2026-06-14
feat: migrate lease repay & liquidation/close swaps onto the remote-lease transport

kostovster:
## Summary
Migrates the lease's repay-with-non-LPN (BuyLpn) and the four position-close swap
flavours (partial/full liquidation, partial/full customer close) off the ICA
out-and-back swap machine onto the remote-lease controller transport:
(fund →) RemoteSwap (asset→LPN) → proceeds drain (LPN home) → funds arrival → finish.
Downstream repay::repay / Repayable::try_repay are unchanged. Closes #648.
Self-contained: no wire bump, no ibc-solray/webapp coordination.

## Commits (bisectable)
1. feat(dex): add a swap-only remote composite
2. feat(lease): migrate BuyLpn repay swap onto the remote transport
3. feat(lease): migrate liquidation and customer-close swaps
4. feat(dex): add a transfer-out-then-remote-swap composite
5. fix(lease): fund the repay swap with a transfer-out leg

## Repay funding (commits 4–5)
The first cut routed the BuyLpn repay swap through the swap-only composite, which
asked the remote account to swap the customer's payment while it was still held on
the local lease contract — an unfunded swap that would strand the payment (caught
in cortex's automated first-pass). Commits 4–5 add a TransferOut→RemoteSwap composite
(a trim of the opening composite without the ICA-open legs, since an opened lease
already holds its account) and route repay through it, so the payment reaches the
remote account before the swap — mirroring the opening flow and the former ICA repay.
The funding leg reports RepayTrx::TransferOut; a regression test asserts the funding
transfer precedes the swap (proven red on the pre-fix code). Liquidation and
customer-close are unaffected: they swap already-remote collateral and legitimately
keep the swap-only composite.

## Behavior change to review (@gmanev)
Liquidation loses its immediate slippage-anomaly exit valve. An under-floor
liquidation swap now re-emits the in-flight leg at the pinned oracle floor and
heals, instead of moving the lease into an admin-resolvable SlippageAnomaly
state. min_out enforcement and proceeds accounting are unchanged → fund-stuck
(not fund-loss) tail risk. The SlippageAnomaly state, its heal transitions, and
the slippage-anomaly event are removed; the query Status::SlippageProtectionActivated
and the anomaly-resolution permission are kept intentionally (wire unchanged).
The four slippage-anomaly tests are removed with that state; controller under-floor
re-emit is covered by a new dex unit test, and reserve top-up on a below-loan
liquidation by a new integration test.

Restoring this valve on the remote transport — a slippage-anomaly terminal plus a
re-quoting, operator-gated heal — is the next leg, tracked in #655.

## Wire / indexer impact
New event ls-repay-transfer-out and new internal state tags (BuyLpnDrain + the
four close-drain arms) are additive; StateResponse/Status and RepayTrx/PositionCloseTrx
inner tags unchanged (drain states report as in_progress). Indexer verified
forward-compatible; ls-repay-transfer-out is not indexed today (additive follow-up
if analytics want it, no ordering constraint).

Removing the SlippageAnomaly variant and reshaping the BuyLpn/close state arms
changes the persisted State schema. This is safe not because an old code path is
preserved, but because the lease `migrate` entry point unconditionally rejects
(UnsupportedMigration) and no remote-lease instances exist on-chain — the new code
is never deserialized over old bytes.

## Quality
cargo +1.94.0 build/fmt/clippy --no-deps clean on dex + lease; tests crate clippy clean.
Tests: dex 43, lease 116, integration 140 — all green. Each commit independently
builds + passes its contract tests. Reviewed for correctness (12-point checklist) and
security (Trail of Bits CosmWasm/cosmos + differential review + semgrep): no
critical/high; the medium is the documented liquidation-valve change above; two low
findings are inert orphaned wire kept for compatibility, tracked separately.

## Follow-ups (tracked separately)
- #655 — the anomaly re-quote design + operator-recovery policy that restores the
  valve dropped above on the remote transport (the next leg).
- cross-repo cleanup of the orphaned anomaly-resolution permission +
  SlippageProtectionActivated.
- #657 — reconcile the dex State-composite naming with the parasite-words rule. The
  naming nits raised on this PR follow the entrenched dex vocabulary and were deferred
  there rather than renamed piecemeal.


protocol/packages/dex/src/impl_/out_swap.rs
```diff
@@ -0,0 +1,430 @@
...
+
+use super::{RemoteSwap, TransferOut, TransferOutRespDelivery};
+use crate::{ForwardToInner, SwapTask as SwapTaskT};
+
+#[derive(Serialize, Deserialize)]
+#[serde(bound(
+    serialize = "SwapTask: Serialize",
+    deserialize = "SwapTask: Deserialize<'de> + SwapTaskT"
+))]
+pub enum State<SwapTask, SwapClient, ForwardToInnerMsg>
```
kostovster: **[STYLE][NAMING] CLAUDE.md Code Style 10: parasite words** — `State` is a new public workflow abstraction and the banned parasite word itself; rename this enum to a domain name such as `OutSwap` or `OutSwapOp`.

protocol/packages/dex/src/impl_/out_swap.rs
```diff
@@ -0,0 +1,430 @@
...
+        fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+            match self {
+                State::TransferOut(inner) => Display::fmt(inner, f),
+                State::TransferOutRespDelivery(inner) => Display::fmt(inner, f),
+                State::RemoteSwap(inner) => inner.fmt(f),
+            }
+        }
+    }
+}
+
```
kostovster: **[STYLE] CLAUDE.md: Method syntax over fully-qualified trait calls** — Use method syntax here, e.g. `inner.fmt(f)`, since there is no ambiguity. The same match already uses method syntax for the `RemoteSwap` arm.

PR #713, opened 2026-07-03
fix: absorb bad establishment acks in OpenProfit and add heal re-emit

kostovster:
## Change

Bug: `OpenProfit::on_remote_response` returned `SwapDecision::Finished(Err(..))` for the `Ok(_other)` (decodable non-`OpenProfit` variant) and `Err(_decode)` (undecodable) arms. Those faults ride the controller callback path, so the `Err` reverts the controller's `ibc_packet_ack`, causing the relayer to redeliver the same malformed establishment ack forever.

Alternatives considered:
1. Err-and-loop (status quo) — never strands state, but relayer loops indefinitely on a malformed ack. Rejected: not operable.
2. Absorb-only — return Ok on the fault arms so the ack commits. Rejected: it WEDGES the profit permanently. The establishment `OpenProfit` packet is one-shot and now consumed; no further callback arrives, and OpenProfit had no heal path, so the contract is stuck pre-Idle with no recovery.
3. Absorb + heal (chosen) — the only shape satisfying both "never strand the relayer" AND "operator-recoverable". On the two fault arms, ABSORB: stay in OpenProfit, emit a named-const `profit-establishment` event (`absorbed=unexpected-response-variant`/`undecodable-response`), return Ok so the ack commits — mirroring dex `remote_swap.rs`'s `absorb`/`emit_absorbed` shape. AND add a `Handler::heal` override that RE-EMITS the establishment `OpenProfit` packet (via the existing `open_profit_msg`, emitting `heal=re-emit`), mirroring how the dex remote-swap leg re-emits its in-flight op on heal. Result: no relayer loop, and `ExecuteMsg::Heal` (permissionless at the contract entry, like the remote-swap heal — the controller still authorises the callback it answers) re-solicits the acknowledgment. Well-formed `OpenProfit` acks are unchanged: learn authority, transition to Idle. Introduced a private `stay_open` shared by absorb and heal to build the stay-in-OpenProfit Continue response. Removed the now-unused `NON_OPEN_PROFIT_RESPONSE` const. No storage/wire change.

## Testing

- bad_variant_ack_is_absorbed_and_stays_open (failed pre-fix: 'expected a continuation, got a finish' -> passes)
- undecodable_ack_is_absorbed_and_stays_open (failed pre-fix: 'expected a continuation, got a finish' -> passes)
- heal_reemits_the_establishment_packet (failed pre-fix: 'handle heal is not supported in OpenProfit' -> passes)
- open_profit_ack_transitions_to_idle (regression guard: passed before and after; asserts happy-path Idle transition unchanged)

Gate: build, `fmt --check`, clippy on the CI-pinned 1.94 toolchain (`lint` + `lint-all`), full test run for the touched packages — all green.

## Independent review

Two independent fresh-context reviews (general-correctness checklist + a security-focused pass over the committed diff) returned **PASS** with zero blocking findings; security verdict: CLEAR.

## Automated-review response (2026-07-03)

Three findings fixed in follow-up commits (test const ordering, indent flatten, heal-revert-path coverage). The remaining four, verified against code:

- **Missing nonce guard on the heal re-emit** — the mechanism as stated (double-transition in `OpenProfit`) does not exist: the first valid ack transitions to `Idle` and `learn_authority` runs exactly once. The real superseded-packet window the heal creates is handled by composition with the settled-state absorb change (#708) — **merge order: #708 lands before or together with this PR**; on this branch alone `Idle` still rejects the superseded callback: `Idle` authorizes the pinned controller and absorbs the late callback with an event, so the superseded packet's ack/timeout always commits. Nonce 0 cannot collide — `Idle` holds nothing in flight, and the swap/drain legs open at nonce ≥ 1. The `NO_NONCE` doc comment is updated to state the actual invariant. A per-emission establishment nonce remains a possible hardening but adds no safety the absorb does not already provide.
- **Silent swallow of decode error** — deliberate: mirrors the remote-swap absorb (`remote_swap.rs`, `ABSORB_UNDECODABLE`), and surfacing the raw serde text would echo a counterparty-authored payload unbounded, against the IBC echo-bound convention.
- **Permissionless heal** — the codebase-wide heal convention (dex swap/transfer heals identically ignore the sender); authorization lives on the callback that answers the re-emission, gated to `Config.remote_profit_controller`. The heal only re-emits at the caller's own gas cost.
- **Unrecoverable error path on heal re-emit** — the `Err` arm is unreachable in production (the drain-vault address is validated before `OpenProfit` is entered and the config is immutable in that state), and if hit reverts only the retryable heal transaction, no ack. Now covered by a test.

## Automated-review response, round 3 (2026-07-04)

- **"Missing entry-point validation on heal" / "remote-swap heal is operator-gated"** — conflates the two dex heals. The live-leg heals are permissionless (`remote_transfer_out.rs` heal ignores the sender; `ExecuteMsg::Heal` carries no access check by design); only the parked slippage-anomaly terminal's heal is admin-gated, because resolving it forfeits a pinned floor. Re-emitting an establishment packet forfeits nothing and grants nothing — the caller pays the gas, and the callback that answers is gated to the pinned controller. The doc comment now names the live-leg referent explicitly.
- **"Idle authorizes" mechanism imprecise** — `idle.rs` carries Idle's own `authz_remote_callback` override authorizing the pinned controller; the state machine invokes exactly that before delivery, and the absorb is Idle's inherited handler behavior. The comment is accurate at doc-comment altitude.
- Remaining listed items are the prior round's threads, already dispositioned above and resolved. This PR now carries the `no-auto-review` label — the disposition record lives here in the body.

### DIFF
--- a/protocol/contracts/profit/src/state/open_profit.rs
+++ b/protocol/contracts/profit/src/state/open_profit.rs
@@ -187,8 +187,11 @@ impl Handler for OpenProfit {
     /// The one operator recovery out of a wedged establishment: an absorbed bad
     /// ack consumed the one-shot `OpenProfit` packet without transitioning, so
     /// no further callback will arrive. Re-emitting the packet re-solicits the
-    /// acknowledgment. Permissionless, like the remote-swap heal — the
-    /// controller still authorises the callback it answers with.
+    /// acknowledgment. Permissionless, like the live-leg transfer-out heal
+    /// (the admin-gated heal is the parked slippage-anomaly terminal's, which
+    /// forfeits a pinned floor — re-emitting an establishment packet forfeits
+    /// nothing) — the controller still authorises the callback it answers
+    /// with, and the caller pays the gas.
     fn heal(
         self,
         _querier: QuerierWrapper<'_>,
--- a/protocol/contracts/profit/src/state/open_profit.rs
+++ b/protocol/contracts/profit/src/state/open_profit.rs
@@ -8,7 +8,10 @@ use dex::{
     Result as SwapDecision,
 };
 use finance::{duration::Duration, instant::Instant};
-use platform::{batch::Batch, message::Response as PlatformResponse};
+use platform::{
+    batch::{Batch, Emit, Emitter},
+    message::Response as PlatformResponse,
+};
 use remote_profit::{
     msg::{NolusReceiver, OpenProfitParams},
     response::{OpenProfitResponse, OperationResponse},
@@ -26,9 +29,17 @@ use super::{Config, State, StateEnum, idle::Idle};
 /// is governance-driven and deploys a new contract.
 const INSTANCE_ORDINAL: u16 = 0;
 
-/// A non-`OpenProfit` acknowledgment cannot have resolved the establishment
-/// packet — the only operation this state emits.
-const NON_OPEN_PROFIT_RESPONSE: &str = "non-open-profit operation response";
+/// The establishment event surfacing a bad-ack absorb or an operator heal, so
+/// monitoring can tell a wedged-then-recovered establishment apart from a
+/// normal cycle.
+const EVENT_TYPE_ESTABLISHMENT: &str = "profit-establishment";
+const EVENT_KEY_ABSORBED: &str = "absorbed";
+const EVENT_KEY_HEAL: &str = "heal";
+const EVENT_VALUE_REEMIT: &str = "re-emit";
+/// A decodable acknowledgment carrying a non-`OpenProfit` variant.
+const ABSORB_UNEXPECTED_VARIANT: &str = "unexpected-response-variant";
+/// An acknowledgment payload that does not decode into an operation response.
+const ABSORB_UNDECODABLE: &str = "undecodable-response";
 
 /// The establishment state.
 ///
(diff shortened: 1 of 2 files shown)


protocol/contracts/profit/src/state/open_profit.rs
```diff
@@ -119,25 +164,40 @@ impl Handler for OpenProfit {
...
+    /// absorbed rather than erroring: the establishment ack rides the callback
+    /// path, so an error would revert the controller's ack and loop the relayer
+    /// forever on the same malformed packet. A well-formed `OpenProfit` ack runs
+    /// the regular flow and lets any downstream failure propagate.
     fn on_remote_response(
         self,
         data: sdk::cosmwasm_std::Binary,
         _nonce: u64,
         querier: QuerierWrapper<'_>,
         env: Env,
```
kostovster: **[SECURITY] Missing nonce guard on IBC ack handling** — `on_remote_response` ignores `_nonce` and `reemit_establishment` re-emits with the same `INSTANCE_ORDINAL = 0`, so duplicate or stale acks from repeated heals are not rejected by nonce mismatch. The documented lease-side heal idempotency model (`remote-lease-wire-contract.md` §636) relies on nonce bumping; this weaker model means two heals whose acks both arrive while still in `OpenProfit` could double-transition.

PR #727, opened 2026-07-10
docs: track CLAUDE.md in the repository

kostovster:
## What

- Track `CLAUDE.md` in the repository (and drop its `.gitignore` entry). Contributor guidance previously lived only in local working copies; tracking it on `main` gives tooling and contributors one shared entry point. Code-style conventions are imported from a private toolkit rather than restated here.
- The content is verified against the current `main`: descriptions of designs that are no longer in the tree were removed, and the workspace/package/test descriptions were aligned with the code as it stands.

## Notes

- Single doc-only commit, rebased onto the current `main`; no code changes.
- `RUNBOOK.md` is already tracked on `main` and unchanged, so no runbook changes are needed here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,9 +16,6 @@
 # Cargo+Git helper file (https://github.com/rust-lang/cargo/blob/0.44.1/src/cargo/sources/git/utils.rs#L320-L327)
 .cargo-ok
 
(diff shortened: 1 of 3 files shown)


gmanev7 (approved):
LGTM

PR #759, opened 2026-07-30
Remove the multi-protocol `cargo-each`-tag build leftovers.

KirilMihaylov:
## What

Leftovers of the retired multi-protocol build scheme — each protocol once built separately via `cargo-each` tags — removed across CI, manifests, tooling, and docs. Three commits:

1. **Remove the dead per-DEX build pass and multi-protocol-era docs.**
   - `.github/workflows/code_coverage.yaml` deleted: `workflow_dispatch`-only, referenced GitHub actions that no longer exist, and enumerated per-network/per-protocol tags that match no package.
   - `ci/build.sh`: the second protocol-workspace pass (`"build" "dex-${dex_type}"`) selected zero packages — no manifest declares a `dex-*` tag — and only re-ran `wasm-opt`/`cosmwasm-check`/sha256 over the first pass's binaries. Removed together with `get_dex_type`, the only in-container consumer of `jq` (also dropped from the `build-protocol` image).
   - README/CLAUDE.md lose only the per-protocol selection steps (`src_by_protocol` copy, `-t <protocol> -t <net>` arguments, `NET`/`PROTOCOL` variables and the feature lists derived from them); the protocol image's `build-configuration` mount is now documented.
2. **Retire the `@agnostic` cargo-each tag.** It marked protocol-agnostic packages back when per-protocol tags existed beside it; today every combination carried it, so it discriminated nothing. Removed from all 33 manifests and every request site.
3. **Drop cargo-each's `generics` combination machinery.** Its single remaining consumer — the `oracle` manifest — declared a placeholder that appeared nowhere in the combination, expanding into two identical combinations and running oracle's stub feature matrix twice per lint/test pass.

## Compatibility notes

* **Retired tags fail silent, not loud.** `cargo each --tag @agnostic …` (or any retired tag) now selects zero packages and exits `0` with no output, so external scripts still requesting retired tags will silently no-op. The remaining tag vocabulary is `ci` and `build`.
* **`generics` keys are rejected.** A manifest still using `generics = { … }` inside `[package.metadata.cargo-each]` now fails deserialization with an unknown-field error naming the key.
* The descriptor JSON formats (`build-configuration/{protocol,topology}.json`) are unchanged.

## Verification

- `cargo each list` captures across all four workspaces × {`ci`, `build`, untagged}: byte-identical before and after the tag retirement; after the `generics` removal the only delta is oracle's duplicated 16-variant stub block appearing once instead of twice.
- Full battery on the final tree, protocol/tests from clean target directories: `cargo build --all-targets`, `cargo fmt --all -- --check`, `cargo lint`, `cargo lint-all`, `cargo run-test` — green in all four workspaces.
- `sh -n ci/build.sh`; descriptor JSONs re-validated; repo-wide residue greps for `@agnostic`, `jq` (in `ci/`), `dex-*` requests, and `generics` (in manifests) all empty.

## Review

We reviewed the full diff against the project's 12-item bug checklist — all items pass: no debug leftovers or commented-out code; no off-by-one changes; all match arms handled (the deleted enum takes its only match with it); error paths stay loud (`deny_unknown_fields` rejects a leftover `generics` key by name); serde `default`/`borrow`/`rename_all`/`deny_unknown_fields` semantics preserved field-for-field across the `Config`/`Combination` restructure; no async, SQL, or concurrency surface; no resource handling changed; no new public surface; no logic inversions (the added mount condition keeps the polarity of the conditional it replaces); the removal deletes per-combination clones outright.

A security-focused pass confirmed: CI integrity gates intact (`cosmwasm-check` capability allowlist, contract-count and size checks untouched — the deleted pass contributed no artifacts); zero dependency or lockfile changes across all 33 manifest edits (verified line-by-line: tag removals plus one `generics` key only); `cargo audit` clean on all four lockfiles (pre-existing allowlisted `paste` advisory only); net attack-surface reduction (`jq` out of the trusted build image, one fewer `workflow_dispatch` entry point, an untagged serde enum removed).


gmanev7 (commented):
LGTM

gmanev7 (approved):

